PR #51, opened 2022-04-04
Feat: Feedback if kind image needs to be downloaded

### LINKED ISSUE #41: feat: feedback if kind image needs to be downloaded
If the image `kindest/node:v1.23.4` is not already available locally it will take some time depending on internet connection to download it.

At the moment the user only sees "Setting up kind cluster" which is fast if the image is available.
If not, this can take some minutes - we should let the user know.

joschkabraun:
Still WIP to fix #41


florian-hoenicke (dismissed):

now/cloud_manager.py
```diff
@@ -40,6 +40,11 @@ def create_local_cluster(kind_path):
             cowsay.cow('see you soon 👋')
             exit(0)
     with yaspin(text="Setting up local cluster", color="green") as spinner:
+        out, err = cmd('docker images kindest/node')
```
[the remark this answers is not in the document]
joschkabraun: I incorporated that which lead to the overloading `cmd` function to taking in lists as arguments as well. Otherwise the splitting of the command string lead to issues.

now/deployment/deployment.py
```diff
@@ -1,11 +1,12 @@
...
 import tempfile
 
-
 def cmd(command, std_output=False, wait=True):
+    if not isinstance(command, list):
+        command = command.split()
     if not std_output:
         process = subprocess.Popen(
-            command.split(), stdout=subprocess.PIPE, stderr=subprocess.PIPE
+            command, stdout=subprocess.PIPE, stderr=subprocess.PIPE
```
makram93, line 9: I think you should remove this `.split()` in else part too.

now/cloud_manager.py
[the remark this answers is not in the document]
florian-hoenicke: you could also use the python api. I think that would be the cleanest
https://docker-py.readthedocs.io/en/stable/images.html#docker.models.images.ImageCollection.list

now/deployment/deployment.py
```diff
@@ -1,11 +1,12 @@
 import subprocess
 import tempfile
 
-
 def cmd(command, std_output=False, wait=True):
+    if not isinstance(command, list):
```
makram93, line 5:
```suggestion
    if isinstance(command, str):
```

makram93 (commented):
Looks good. Some comments

joschkabraun:
Incorporated changes. Left in option to send list of commands to `cmd` as it seemed useful in the future.

now/cloud_manager.py
```diff
@@ -41,6 +42,9 @@ def create_local_cluster(kind_path):
             cowsay.cow('see you soon 👋')
             exit(0)
     with yaspin(text="Setting up local cluster", color="green") as spinner:
+        kindest_images = docker.from_env().images.list('kindest/nod')
```
makram93, line 45:
```suggestion
        kindest_images = docker.from_env().images.list('kindest/node')
```

joschkabraun: thank you!

makram93 (approved):
LGTM!

florian-hoenicke (changes requested):
docker==5.0.3 has to be added to requirements_pip.txt

florian-hoenicke (approved):

florian-hoenicke:
Thanks. Nice first pr :)

PR #53, opened 2022-04-05
feat: handle bad cluster

florian-hoenicke:


now/cloud_manager.py
```diff
@@ -18,37 +18,43 @@ def create_local_cluster(kind_path):
...
 
 def is_local_cluster(**kwargs):
+    command = f'{kwargs["kubectl_path"]} get nodes -o json'
     out, error = cmd(f'{kwargs["kubectl_path"]} get nodes -o json')
-    out = json.loads(out)
+    try:
+        out = json.loads(out)
+    except:
+        print(f'Command {command} gives the following error: {out} {error}')
+        exit(1)
```
makram93, line 57: Why do we need to put it in try block?

florian-hoenicke: in case, the command does not work, an error is thrown and the output is not a json.

makram93: Then we can just print the error component in string format: 
```suggestion
    try:
        out = json.loads(out)
    except:
        print(f'Command {command} gives the following error: {error.decode("utf-8")}')
        exit(1)
```

florian-hoenicke:
@makram93 Thanks, I accepted the suggestion you gave me

makram93 (approved):

Issue #42, opened 2022-03-25
fix: abort if docker runs out of disk space.


slettner:
<img width="1768" alt="Screenshot 2022-03-25 at 12 02 17" src="https://user-images.githubusercontent.com/51201318/160112407-a72ca905-73bb-4256-a11e-ca1a91345a60.png">

Jina-Now should probably abort or handle this error in some way.
It can be solve by increasing the disk space resource in the docker settings.

florian-hoenicke:
Thanks for reporting. You are right we have to address this error instead of just continuing

PR #58, opened 2022-04-05
fix: terminate when failed to create cluster

makram93:
Terminate the program execution when the kind fails to create cluster due to any issue and print the error as it is: #42

### DIFF
--- a/now/cloud_manager.py
+++ b/now/cloud_manager.py
@@ -46,7 +46,7 @@ def create_local_cluster(kind_path):
         )
         if err and 'failed to create cluster' in err.decode('utf-8'):
             print('\n' + err.decode('utf-8').split('ERROR')[-1])
-            exit(0)
+            exit(1)
         spinner.ok("📦")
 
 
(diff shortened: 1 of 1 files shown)


now/cloud_manager.py
```diff
@@ -41,9 +41,12 @@ def create_local_cluster(kind_path):
             cowsay.cow('see you soon 👋')
             exit(0)
     with yaspin(text="Setting up local cluster", color="green") as spinner:
-        cmd(
+        _, err = cmd(
             f'{kind_path} create cluster --name {cluster_name} --config {cur_dir}/kind.yml',
         )
+        if err and 'failed to create cluster' in err.decode('utf-8'):
+            print('\n' + err.decode('utf-8').split('ERROR')[-1])
+            exit(0)
```
florian-hoenicke, line 49:
```suggestion
            exit(1)
```
 A successful command returns a 0. In this case, it should return non zero

florian-hoenicke (changes requested):

florian-hoenicke (approved):
LGTM

PR #59, opened 2022-04-05
fix: move import inside func to make exec fast

makram93:
I found that after fresh installation in a new environment, on calling `jina --help` and `streamlit --help` takes time for the first execution, however second onwards it is fast. Same behavior is observed with `jina-now`. I have moved the import statement inside func so that there's less delay on calling `jina-now --help` #56


florian-hoenicke (approved):
LGTM

PR #73, opened 2022-04-06
fix: handle ctrl+c and remove os clear

makram93:
I have added sigmap which wil handle the `SIGINT` & `SIGTERM` interruptions.
Also the `os.system['clear']` is removed as it causes deleting the terminal history which is not wanted!


now/dialog.py
```diff
@@ -92,7 +93,7 @@ def prompt_plus(questions, attribute, **kwargs):
         if attribute in answer:
             return answer[attribute]
         else:
-            os.system('clear')
+            print("\n" * 10)
```
makram93, line 96: Replacing the `clear` command with buffer print. Not very elegant but does the work and does not alter the terminal

Issue #79, opened 2022-04-07
fix: gpu bug reported by tobi


florian-hoenicke:
Please investigate this bug:
https://files.slack.com/files-pri/T0189QW56UE-F03AJ33BX19/untitled.txt

PR #92, opened 2022-04-11
fix: use cpu only

makram93:
#79


makram93:
I have tested this on my linux laptop and the GPU problem does not show anymore

Issue #82, opened 2022-04-07
feat: maintain aspect ratio of images in example dataset


florian-hoenicke:
Make sure the aspect ratio of the images is preserved when creating the showcase datsets
<img width="673" alt="image" src="https://user-images.githubusercontent.com/11627845/162333927-89e1b5fe-2e11-46c9-a227-597c6b125870.png">

PR #94, opened 2022-04-12
feat: thumbnail images to (256, 256)

joschkabraun:
Thumbnail images to (256, 256) to solve #82 and #72


florian-hoenicke (approved):
LGTM

PR #100, opened 2022-04-14
feat: no head deployments

florian-hoenicke:


makram93 (approved):
LGTM

florian-hoenicke:
@joschkabraun thanks for the feedback. I added the requested test which also considers the fine-tuning case and therefore tests different number of pods

joschkabraun (approved):

joschkabraun:
The currently thrown error can be resolved by `moving kwargs.pop('dataset', None)`  from after `run_backend.run(...)` to above it in `run_all_k8s.py`

florian-hoenicke:
thanks for the feedback - fixed

PR #111, opened 2022-04-21
feat: webcam support

makram93:
Added support for video streaming - Webcam. Used two different libraries, however, they fail to work when the frontend application is running on a remote server. In this scenario, the libraries fail to initialize the webcam as it needs the application to have an HTTPS URL.

For remote deployment - `Webcam` options will give error but this is allowed for now!
image


joschkabraun:
LGTM

joschkabraun (approved):

florian-hoenicke:
LGTM - great work

PR #112, opened 2022-04-21
feat: add bff using FastAPI

makram93:
Added initial commit.

TODO:
- [x] Correct the request/response format
- [x] Match the endpoints url
- [ ] Docker containerised - TODO: In a separate PR
- [x] Implement the search - and two use-cases - Image, Texts
- [x] Add unit test cases for the endpoints
- [x] Add integration test

Layout:
image


florian-hoenicke:
I would recommend combining `/search/{query}` with `/search/`
I think it should be one endpoint with different parameters.
Otherwise, it might be confusing to the user.
Instead of doing the binary file upload, you could just have one json endpoint for search which gets base 64 encoded images/ text

florian-hoenicke (changes requested):
I would recommend combining /search/{query} with /search/
I think it should be one endpoint with different parameters.
Otherwise, it might be confusing to the user.
Instead of doing the binary file upload, you could just have one json endpoint for search which gets base 64 encoded images/ text

now/bff/v1/routers/image.py
```diff
@@ -0,0 +1,66 @@
...
+    limit: int = 10,
+):
+    """
+    Retrieve matching images for a given image uri as query
+    """
+    # TODO: Uri or FileUploader?
+    contents = image_file.file.read()
+    query_doc = Document(blob=contents)
+    query_doc.convert_blob_to_image_tensor(224, 224)
+    c = Client(host=host, port=31080)
```
bwanglzu, line 63: what will happen if you have 100 users calling the endpoint? everytime you initialize a new Client instance?

bwanglzu (commented):
left one comment, btw, do you need to handle user authentification?

makram93:
> left one comment, btw, do you need to handle user authentification?

No, we do not need any authentication for now

requirements-dev.txt
```diff
@@ -1,2 +1,3 @@
 pytest
 pytest-mock==3.7.0
+pytest==0.75.2
```
joschkabraun, line 3: shall we remove the first line then?

now/bff/v1/routers/image.py
makram93: We do not have a similar scenario as Finetuner API. As in our case, Frontend and BFF will be deployed for each user running jina now and then they can call the endpoints for their deployed Flow (either local or GKE).

now/bff/v1/routers/image.py
```diff
@@ -0,0 +1,66 @@
...
+    host: str = 'localhost',
+    limit: int = 10,
+):
+    """
+    Retrieve matching images for a given image uri as query
+    """
+    # TODO: Uri or FileUploader?
+    contents = image_file.file.read()
+    query_doc = Document(blob=contents)
+    query_doc.convert_blob_to_image_tensor(224, 224)
```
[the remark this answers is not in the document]
makram93: Thanks for pointing this out. I have removed it and unified the endpoints and now it simply accepts base64 encoded image/text. So, whatever resizing is needed should be done at the client/Flow and not be handled in the endpoint logic

requirements-dev.txt
makram93: Good observation

now/bff/v1/routers/image.py
```diff
@@ -0,0 +1,66 @@
...
+    response_model=JinaResponseModel,
+    summary='Add more data to the indexer',
+)
+def index(data: List[str], host: str):
+    """
+    Append the image data to the indexer
+    """
+    index_docs = DocumentArray()
+    for text in data:
+        index_docs.append(Document(text=text))
```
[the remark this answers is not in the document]
makram93: Which part? It is intentional `text=text`

florian-hoenicke:
I ran it locally and checked the response of the bff.
It returns a DocumentArray at the moment.
However the ResponseModel should also be adjusted to the specific usecase:
For Instance, when using text as output modality, the user would expect a responsemodel like:
```
[
  {
    id: 3457483,
    text: first text,
    tags: {author: name, url:https://...}
  },
  {
    text:...
]
```
image

PR #120, opened 2022-04-27
feat: data loading music

slettner:


now/data_loading/data_loading.py
```diff
@@ -1,18 +1,65 @@
...
+    Modalities,
+    Qualities,
+)
 from now.data_loading.convert_datasets_to_jpeg import to_thumbnail_jpg
-from now.dialog import QUALITY_MAP, Modalities
+from now.data_loading.utils import load_mp3
+from now.dialog import UserInput
 from now.utils import download, sigmap
 
+IMAGE_EXTENSIONS = {'.png', '.jpg', '.jpeg', '.tiff', '.bmp', '.gif'}
```
florian-hoenicke, line 24: let's use all image extensions like `webp` and also more audio extensions. Maybe, there is a library which knows all. Or we assume that all files in a folder are images or audio, and we just try to parse them and if parsing fails, we could ignore the related files
Moreover, there are a lot of different name conventions for each file format like `.JPEG` or `.PNG`

now/data_loading/data_loading.py
```diff
@@ -1,18 +1,65 @@
...
+
+    if not user_input.is_custom_dataset:
+        print('⬇  Download DocArray')
+        url = get_dataset_url(
+            user_input.data, user_input.quality, user_input.output_modality
+        )
+        da = _fetch_da_from_url(url)
+
+    else:
+        if user_input.custom_dataset_type == DatasetTypes.DOCARRAY:
```
florian-hoenicke, line 45: Really nice to have these enum like types. Makes the code much more readable

now/data_loading/data_loading.py
```diff
@@ -33,149 +80,94 @@ def _fetch_da_from_url(
...
+        exit(1)
 
-    new_da = DocumentArray()
-    for i, d in enumerate(da):
-        new_doc = deepcopy(d)
-        new_doc.id = str(uuid.uuid4())
-        new_da.append(new_doc)
-    return new_da
 
+def _folder_contains_only(path: str, extensions: Set) -> bool:
```
florian-hoenicke, line 93: I like that you split everything into smaller functions

now/data_loading/utils.py
```diff
@@ -17,3 +22,19 @@ def upload_to_gcloud_bucket(project: str, bucket: str, location: str, fname: str
 
     tensor = bucket.blob(location + '/' + fname)
     tensor.upload_from_file(content, timeout=7200)
+
+
+def load_mp3(path_to_mp3: str) -> Tuple[np.ndarray, int]:
+    sound = AudioSegment.from_mp3(file=path_to_mp3)
```
florian-hoenicke, line 28: do we have other audio data types as well.
If this will be handled in another pr it is also fine

now/run_backend.py
```diff
@@ -62,16 +63,7 @@ def run(user_input: UserInput, is_debug, tmpdir, kubectl_path: str):
...
-        user_input.output_modality,
-        user_input.data,
-        user_input.quality,
-        user_input.is_custom_dataset,
-        user_input.custom_dataset_type,
-        user_input.dataset_secret,
-        user_input.dataset_url,
-        user_input.dataset_path,
-    )
+    dataset = load_data(user_input)
```
florian-hoenicke, line 66: great to just pass the user_input instead of all single attributes

tests/unit/data_loading/test_data_loading.py
```diff
@@ -0,0 +1,129 @@
+""" This suite tests the data_loading.py module """
```
florian-hoenicke, line 1: great that this is tested

florian-hoenicke (changes requested):

now/data_loading/data_loading.py
slettner: Ok, so we remove the check and just try to parse? Makes sense. For audio, we planned to load the file only to save bandwidth, should we try to parse the audio to check if its valid as well (would be costly).
Alternatively, we could filter the files to the valid extensions and set the filter to the supported types of PIL for image (I think this is what docarray uses) and librosa for audio? 

What do you think?

now/data_loading/utils.py
slettner: We can use librosa which support 6 types or so. I would happy to do this in another PR. 
Created an issue here https://github.com/jina-ai/now/issues/144

This will go into the executor because we load the audio there.

now/data_loading/data_loading.py
slettner: I just made the following changes:

- images and audio files are loaded best-effort. I.e. if a uri can't be loaded it is ignored
- for audio file, only the binary file data is stored but only after the format is checked so that we don't load arbitrary files from folder

The audio files are later loaded in the executor. This automatically extends the support of audio files to everything that ffmpeg support. 

Changes 6c5840bc8b3bc627a3837e4dad845db3df5d280b

slettner:
Closes https://github.com/jina-ai/now/issues/144

now/data_loading/utils.py
slettner: See the comment above also.

florian-hoenicke: okay, sounds good to me

now/data_loading/data_loading.py
florian-hoenicke: I like this change. That means it uses all common image formats of the user

florian-hoenicke:
LGTM

PR #128, opened 2022-04-28
feat: update logo shown on frontend

makram93:
I have updates the torch requirements to `torch>=1.10.2` and the logo looks like this:

image

Also, we get userwarnings from Jina which is unwanted as shown here:
image
But now it no more shows if the program is run via `jina-now start`. `jina now start` will still give this error as the control is first with jina and then with NOW
image


setup.py
```diff
@@ -35,7 +35,7 @@
     ]
 sys_platform = platform.system().lower()
 # if sys_platform == 'darwin':
-torch_requirement = ['torch==1.10.2']
+torch_requirement = ['torch>=1.10.2']
```
joschkabraun, line 38: Do we support torch>=1.10.2?

joschkabraun (approved):
LGTM

joschkabraun (approved):

requirements.txt
```diff
@@ -1,4 +1,4 @@
-torchvision==0.11.3
+torchvision>=0.11.3
```
florian-hoenicke, line 1: should be handled in a different pr next time

florian-hoenicke (changes requested):
Great initiative updating the logo.
However, the logo should be jpg or SVG.

florian-hoenicke (changes requested):
why are the banners added?

makram93:
> Great initiative updating the logo. However, the logo should be jpg or SVG.

Are you referring to the `favicon.png`? I cross-referenced with other repo - docarray and jina, both of which have favicon and banner in png format.



> why are the banners added?

It is not important but to have uniformity with other repos such as [here](https://github.com/jina-ai/jina/blob/master/docs/conf.py#L168), I added it for NOW. Usage of this can be found [here](https://github.com/wpilibsuite/sphinxext-opengraph/blob/main/README.md#:~:text=ogp_image,defined%20by%20ogp_site_name.).

requirements.txt
makram93: I will revert this for this PR

now/run_all_k8s.py
```diff
@@ -58,7 +58,7 @@ def run_k8s(os_type: str = 'linux', arch: str = 'x86_64', **kwargs):
             **kwargs,
         )
         with tempfile.TemporaryDirectory() as tmpdir:
-            docker_frontend_tag = '0.0.7'
+            docker_frontend_tag = '0.0.8-dev-update-logo'
```
florian-hoenicke, line 61: I think this should be just 0.0.8.
Or correct me please @joschkabraun if I'm mistaken.

florian-hoenicke (changes requested):

florian-hoenicke (approved):
LGTM

now/run_all_k8s.py
makram93: I have updated the version. It should be `0.0.8` now

PR #121, opened 2022-04-27
feat(frontend-cd): added frontend deployment to ci

joschkabraun:
closes #116

PR #140, opened 2022-04-29
Feat clean text data

### LINKED ISSUE #119: feat: clean text datasets
This is a follow-up ticket to improve the quality of the image-to-text search

image
image

1. blue: replace sware words with `[_]` or with 💀 
2. red: remove redundant results
3. yellow: artist should be less prominent. Please make the font size smaller
4. dark green: some results are not complete sentences - If there is something we can do about it, let's filter them out
5. light green:  Sentences should have punctuation at the end
6. violet: filter results with special characters

joschkabraun:
Closes #119.

Note need to rerun CI after https://github.com/jina-ai/now/pull/121 has been merged to build and push new frontend.

### DIFF
--- a/now/data_loading/build_datasets.py
+++ b/now/data_loading/build_datasets.py
@@ -508,17 +508,30 @@ def _build_lyrics(
         lyrics = lyrics.sample(frac=1)
 
     # create sentences from lyrics
-    data = []
-    for idx, row in lyrics.iterrows():
+    data, all_sentences = [], []
+    for idx, row in tqdm(lyrics.iterrows()):
         if 0 < max_size <= len(data):
             break
         row = row.to_dict()
         _sentences = row.pop('Lyric').split('\n')
-        # filter empty and repeating sentences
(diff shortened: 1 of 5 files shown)


now/data_loading/build_datasets.py
```diff
@@ -508,17 +508,30 @@ def _build_lyrics(
         lyrics = lyrics.sample(frac=1)
 
     # create sentences from lyrics
-    data = []
-    for idx, row in lyrics.iterrows():
+    data, all_sentences = [], []
```
florian-hoenicke, line 511: I would recommend not to use multi-assignments since single statements will have better readability. And source control works better with single statements. But in this case it is okay I guess

florian-hoenicke:
Good work: Removing the inappropriate words works 👍 
image

now/data_loading/build_datasets.py
joschkabraun: That makes sense, thank you!

PR #148, opened 2022-05-02
refactor: make dialog steps more descriptive

slettner:


now/dialog.py
```diff
@@ -130,12 +132,34 @@
...
+        user_input = _configure_dataset_text(user_input, **kwargs)
     else:
-        return _configure_dataset_music(user_input, **kwargs)
+        user_input = _configure_dataset_music(user_input, **kwargs)
+
+    dataset = user_input.data
+    if dataset in AVAILABLE_DATASET[user_input.output_modality]:
+        user_input.is_custom_dataset = False
+        if user_input.output_modality == Modalities.MUSIC:
+            return user_input
```
joschkabraun, line 150: I would move the logic around whether `_configure_quality` is called into `configure_user_input` as it reduces need to maintain and write code for this. Also, it doesn't fit to the name `_configure_dataset`.

now/dialog.py
```diff
@@ -130,12 +132,34 @@
         **kwargs,
     )
     user_input.output_modality = modality
-    if modality == Modalities.IMAGE:
-        return _configure_dataset_image(user_input, **kwargs)
-    elif modality == Modalities.TEXT:
-        return _configure_dataset_text(user_input, **kwargs)
+    return user_input
```
joschkabraun, line 135: You can remove the return values as the object is modified in-place.

now/dialog.py
```diff
@@ -222,26 +246,6 @@
...
-        else:
-            _parse_custom_data_from_cli(dataset, user_input)
-            if user_input.output_modality == Modalities.MUSIC:
-                return _configure_cluster(user_input, **kwargs)
-            else:
-                return _configure_quality(user_input, **kwargs)
-
-
 def _configure_custom_dataset(user_input: UserInput, **kwargs) -> UserInput:
     def configure_docarray() -> UserInput:
```
joschkabraun, line 250: I would move all those functions, which are in the if-branches called, into the if-branches as they're short and not used anywhere else. Also they will be even shorter when `_configure_quality` is called in `configure_user_input`

joschkabraun (changes requested):
Nice PR!

slettner:
Thanks for the feedback @joschkabraun - very constructive.
I summarized the your suggestions in this commit https://github.com/jina-ai/now/pull/148/commits/f262d137a979e406dbf335377d32467e574278bb

now/dialog.py
```diff
@@ -301,14 +264,31 @@
...
     if custom_dataset_type == 'docarray':
-        return configure_docarray()
+        dataset_secret = _prompt_value(
+            name='dataset_secret',
+            prompt_message='Please enter your docarray secret.',
+            prompt_type='password',
+        )
+        user_input.dataset_secret = dataset_secret
+
     if custom_dataset_type == 'url':
```
joschkabraun, line 274: just saw those are actually if ... elif ... elif

now/dialog.py
```diff
@@ -222,63 +241,7 @@
     return user_input
```
joschkabraun, line 241: can remove return value and simply say `user_input.sandbox = _prompt_value(...)`

joschkabraun:
Nice PR! This really helps with understanding and maintaining the dialog.

now/dialog.py
slettner: Good catch, makes it even more concise!

now/dialog.py
slettner: https://github.com/jina-ai/now/pull/148/commits/6a9f8c5651406e504350b2e6e01aa54387cb0d33

joschkabraun (approved):
LGTM!

PR #150, opened 2022-05-02
fix: waiting at zero percent

florian-hoenicke:


now/deployment/flow.py
```diff
@@ -160,19 +160,21 @@ def deploy_flow(
         tmpdir,
         kubectl_path=kubectl_path,
     )
-    print(
-        f'▶ indexing {len(index)} documents - if it stays at 0% for a while, it is all good - just wait :)'
+    print(f'▶ indexing {len(index)} documents')
+    client = Client(host=gateway_host, port=gateway_port)
+    request_size = 64
+
+    progress_bar = (
```
slettner, line 167: Is it that now the 0% do not appear, but the progress bar start only after the flow is ready to take requests?

florian-hoenicke: There was a bug before in the core where the gateway said that everything is ready even though the clip encoder was not up and running. That required us to make a lot of retries for the first request resulting in a long time waiting at 0%

slettner: I understand, thanks for clarifying :)

PR #151, opened 2022-05-02
fix: deactivate sandbox

florian-hoenicke:
sometimes, sandbox is hanging. Therefore, we need to deactivate it for now. 
We were still not able to reproduce the issue in a deterministic way.

### DIFF
--- a/now/dialog.py
+++ b/now/dialog.py
@@ -209,16 +209,18 @@ def _configure_dataset_music(user_input: UserInput, **kwargs):
 
 
 def _configure_sandbox(user_input: UserInput, **kwargs):
-    sandbox = _prompt_value(
-        name='sandbox',
-        prompt_message='Use Sandbox to save memory? (process data on our servers)',
-        choices=[
-            {'name': '⛔ no', 'value': False},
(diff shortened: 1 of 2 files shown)


slettner (approved):
LGTM!

joschkabraun:
Also, LGTM! Do we have an issue somewhere to track the bug?

now/dialog.py
```diff
@@ -231,6 +231,7 @@ def _configure_sandbox(user_input: UserInput, **kwargs):
         ],
         **kwargs,
     )
+    user_input.sandbox = False
```
makram93, line 234: But this will still prompt the user with the sandbox option and then we hard code it to False. Shouldn't we comment it and not even prompt?

florian-hoenicke: Yes, thanks a lot. I forgot to comment this after the most recent commit.
Nice catch

makram93 (approved):
LGTM!

mapleeit:
Can we revert this since Sandbox has been fixed with jina >= 3.3.24

Seemed a grpc version problem in Core.

PR #153, opened 2022-05-03
Test no spinning during tests

florian-hoenicke:


now/log/log.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def __enter__(self):
+        if not TEST:
+            return super().__enter__()
+        else:
+            return self
+
+    def __exit__(self, exc_type, exc_val, traceback):
+        if not TEST:
+            return super().__exit__(exc_type, exc_val, traceback)
```
makram93, line 22: Why don't we keep it in the `utils` file?

florian-hoenicke: Usually, utils files are not so good to have, since it is not very specific about what they are doing. This `log.py` file could be used to collect more logging relevant code in the future

PR #179, opened 2022-05-19
fix: update the cli args

makram93:
- Update the CLI args based on the recent changes
- Also updated the CI/CD allowing debugging by `ssh`'ing into the runner


joschkabraun:
LGTM!

PR #188, opened 2022-05-22
refactor: separated apps

florian-hoenicke:
The logic of our apps is currently split all over the code base. This pr aims to bundle the logic within single classes wich allows parallel development on the applications


now/apps/base/app.py
```diff
@@ -0,0 +1,140 @@
...
+from docarray import DocumentArray
+
+from now.datasource.datasource import Datasource
+
+
+class AppConfig:
+    env: Dict
+
+
+class JinaNOWApp:
```
slettner, line 13: We could use the `abc` package for the `JinaNOWApp`
It is an abstract class and using the decorator users of the interface know at instance creation time if they missed to implement one of the abstract method.
```
import abc

class JinaNowApp(object):
    __metaclass__ = abc.ABCMeta
    
    @abc.abstractmethod
    def method_to_implement(self):
        raise NotImplementedError('method_to_implement')
```

now/apps/base/app.py
```diff
@@ -0,0 +1,140 @@
...
+
+    @property
+    def description(self) -> str:
+        """
+        Short description of the app.
+        """
+        return 'Jina NOW app'
+
+    @property
+    def input_modality(self) -> str:
```
slettner, line 34: We could typehint the `Modalities` Enum here, maybe it motivates user to use our Enum or add their modality to it if it misses.

now/apps/base/app.py
```diff
@@ -0,0 +1,140 @@
...
+    def flow_yaml(self, value: str):
+        """
+        Set the flow yaml.
+        """
+        self._flow_yaml = value
+
+    @property
+    def bff(self) -> Optional[str]:
+        """
+        Used to configure the bff which is used to define input and
```
slettner, line 65: This sentence looks incomplete :)
It is also not clear to me what is meant here. Is it the url of the docker image to use for the bff?
Maybe the naming can be more explicit.

now/apps/base/app.py
```diff
@@ -0,0 +1,140 @@
...
+        :return: the path of the bff
+        """
+        return None
+
+    @property
+    def playground(self) -> Optional[str]:
```
slettner, line 71: Same here, what string is expected from this?

now/apps/base/app.py
```diff
@@ -0,0 +1,140 @@
...
+
+    @property
+    def example_datasource(self) -> List[Datasource]:
+        """
+        Get a list of example datasets for the app.
+
+        """
+        return []
+
+    def check_requirements(self) -> bool:
```
slettner, line 108: Can we have a separate property that just returns a list of requirements which the user implements?
I assume itself the check is the same independent of the requirement and can be implemented in the base class.

```
class JinaNOWApp:
  def __init__(self):
      self._check_requirements()
  
  def _check_requirements(self):
      for req in self.requirements: #implemented by the user
          assert req.exists

    @property
    @abstractmethod
    def requirements(self) -> List[str]:
        raise NotImplementedError()

class MyApp(JinaNOWApp):
    @property
    def requirements(self) -> List[str]:
        return ['torch']
```

now/apps/image_to_image/app.py
```diff
@@ -0,0 +1,29 @@
...
+    @property
+    def input_modality(self) -> str:
+        return Modalities.IMAGE
+
+    @property
+    def output_modality(self) -> str:
+        return Modalities.IMAGE
+
+    @property
+    def options(self) -> List[Dict]:
```
slettner, line 25: I like this design with the options now :)

now/apps/base/app.py
slettner: We could also add the context manager interface to the app with __enter__ and __exit__ that wrap setup and cleanup. Optional I think

now/apps/base/app.py
slettner: ah nevermind this is not referring to the python packages but other dependencies. Then the suggestion might not make sense

now/constants.py
```diff
@@ -1,36 +1,31 @@
...
 
-class DatasetTypes:
+
+class DatasetTypes(BetterEnum):
     DEMO = 'demo'
     PATH = 'path'
     URL = 'url'
     DOCARRAY = 'docarray'
 
-    @classmethod
```
slettner, line 20: Nice!

now/data_loading/data_loading.py
```diff
@@ -21,7 +21,7 @@
 from now.utils import download, sigmap
 
 
-def load_data(user_input: UserInput) -> DocumentArray:
+def load_data(modality, user_input: UserInput) -> DocumentArray:
```
slettner, line 24: I think we should avoid to use the word `modality` for any argument because it is not clear if input or output is meant from now on. Not sure if it is applicable here.
And aren't the modalities attached to the `user_input` object?

now/apps/image_to_text/app.py
```diff
@@ -0,0 +1,29 @@
...
+
+from docarray import DocumentArray
+from now_common import options
+
+from now.apps.base.app import JinaNOWApp
+from now.constants import Modalities
+from now.run_backend import finetune_flow_setup
+
+
+class image_to_text(JinaNOWApp):
```
slettner, line 11: Shouldn't the apps implement `example_datasource`?

slettner: And `flow_yaml`?

now/datasource/example.py
```diff
@@ -0,0 +1,22 @@
...
+from now.datasource.datasource import Datasource
+
+
+class ExampleDatasource(Datasource):
+    def __init__(self, ds_id, display_name, modality):
+        self.ds_id = ds_id
+        self.display_name = display_name
+        self.modality = modality
+
+    def get_data(self, quality: str) -> DocumentArray:
```
slettner, line 13: This signature misses the `mapping` arg from the baseclass.
The baseclass interface is not really guiding the user a lot at the moment. 
Maybe we should start with an explicit ObjectStorageDataset or something like that because all our app are using that atm. and abstract the interface furhter once we work on the database table column mapping stuff?

now/deployment/flow.py
```diff
@@ -122,78 +99,27 @@ def deploy_k8s(f, ns, num_pods, tmpdir, kubectl_path):
...
+    ns: str,
+    env_file: str,
     index: DocumentArray,
     tmpdir: str,
     kubectl_path: str,
 ):
     from jina import Flow
     from jina.clients import Client
 
-    finetuning = finetune_settings.perform_finetuning
```
slettner, line 185: Great change, this had nothing to do with `deploy_flow`!

now/dialog.py
```diff
@@ -49,25 +45,42 @@ class UserInput:
...
+    quality: Optional[Qualities] = Qualities.MEDIUM
 
     # cluster related
     cluster: Optional[str] = None
     create_new_cluster: Optional[bool] = False
     deployment_type: Optional[str] = None
 
 
-def configure_user_input(**kwargs) -> UserInput:
+def _configure_app_options(app_instance: JinaNOWApp, user_input, **kwargs):
```
slettner, line 56: Slik!

now/run_backend.py
```diff
@@ -14,24 +47,69 @@ def run(user_input: UserInput, tmpdir, kubectl_path: str):
...
     """
-    dataset = load_data(user_input)
-
-    finetune_settings = parse_finetune_settings(user_input, dataset)
-    if finetune_settings.perform_finetuning:
-        print(f'🔧 Perform finetuning!')
-        finetune_settings.finetuned_model_name = finetune_now(
-            user_input, dataset, finetune_settings, kubectl_path
+    dataset = load_data(app_instance.output_modality, user_input)
+    env = app_instance.setup(dataset, user_input, kubectl_path)
```
slettner, line 51: My IDE complains that `user_input` expects a `dict` not `UserInput`. Can we just change the signature?

slettner: If we use a context manager for the app we can manage the context for the tmpdir in the app. Just an idea..
```
with app_instance(dataset, user_input, kubectl_path) as app:
    app.tmpdir
```
Should the `load_data` call be part of `setup`?

slettner:
Great PR, a lot of improvements! I ran one image to text example and it runs smooth 🙌🏼

now/apps/base/app.py
```diff
@@ -0,0 +1,140 @@
...
+        """
+        Used to configure the flow yaml in the Jina NOW app.
+        :return: either the path, to the yaml or the yaml content.
+        """
+        return self._flow_yaml
+
+    @flow_yaml.setter
+    def flow_yaml(self, value: str):
+        """
+        Set the flow yaml.
```
makram93, line 58:
```suggestion
        Set the flow yaml. Either the path to the yaml or the yaml content.
```

now/apps/base/app.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def check_requirements(self) -> bool:
+        """
+        Returns true if all requirements on the system are satisfied. Else False.
+        """
+        return True
+
+    # TODO Remove kubectl_path. At the moment, the setup function needs kubectl because of finetuning a custom
+    #  dataset with local deployment. In that case, inference is done on the k8s cluster.
+    def setup(self, da: DocumentArray, user_config: Dict, kubectl_path: str) -> Dict:
```
makram93, line 116:
```suggestion
    def setup(self, da: DocumentArray, user_config: UserInput, kubectl_path: str) -> Dict:
```

now/apps/base/app.py
```diff
@@ -0,0 +1,140 @@
...
+            - finetune a model + push the artifact
+            - notify other services
+            - check if starting the app is currently possible
+        :param da:
+        :param user_config: user configuration based on the given options
+        :return: dict used to replace variables in flow yaml and to clean up resources after the flow is terminated
+        """
+        pass
+
+    def cleanup(self, app_config: dict) -> None:
```
makram93, line 130: I am not sure if you are referring to this class `AppConfig`. If so, then it is best to use `AppConfig` as the typehint

now/data_loading/data_loading.py
```diff
@@ -46,8 +39,11 @@ def load_data(user_input: UserInput) -> DocumentArray:
             da = _fetch_da_from_url(user_input.dataset_url)
         elif user_input.custom_dataset_type == DatasetTypes.PATH:
             print('💿  Loading DocArray from disk')
```
makram93, line 41:
```suggestion
            print('💿  Loading files from disk')
```

now/apps/base/app.py
florian-hoenicke: Thanks, done

now/apps/base/app.py
florian-hoenicke: Thanks, changed

now/apps/base/app.py
florian-hoenicke: Also not well defined currently

now/apps/image_to_image/app.py
florian-hoenicke: yeahhh :)

now/constants.py
florian-hoenicke: :)

now/data_loading/data_loading.py
florian-hoenicke: Good point, I changed it to output_modality.
The user is not defining the modalities anymore themselves.
Input and output modality is implicitly defined via the app they select.

now/apps/image_to_text/app.py
florian-hoenicke: You are absolutely right. I added a comment to `example_datasource`. For this pr I wanted to introduce the concept already but it is not used. A follow-up pr will handle this.

requirements-test.txt
```diff
@@ -1,2 +1,10 @@
 pytest
-pytest-mock
\ No newline at end of file
+pytest-mock
+pytest-timeout
```
joschkabraun, line 3: I thought we removed this file as this is actually handled in `setup.py` or are we actually using it?

joschkabraun (commented):
Really nice PR and I like the refactoring you suggested!

now/dialog.py
florian-hoenicke: :)

now/apps/base/app.py
florian-hoenicke: ✅

now/apps/base/app.py
florian-hoenicke: Thanks. Done

now/apps/base/app.py
florian-hoenicke: Thanks, I removed the class since it is not used

now/dialog.py
```diff
@@ -49,25 +45,42 @@ class UserInput:
...
+            **kwargs,
+        )
+        setattr(user_input, option['name'], val)
+    # _configure_quality(user_input, **kwargs)
+
+
+def _construct_app(user_input) -> JinaNOWApp:
+    return getattr(
+        importlib.import_module(f'now.apps.{user_input.app}.app'), f'{user_input.app}'
+    )()
```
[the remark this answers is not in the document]
florian-hoenicke: Thanks for the suggestion. Makes sense. Done

requirements-test.txt
florian-hoenicke: Thanks, good point. We should remove the redundancy.
I created a pr tackling this separately:
https://github.com/jina-ai/now/pull/221

now/finetuning/settings.py
```diff
@@ -4,13 +4,14 @@
...
 
-from now.constants import DemoDatasets, Modalities, Qualities
+from now.constants import Apps, DemoDatasets, Qualities
 from now.dialog import UserInput
 
 TUNEABLE_DEMO_DATASETS = {
-    Modalities.IMAGE: [DemoDatasets.DEEP_FASHION, DemoDatasets.BIRD_SPECIES],
-    Modalities.TEXT: [],
-    Modalities.MUSIC: [
+    Apps.TEXT_TO_IMAGE: [DemoDatasets.DEEP_FASHION, DemoDatasets.BIRD_SPECIES],
```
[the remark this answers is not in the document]
florian-hoenicke: You got me. I tried to do this in a later refactoring but you are right. This should be done here as well. We should aim for nothing app-specific within our main code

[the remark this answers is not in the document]
florian-hoenicke: done

now/finetuning/run_finetuning.py
```diff
@@ -53,7 +53,7 @@ def finetune_now(
 
     :return: Path to the tuned model.
     """
-    if user_input.output_modality == Modalities.MUSIC:
+    if user_input.app == Apps.MUSIC_TO_MUSIC:
```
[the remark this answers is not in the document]
florian-hoenicke: Nice catch. I moved it to the setup function.
I think there is no general property we could have for this for now since each app might do the fine-tuning a bit differently. But it is good not to handle this within the main code anymore.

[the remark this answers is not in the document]
florian-hoenicke: done

now/finetuning/embeddings.py
```diff
@@ -37,3 +38,27 @@ def embed_now(user_input: UserInput, dataset: DocumentArray, kubectl_path: str):
...
         dataset[doc.id].embedding = doc.embedding
+
+
+def get_encoder_config(user_input: UserInput) -> _ExecutorConfig:
+    """
+    Gets the correct Executor running the pre-trained model given the user configuration.
+    :param user_input: Configures user input.
+    :return: Small data-transfer-object with information about the executor
+    """
+    if user_input.app == Apps.MUSIC_TO_MUSIC:
```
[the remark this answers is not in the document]
florian-hoenicke: Thanks, I moved it to the apps. You are right. There should be no logic regarding the apps in our core code

florian-hoenicke:
Thanks for the review!

PR #198, opened 2022-05-23
Execute CI tests in parallel

### LINKED ISSUE #189: test: split up integration tests
When creating a pr it can take quite some time until we know if an... tbd

joschkabraun:
Closes #189.


.github/workflows/ci.yml
```diff
@@ -167,33 +167,21 @@ jobs:
...
   core-test:
-    needs: prep-testbed
     runs-on: ubuntu-latest
+    needs: update-bff-playground-if-required
     strategy:
       fail-fast: false
       matrix:
         python-version: [3.7]
-        test-path: ${{fromJson(needs.prep-testbed.outputs.matrix)}}
+        group: [1, 2, 3, 4]     # set to the number of end-to-end tests
```
makram93, line 177: What does this group do? And how are the tests (unit + integration) divided?

.github/workflows/ci.yml
```diff
@@ -210,38 +198,48 @@ jobs:
...
+      - name: Download all coverage artifacts
+        uses: actions/download-artifact@v2
+      - name: Run coverage
+        run: |
+          coverage combine coverage*/.coverage*
+          coverage xml
+      - name: Upload coverage from test to Codecov
+        uses: codecov/codecov-action@v2
+        with:
+          fail_ci_if_error: false
```
makram93, line 242:
```suggestion
```
It already defaults to `false` as mentioned [here](https://github.com/codecov/codecov-action#:~:text=Specify%20if%20CI%20pipeline%20should%20fail%20when%20Codecov%20runs%20into%20errors%20during%20upload.%20Defaults%20to%20false) so we may not need to explicitly specify here:

makram93, line 242:
```suggestion
```
It already defaults to `false` as mentioned [here](https://github.com/codecov/codecov-action#:~:text=Specify%20if%20CI%20pipeline%20should%20fail%20when%20Codecov%20runs%20into%20errors%20during%20upload.%20Defaults%20to%20false) so we may not need to explicitly specify here:

requirements-dev.txt
```diff
@@ -1,2 +1,4 @@
 pytest
 pytest-mock==3.7.0
+pytest-split==0.8.0
+pytest-cov
```
makram93, line 4: I think we can omit this as this file is nowhere used. In my next PR I will delete this file as we only make use of
`pip install --no-cache-dir ".[full,test]"` in the CI/CD

makram93, line 4: I think we can omit this as this file is nowhere used. In my next PR I will delete this file as we only make use of
`pip install --no-cache-dir ".[full,test]"` in the CI/CD

.test_durations
```diff
@@ -0,0 +1,94 @@
+{
```
makram93, line 1: Can you also add a note here - 
`When there are major changes in the test cases (execution time) compared to what's stored in this file, it's recommended to update the duration information with --store-durations to ensure that the splitting is in balance.`

`pytest --store-durations`

makram93, line 1: Can you also add a note here - 
`When there are major changes in the test cases (execution time) compared to what's stored in this file, it's recommended to update the duration information with --store-durations to ensure that the splitting is in balance.`

`pytest --store-durations`

makram93 (commented):
A very nice PR. I learned about parallel execution today :)
I have left some minor comments.

.github/workflows/ci.yml
joschkabraun: They are the indexes of groups to run the tests in parallel. The tests are split into the groups such that the time needed to complete them is roughly the same. For that `.test_durations` is used.

.github/workflows/ci.yml
joschkabraun: Good point. Thank you!

.test_durations
joschkabraun: Good idea! Added it to `tests/README.md`

joschkabraun:
Thank you Kalim for the feedback!

makram93 (approved):
Once the CI passes, it is good to go!

PR #204, opened 2022-05-25
feat: use prepared linear heads for music

### LINKED ISSUE #203: feat: separate ds for music finetuning
For the modality Music, we need separate index and training datasets, otherwise the index would contain many duplicates between different finetune classes because songs can appear in multiple classes. This redundancy makes the index unnecessary large. 

Either we store separate datasets for indexing and finetuning or we skip the finetuning entirely and use a pre-trained Linear Head from the hub.

slettner:
close #203


florian-hoenicke (approved):
LGTM

PR #206, opened 2022-05-25
feat: add router for modalities

makram93:
Separate bff for different modalities

### DIFF
--- a/deployment/bff/app/app.py
+++ b/deployment/bff/app/app.py
@@ -30,6 +30,9 @@ def get_app_instance():
             'author': AUTHOR,
             'email': EMAIL,
         },
+        docs_url='/docs',
+        redoc_url='/redoc',
+        openapi_url='/openapi.json',
     )
 
     @app.get('/ping')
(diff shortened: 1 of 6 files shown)


deployment/bff/app/app.py
```diff
@@ -28,9 +30,6 @@ def build_app():
             'author': AUTHOR,
             'email': EMAIL,
         },
```
joschkabraun, line 32: Previously, I needed to include the parameters `docs_url`, `redoc_url` and `openapi_url` in the constructor in order to make the docs accessible through the Nginx proxy (https://stackoverflow.com/a/67435809). I couldn't get the docker-compose `docker-compose.yml` started just now to check if it's working.

joschkabraun (changes requested):
Nice PR!
After investigating the comment, this LGTM :)

deployment/bff/app/app.py
makram93: It works now - though the url will not be `...:../api/docs` but rather `...:../api/v1/modality/docs`

PR #214, opened 2022-05-27
feat: integrate finetune api

### LINKED ISSUE #213: feat: integrate with finetuner.fit
At the moment, we use the community version of finetuner. However, the finetuner.fit api is ready to be used.
Migrating gives us the following benefits:
- we won't need the local dependencies like PyTorch anymore since the finetuner client is lean.
- we dogfeed our own service
- finetuner.fit comes with improvements of the training and new futures like training with multiple modalities.

slettner:
close #213

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 torchvision==0.11.3
 transformers==4.16.2
-finetuner==0.4.0
+finetuner==0.4.1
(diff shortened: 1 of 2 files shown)


slettner:
Replaced by https://github.com/jina-ai/now/pull/260

PR #216, opened 2022-05-30
Feat: add playground survey

fogx:
add a survey link in the playground to find out what users like/don't like

### DIFF
--- a/now/data_loading/data_loading.py
+++ b/now/data_loading/data_loading.py
@@ -23,7 +23,7 @@ from now.utils import download, sigmap
 
 def load_data(user_input: UserInput) -> DocumentArray:
     """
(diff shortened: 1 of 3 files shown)


now/data_loading/data_loading.py
```diff
@@ -23,34 +23,34 @@
 
 def load_data(user_input: UserInput) -> DocumentArray:
     """
-    Based on the user input, this function will pull the configured DocArray.
+    Based on the user input, this function will pull the configured DocArray dataset.
```
makram93, line 26: Always try to merge the `main` branch to your feature branch and if there's any conflict resolve it locally before pushing. That way other commits changes will not show up

PR #220, opened 2022-05-30
fix: convert tensor first

### LINKED ISSUE #210: Fix: Sample Images of nft-monkey aren't searchable
In neither local nor remote deployment are the nft-monkey sample images searchable:

image

When logging into the playground pod I came to the hypothesis that there is an issue with the samples images we uploaded for this dataset and how the frontend loads them.
As one can see if we try to convert uri to blob from the document we access a file which doesn’t actually exist. The solution is that we need to set the `mime_type = image` and convert tensor to blob. But implementing this leads to no search results from returned from BFF.
Screenshot 2022-05-25 at 17 13 07
Screenshot 2022-05-25 at 17 12 04
Screenshot 2022-05-25 at 17 13 07

makram93:
closes #210

### DIFF
--- a/deployment/playground/playground.py
+++ b/deployment/playground/playground.py
@@ -20,7 +20,7 @@ root_data_dir = (
     'https://storage.googleapis.com/jina-fashion-data/data/one-line/datasets/'
 )
 
-ds_set = {
+ds_set = [
     'nft-monkey',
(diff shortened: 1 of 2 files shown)


florian-hoenicke (approved):
LGTM

PR #226, opened 2022-05-31
fix: update the endpoints based on app

### LINKED ISSUE #225: fix: update bff endpoints based on app

makram93:
closes #225


deployment/playground/playground.py
```diff
@@ -63,14 +63,15 @@ def deploy_streamlit():
     )
     OUTPUT_MODALITY = query_parameters.get('output_modality')[0]
     INPUT_MODALITY = query_parameters.get('input_modality')[0]
+    APP = query_parameters.get('app')[0]
```
joschkabraun, line 66: Isn't `APP = f"{INPUT_MODALITY}-to-{OUTPUT_MODALITY}"` or the other way round?

makram93: That's true. I can make use of `imput_modality` and `output_modality`. Though, I have updated the `playground_url` in `run_all_k8s` to incorporate this `app` query param. I think I will change it and make use of input and output modality

florian-hoenicke:
LGTM

PR #227, opened 2022-05-31
feat: bump version

florian-hoenicke:


makram93:
We do not to update this file if the version release is just a minor one (or sequential), i.e, from `0.0.7` to `0.0.8`. Only when there is a major version change (non-sequential) then we should update this file, i.e., from `0.0.7` to `0.1.0`. 
To make a minor release, go to `Actions` and run the `Manual release` GitHub actions. Deployment and version upgrade will be taken care of automatically

PR #237, opened 2022-06-01
feat: add text-to-text app and update indexing in bff

### LINKED ISSUE #235: feat: add text-to-text app

joschkabraun:
Closes #235 and #244.

- [x] Add Sentence-BERT for text-to-text search to improve search results over CLIP text encoder.
- [x] Update BFF indexing to allow for indexing with `tags` attribute to retrieve those later.

Minor updates:
- [x] Refactor 'app' folder to move logic for fine-tuning to the apps.
- [x] Update test run time


now/apps/base/app.py
```diff
@@ -53,12 +53,12 @@ def flow_yaml(self) -> str:
         """
         return self._flow_yaml
 
-    @flow_yaml.setter
-    def flow_yaml(self, value: str):
+    def set_flow_yaml(self, finetuning: bool):
```
slettner, line 56: Is there a reason for renaming and removing and implementation? I am used to the previous naming convention for setter.

slettner: Just read further, I assume it is because of the finetuning arg. 
Could this also be passed to the app constructor itself? Probably it is not available at that time?

now/apps/image_to_image/app.py
```diff
@@ -22,10 +34,26 @@ def input_modality(self) -> Modalities:
     def output_modality(self) -> Modalities:
         return Modalities.IMAGE
 
+    def set_flow_yaml(self, finetuning: bool = False):
```
slettner, line 37: Maybe we can consider the finetuning arg as a separate property of the app and set it before the flow yaml is set. This way we could keep the old signature for `set_flow_yaml`.

now/apps/image_to_image/app.py
```diff
@@ -1,15 +1,27 @@
...
+)
 from now.dataclasses import UserInput
 from now.run_backend import finetune_flow_setup
 
 
 class ImageToImage(JinaNOWApp):
+    def __init__(self):
+        now_package_dir = os.path.abspath(os.path.join(__file__, '..', '..', '..'))
+        flow_dir = os.path.join(now_package_dir, 'deployment', 'flow')
+        self._flow_yaml = os.path.join(flow_dir, 'flow-clip.yml')
```
slettner, line 23: Is there a reason why this is already set here? Look very similar to the code in `set_flow_yaml`

now/finetuning/settings.py
```diff
@@ -90,11 +91,16 @@ def _is_bi_modal(user_input: UserInput, dataset: DocumentArray) -> bool:
...
-    user_input: UserInput, dataset: DocumentArray, finetune_datasets: Tuple
+    app_instance: JinaNOWApp,
+    user_input: UserInput,
+    dataset: DocumentArray,
+    finetune_datasets: Tuple,
 ) -> FinetuneSettings:
     """This function parses the user input configuration into the finetune settings"""
     return FinetuneSettings(
-        pre_trained_embedding_size=_get_pre_trained_embedding_size(user_input),
+        pre_trained_embedding_size=app_instance.pre_trained_embedding_size[
```
slettner, line 101: Good change to move this to the app!

now/apps/image_to_image/app.py
slettner: On the other side we have to make sure that the two setters are called in the right order because `flow_yaml` depends on the finetuning arg .. not sure here

slettner: Maybe we have to refactor the construction process a bit at some point. I could imagine that a builder pattern or something similar would make different construction steps and their dependencies more explicit but that is a separate PR. 

I guess it is best to keep it as you proposed it @joschkabraun :)

now/apps/base/app.py
joschkabraun: We currently have `JinaNOWApp` and `UserInput`. The former one is static and provides the options associated with each app. The latter one is dynamic and changes with each user input, e.g. fine-tuning.
Maybe there is a better way to let those two objects interact.

joschkabraun: I think a better solution is to rename this to `get_flow_yaml` and call that at the appropriate times. Thanks for the input!

now/apps/image_to_image/app.py
joschkabraun: I agree, I am not happy with how it currently is handled. Let's refactor this later on!

now/apps/base/app.py
slettner: understand, I think I have to look a bit more into `JinaNowApp` to understand what it represents.

makram93 (commented):
Overall looks good to me! Some minor comments :)

now/apps/music_to_music/app.py
```diff
@@ -23,6 +27,18 @@ def input_modality(self) -> Modalities:
...
+    def set_flow_yaml(self, finetuning: bool):
+        now_package_dir = os.path.abspath(os.path.join(__file__, '..', '..', '..'))
+        flow_dir = os.path.join(now_package_dir, 'deployment', 'flow')
+        self._flow_yaml = os.path.join(flow_dir, 'ft-flow-music.yml')
+
+    @property
+    def pre_trained_embedding_size(self) -> Dict[Qualities, int]:
+        return {
+            Qualities.MEDIUM: 512,
+        }
```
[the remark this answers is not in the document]
joschkabraun: Yes, it is set to `Qualities.MEDIUM` by default: https://github.com/jina-ai/now/blob/1e39d77347da3fcab39558e9730dba0db9e473ee/now/finetuning/settings.py#L39

now/apps/image_to_image/app.py
joschkabraun: Good point!

now/apps/text_to_text/app.py
```diff
@@ -0,0 +1,61 @@
...
+from now.constants import CLIP_USES, IMAGE_MODEL_QUALITY_MAP, Modalities, Qualities
+from now.dataclasses import UserInput
+from now.run_backend import finetune_flow_setup
+
+
+class TextToText(JinaNOWApp):
+    def __init__(self):
+        now_package_dir = os.path.abspath(os.path.join(__file__, '..', '..', '..'))
+        flow_dir = os.path.join(now_package_dir, 'deployment', 'flow')
+        self._flow_yaml = os.path.join(flow_dir, 'flow-clip.yml')
```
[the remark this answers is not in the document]
joschkabraun: Yes, thanks!

now/apps/base/app.py
```diff
@@ -53,12 +52,13 @@ def flow_yaml(self) -> str:
...
         """
         Set the flow yaml. Either the path to the yaml or the yaml content.
+        :param finetuning: whether flow for finetuning should be used
         """
-        self._flow_yaml = value
+        curdir = os.path.realpath(__file__)
+        self._flow_yaml = os.path.join(curdir, 'flow.yml')
```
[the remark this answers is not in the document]
joschkabraun: I think pointing to the other flow files makes a lot of sense. We can let it point to the same directory once we migrate the apps out of the repository.

now/apps/text_to_text/flow-sbert.yml
```diff
@@ -0,0 +1,20 @@
+jtype: Flow
+with:
+  name: nowapi
+  port_expose: 8080
+  cors: True
+executors:
+  - name: encoder_text
+    uses: ${{ ENV.ENCODER_NAME }}
+    resources:
+      memory: 8G
```
makram93, line 10: This should be updated to include `resources` under `jcloud`. For more reference you can read this [thread](https://jinaai.slack.com/archives/C01E03LN3K7/p1654676429965409)
Also, please update other flows accordingly. I'm sorry I didn't do it in my last PR

makram93 (changes requested):
LGTM! 🚀 
Some minor comments

PR #240, opened 2022-06-03
chore: update docarray version to 0.13.17

slettner:
This PR closes #223 

From this migration onwards, the dataset carry the docarray version in the name
Screenshot 2022-06-06 at 16 15 36

All datasets have been migrated to version 0.13.17 an stored in the Storage [Bucket](https://console.cloud.google.com/storage/browser/jina-fashion-data/data/one-line/datasets?project=jina-simpsons-florian&pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))&prefix=&forceOnObjectsSortingFiltering=false)

The migration is performed automatically using this [Repo](https://github.com/jina-ai/now-ds-migrator)


makram93:
We also have samples `.bin` files that are loaded when the `playground.py` executes - for the readily available dataset. Should we migrate them too, to keep the `docarray` usage uniform? (I am not sure, if the samples work with the new docarray version, if so then we don't need to migrate them)
image

slettner:
Good that you mentioned it actually. I think it is best if we keep the docarray version in frontend and backend consistent with JinaNOW and the data we are using.

makram93:
> Oh, I overlooked those. I don't think any docarray is properly working without the migration. @makram93 can you point me to a list or location in the code where I can find the missing datasets?

@slettner Here is the [list of datasets](https://github.com/jina-ai/now/blob/main/deployment/playground/playground.py#L23-L36) that have corresponding sample files available on the google bucket. All the sample files have the `*10.bin` suffix which implies that they only have 10 data points, so it is easy to distinguish the sample files.

This is [where the URL](https://github.com/jina-ai/now/blob/main/deployment/playground/playground.py#L88-L90) for the sample files to be downloaded are formed.

joschkabraun:
Nice PR! Also like the repo you created for it.

joschkabraun (approved):

florian-hoenicke:
LGTM

PR #241, opened 2022-06-07
feat: add apps as subcommand

makram93:
Add apps as a subcommand and make arguments supported for each apps unique to them... 

- Users can now enter `jina-now start -h` to get the below help screen, where `image_to_image`, `image_to_text` and `text_to_image` are the subcommand corresponding to each app.  image

- When the user enters `jina-now start image_to_image -h`, he/she gets the help (or additional arguments) for that respective app only
- There's also an argument `--app` whose value/choices are the same as `subcommands` and is picked from `Apps` Enum. If used, the dialog to select the app will not show. Also, when using the `subcommand` the `--app` argument's value is overridden. For e.g: `jina-now start --app image_to_text image_to_image` will have the `app=image_to_image` when argument parsing is done.


joschkabraun:
Great PR! I would refactor it such that all the configurations of the apps are handled within the app and not somewhere else.

now/cli/parser.py
```diff
@@ -53,35 +54,43 @@
...
-    """Add the arguments for the jina now to the parser
-    :param parser: an optional existing parser to build upon
-    :return: the parser
-    """
+def set_img2txt_parser(parser):
+    parser.add_argument(
+        '--quality',
+        help='Choose the quality of the model that you would like to finetune',
+        type=str,
+    )
```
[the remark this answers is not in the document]
makram93: Thanks for the suggestion. I like it too.. Keeping all maintenance in one place.

now/apps/base/app.py
```diff
@@ -109,6 +113,25 @@ def example_datasource(self) -> List[Datasource]:
...
 
+    def set_app_parser(self, parser, formatter) -> None:
+        """
+        This parser reads from the `options` property and parses it
+        to form the command line arguments for app
+        """
+        if self.app != '':
+            parser = parser.add_parser(
+                self.app,
+                help=self.description,
```
florian-hoenicke, line 124: nice. I like this way of getting the attributes. Minimizes the effort of implementing the derived class

now/cli/parser.py
```diff
@@ -17,10 +21,10 @@ def set_base_parser():
         'Help': ('💬', 'https://slack.jina.ai'),
         'Hiring!': ('🙌', 'https://career.jina.ai'),
     }
-    url_str = '\n'.join(f'- {v[0]:<10} {k:10.10}\t{v[1]}' for k, v in urls.items())
+    url_str = '\n'.join(f'- {v[0]:<15} {k:10.10}\t{v[1]}' for k, v in urls.items())
```
florian-hoenicke, line 24: What is this logic doing?

makram93: Ah this is from the core repo... It just prints the extra message at the end
image

now/finetuning/embeddings.py
```diff
@@ -47,6 +47,6 @@ def get_encoder_config(encoder_uses: str, artifact: str) -> _ExecutorConfig:
     """
     return _ExecutorConfig(
         name='encoder',
-        uses=f'jinahub+docker://{encoder_uses}',
+        uses=f'jinahub+sandbox://{encoder_uses}',
```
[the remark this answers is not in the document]
makram93: Yes, sorry. I was working on it and by mistake, it got added to this PR. Thanks! :)

now/apps/base/app.py
```diff
@@ -22,6 +22,10 @@ def __init__(self):
         curdir = os.path.realpath(__file__)
         self._flow_yaml = os.path.join(curdir, 'flow.yml')
 
+    @property
+    def app(self) -> str:
+        return ''
```
florian-hoenicke, line 27: raise NotImplemented error. This is a mandatory attribute without default value

now/cli/parser.py
```diff
@@ -99,22 +87,35 @@ def set_start_parser(sp=None):
...
+
+    sub_parser = parser.add_subparsers(
+        dest='app',
+        description='use `%(prog)-8s [sub-command] --help` '
+        'to get additional arguments to be used with each sub-command',
+        required=False,
     )
 
+    # Set parser args for the Image to Image app
+    ImageToImage().set_app_parser(sub_parser, formatter=_chf)
```
florian-hoenicke, line 106: If you add a new app, you would have to make a code change here. For instance, the MusicToMusic app would have to be enabled as you are stating in the comment.
A better way of doing it is to automatically add everything which is in the Apps list. in the constants.py

florian-hoenicke (changes requested):

now/cli/parser.py
makram93: Thanks for the suggestion! 👍

now/cli/parser.py
```diff
@@ -82,24 +72,46 @@ def set_start_parser(sp=None):
...
+
+    sub_parser = parser.add_subparsers(
+        dest='app',
+        description='use `%(prog)-8s [sub-command] --help` '
+        'to get additional arguments to be used with each sub-command',
+        required=False,
+    )
+
+    # Set parser args for the enabled apps
+    for app in Apps():
```
florian-hoenicke, line 112: nice - I like

florian-hoenicke:
Well done - that is a lean solution

Issue #242, opened 2022-06-07
feat: update bff music


slettner:
Similar to text and image, we need the `/search` and `/index` endpoints for the music modality. 

Acceptance Criteria: 
- [x] implemented `/index` endpoint and data models
- [x] implemented `/search` endpoint and data models
- [x] add unit tests for both endpoints

PR #243, opened 2022-06-07
feat: update bff music

slettner:
This PR adds two endpoints to the BFF for the music modality as described in [here](https://github.com/jina-ai/now/issues/242) and unti tests.

Further, it introduces dependency injection for the Jina client object to reduce code duplication.
```
def get_jina_client(host: str = 'localhost', port: int = 31080) -> Client:
    ...

@router.post(
    ...
)
def search(data: NowMusicSearchRequestModel, jina_client=Depends(get_jina_client)):
    ....
    docs = jina_client.post('/search', query_doc, parameters={"limit": data.limit})
```
The function `get_jina_client` can now be used by all endpoints and the args will be parsed from the request body.

More [information](https://fastapi.tiangolo.com/tutorial/dependencies/)


florian-hoenicke:
LGTM

PR #257, opened 2022-06-13
fix: now stop delete namespace

slettner:
Currently, it was not possible for me to only delete the namespace when running `jina-now stop`.

### Current Code
```
    if len(choices) == 0:
        ...
    else:
        questions = [
            {
                'type': 'list',
                'name': 'cluster',
                'message': 'Which cluster do you want to delete?',
                'choices': choices,
            }
        ]
        cluster = maybe_prompt_user(questions, 'cluster', **kwargs)
    if cluster == 'kind-jina-now':
        ...
    elif 'wolf.jina.ai' in cluster:
        ...
    else:
        with yaspin_extended(
            sigmap=sigmap, text=f"Remove jina NOW from {cluster}", color="green"
        ) as spinner:
            cmd(f'{kwargs["kubectl_path"]} delete ns nowapi')
            spinner.ok('💀')
        cowsay.cow(f'nowapi namespace removed from {cluster}')
```
The namespace deletion branch is in the last `else` which can only reached when the users selects a choice other than 
a wolf cluster or the standard `kind-jina-now` cluster. But the options don't offer any other choice, therefore the last `else` cannot be reached.

In this PR remove the last `else` and add a question in the `kind-jina-now` case that asks whether the user only wants to delete the namespace or the entire cluster.

### DIFF
--- a/now/run_all_k8s.py
+++ b/now/run_all_k8s.py
@@ -40,22 +40,22 @@ def stop_now(contexts, active_context, **kwargs):
         ]
         cluster = maybe_prompt_user(questions, 'cluster', **kwargs)
     if cluster == 'kind-jina-now':
-        only_delete_namespace = maybe_prompt_user(
+        delete_cluster = maybe_prompt_user(
             [
                 {
                     'type': 'list',
-                    'name': 'delete-ns',
-                    'message': 'Do you want to delete the namespace only and keep the cluster?',
+                    'name': 'delete-cluster',
+                    'message': 'Do you want to delete the entire cluster or just the namespace?',
                     'choices': [
-                        {'name': '⛔ no', 'value': False},
-                        {'name': '✅ yes', 'value': True},
+                        {'name': '⛔ no, keep the cluster', 'value': False},
+                        {'name': '✅ yes, delete everything', 'value': True},
                     ],
                 }
             ],
-            attribute='delete-ns',
+            attribute='delete-cluster',
             **kwargs,
         )
-        if not only_delete_namespace:
+        if delete_cluster:
             with yaspin_extended(
                 sigmap=sigmap, text=f"Remove local cluster {cluster}", color="green"
             ) as spinner:
--- a/now/run_all_k8s.py
+++ b/now/run_all_k8s.py
@@ -26,7 +26,6 @@ def stop_now(contexts, active_context, **kwargs):
     if os.path.exists(user(JC_SECRET)):
         flow_details = get_remote_flow_details()
         choices += [flow_details['gateway']]
-        flow_id = flow_details['flow_id']
     if len(choices) == 0:
         cowsay.cow('nothing to stop')
         return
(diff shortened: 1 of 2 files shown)


makram93:
> Currently, it was not possible for me to only delete the namespace when running `jina-now stop`.

I agree with this - If the user selects `create new`, we create the new cluster with a standard name but upon stopping we delete the entire cluster. For this, I agree with your changes to allow either namespace deletion or the entire cluster deletion. However, if a standard(`kind-jina-now`) already exists and yet the user selects `create new cluster` then we should delete the namespace from `kind-jina-now` and proceed with the deployment. This should be addressed in this PR as well.

> The namespace deletion branch is in the last `else` which can only reached when the users select a choice other than a wolf cluster or the standard `kind-jina-now` cluster. But the options don't offer any other choice, therefore the last `else` cannot be reached.

I disagree with this. When we list the cluster name - the user can select an existing cluster and we then create a separate namespace for our deployment without affecting other pods in the same cluster. So when the user selects this cluster to stop we only delete the namespace we created without affecting the whole cluster. So, the last `else` covers the case when the cluster is neither the standard nor the cloud, it deletes only the namespace we created. Last `else` should not offer any other options other than deleting the namespace from the selected existing cluster

> In this PR remove the last `else` and add a question in the `kind-jina-now` case that asks whether the user only wants to delete the namespace or the entire cluster.

This is good but as I said above we have to also consider that if this cluster already exists and the user still wants to create new then we just delete the namespace and redeploy.

makram93:
So, the current flow is that we retrieve all the clusters from the `~/.kube/config` file. User can select any of the existing clusters. The scenario could be as follows:
- The user selects a cluster context that we did not create, i.e. it is not a `kind-jina-now` cluster. In such case we create our namespace `nowapi` and continue with the deployment. In case the `nowapi` namespace already exists in the selected cluster context then we call `ask_existing` which will delete the namespace and recreate else terminate as we assume the flow is already deployed if there's an existing namespace
- The user selects a cluster that we created, i.e., `kind-jina-now`. Then we again call the `ask_existing` function and repeat the same. The assumption is that if there's a `nowapi` namespace existing in the selected cluster context then a flow is already deployed with indexed data, hence the user is asked if he wants to delete and redeploy.
- The user selects `create new`. Then we run `kind get clusters` [here](https://github.com/jina-ai/now/blob/main/now/cloud_manager.py#L23) to see if there's already a `kind-jina-now` cluster, if it exists then we ask if the user would like to recreate it. In which case complete cluster is deleted and recreated

So to answer your question
> I did not look at the use case where a use creates another deployment in the same cluster which is a good use case.

It is up to the user which cluster he wants to choose or rather create a new one. We always maintain a namespace. This way we can delete `nowapi` namespace from those clusters which we did not create safely. In the case of a new `kind-jina-now` cluster, we have the option to either just delete the namespace thereby saving the time of creating a whole new cluster next time, or delete the whole cluster.
image

makram93 (approved):
LGTM!

joschkabraun (approved):
Nice, PR! LGTM :)

PR #259, opened 2022-06-13
Feat share button

makram93:
Add social share button


deployment/playground/playground.py
```diff
@@ -90,6 +91,7 @@ def deploy_streamlit():
             # for now deactivated sample images for text
             output_modality_dir = 'text'
             data_dir = root_data_dir + output_modality_dir + '/'
+            print(data_dir + DATA + f'.txt10-{docarray_version}.bin')
```
joschkabraun, line 94: don't need the print

deployment/playground/playground.py
```diff
@@ -58,15 +59,15 @@ def deploy_streamlit():
...
-    HOST = query_parameters.get('host')[0]
-    PORT = (
-        query_parameters.get('port')[0] if 'port' in query_parameters.keys() else None
-    )
-    OUTPUT_MODALITY = query_parameters.get('output_modality')[0]
-    INPUT_MODALITY = query_parameters.get('input_modality')[0]
-    DATA = (
-        query_parameters.get('data')[0] if 'data' in query_parameters.keys() else None
-    )
+    HOST = "grpcs://nowapi-1cf62a2f29.wolf.jina.ai"  # query_parameters.get('host')[0]
```
joschkabraun, line 62: You forgot to comment out the values for testing

now/deployment/deployment.py
```diff
@@ -6,7 +6,7 @@
 
 
 def deploy_wolf(path: str, name: str, env_file: str = None):
-    return CloudFlow(path=path, name=name, env_file=env_file).__enter__()
+    return
```
joschkabraun, line 9: What is the purpose of this? If that is intended, we should remove this function.

joschkabraun (changes requested):
Nice, adding of sharing button!

deployment/playground/playground.py
makram93: Was testing it locally, forgot to revert this. Thanks for pointing it out

Issue #232, opened 2022-06-01
feat: pixel tracking on playground


florian-hoenicke:
- Include google analytics in hosted playground

fogx:
if we do this we need to also add a privacy widget :(
And if we are already adding a privacy widget, we should also consider adding [hotjar](https://www.hotjar.com/). Its really nice.

fogx:
add utm code to link on readme

fogx:
For the pixel tracking we need to add a js snippet to the streamlit app. Here are two threads regarding the topic.
We'll add a google tag manager snippet, because we can use it to modify events and add integrations without having to modify the source code. Then we can add a hotjar integration on top.

https://medium.com/@stevernewman/how-to-add-google-analytics-to-a-streamlit-application-face3c9f8f97
https://discuss.streamlit.io/t/how-to-add-google-analytics-or-js-code-in-a-streamlit-app/1610/35

For the cookie consent, we will use a cookie consent tool - probably cookiebot. However, they require us to define a domain name, which we don't have right now. 
If you have any solutions, please implement. Otherwise i would suggest we skip the consent banner for the next week and just add tracking without it. Then i can fix it when i'm back.

GTM snippet:
https://developers.google.com/tag-manager/quickstart

 Paste this code as high in the <head> of the page as possible: 
```
<!-- Google Tag Manager -->
<script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
})(window,document,'script','dataLayer','GTM-MKCJBDS');</script>
<!-- End Google Tag Manager -->
```
 Additionally, paste this code immediately after the opening <body> tag: 
```
<!-- Google Tag Manager (noscript) -->
<noscript><iframe src="https://www.googletagmanager.com/ns.html?id=GTM-MKCJBDS"
height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
<!-- End Google Tag Manager (noscript) -->
```

fogx:
@fogx 
- [ ] Change GTM account to use jina.ai general account and update the snippet.
- [ ] test if traffic is detected properly

fogx:
[...]

PR #268, opened 2022-06-21
feat: add tracking

joschkabraun:
In order to add tracking (#232) we need to modify the `static/index.html` file in the `streamlit` package. Note, the current solution simply overwrites this file. This means if we update`streamlit` in the future, we need to update our modifications of the file.

Solution inspired from: https://discuss.streamlit.io/t/adding-a-meta-description-to-your-streamlit-app/17847/5


deployment/playground/Dockerfile
```diff
@@ -7,10 +7,11 @@ COPY requirements.txt /requirements.txt
 RUN pip3 install -r requirements.txt
 COPY . /
 
+# needed to copy custom streamlit index.html file to add tracking
+COPY custom_streamlit_index.html /usr/local/lib/python3.9/site-packages/streamlit/static/index.html
```
makram93, line 11: `custom_streamlit_index.html` has to be rebuilt every time there's a change in our `playground.py`? If so, can you also specify how to build this file and maybe add a note in the `playgroun.py`?

makram93: Oh okay, so it is dependent on the streamlit package, if we update the version of streamlit package then we have to update our `custom_streamlit_index.html` accordingly. 

Nice solution! :)

deployment/playground/custom_streamlit_index.html
```diff
@@ -0,0 +1,230 @@
+<!doctype html>
```
makram93, line 1: Can you provide the source or reference for this file? It would be easier for any other developer to quickly modify it incase we bump the streamlit package version

florian-hoenicke:
@fogx Could you please update the tracking snippet and check if it is working?

fogx (approved):

now/constants.py
```diff
@@ -1,13 +1,12 @@
...
 
-DOCKER_BFF_PLAYGROUND_TAG = '0.0.70-feat-tracking'
+DOCKER_BFF_PLAYGROUND_TAG = '0.0.73-feat-tracking'
 
 
 class Modalities(BetterEnum):
     TEXT = 'text'
     IMAGE = 'image'
     MUSIC = 'music'
-    VIDEO = 'video'
```
makram93, line 10: I think this has to be there? @florian-hoenicke

fogx: You're right. We removed this and ll 91-94 in now/dialog.py for debugging because of the following error:
`AttributeError: type object 'Apps' has no attribute 'TEXT_TO_VIDEO'`
I will add it back, but the error still occurs right now. Please test and correct @florian-hoenicke

florian-hoenicke: But how it comes that the tests are running through? Did you merge main?

makram93: I think it throws an error when using the dialog because the `text_to_video` is not there in `Apps`... but if we pass args it is fine hence the tests are successful. I'll quickly push the fix

makram93: Once the CI is green, it is good to merge

PR #273, opened 2022-06-22
feat(datasource): add demo datasets to apps

slettner:


slettner:
This PR only makes apps list their available demo datasets.
This is needed to list them in the storefront. 

The `get_data` logic is not used yet. 
All the data loading logic lives inside `data_loading` module. Instead, apps should know all about the data they support and use the `Datasources` to implement loading logic. `data_loading` should be replaced by `data_sources` in the future. 

For example, if a user specifies a custom docarray via the storefront, the now library will construct the data by creating a `CustomDocarray(Datasource)` and call its `get_data` function.

slettner:
In think datasource can be grouped into 
- DemoDataset
- CustomDocarray
- LocalFiles
- Database

now/datasource/datasource.py
```diff
@@ -1,46 +1,61 @@
...
-        #  This signature misses the mapping arg from the baseclass.
-        #  The baseclass interface is not really guiding the user a lot at the moment.
-        #  Maybe we should start with an explicit ObjectStorageDataset or something like that
-        #  because all our app are using that atm. and abstract the interface furhter once we work on
-        #  the database table column mapping stuff?
-
-        url = f'{BASE_STORAGE_URL}/{self.modality}/{self.ds_id}{("." + quality) if quality is not None else ""}.bin'
-        # TODO  return document array from url
-        raise NotImplementedError()
+        self.modalities_folder = modality_folder
```
makram93, line 43: What is this needed for?

joschkabraun (approved):
LGTM!

PR #275, opened 2022-06-23
test: balance

florian-hoenicke:


makram93:
Estimated duration for group 3 is 999 s but actually it takes only 8s. 
image
image

One reason could be that `.test_durations` was created by running the test cases on local machine which may not be as fast as the ubuntu runner on github. However, I calculated that even if we count the durations specified in the file for the test cases being used in group 3 sums up to ~85s which is much much lower than the estimated time that it shows in the above ss. 
Maybe we should re-run the test cases and update the `test_durations` file

florian-hoenicke:
Good point. It is because durations for the other tests are missing. Then, it takes the average. I will do the rerun for all durations

PR #276, opened 2022-06-23
fix: update version

makram93:
Bump jina version in the `requirements.txt` file and also fix it in the flow.yml files

### DIFF
--- a/now/constants.py
+++ b/now/constants.py
@@ -1,6 +1,6 @@
 from now.utils import BetterEnum
 
-DOCKER_BFF_PLAYGROUND_TAG = '0.0.70-1'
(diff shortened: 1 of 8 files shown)


now/constants.py
```diff
@@ -1,6 +1,6 @@
 from now.utils import BetterEnum
 
-DOCKER_BFF_PLAYGROUND_TAG = '0.0.69-2'
+DOCKER_BFF_PLAYGROUND_TAG = '0.0.70-1'
```
florian-hoenicke, line 3: why do we need to update this?

florian-hoenicke: ah okay got it - nevermind :D

PR #278, opened 2022-06-23
feat: integrate finetuner client

slettner:


slettner:
closes https://github.com/jina-ai/now/issues/213

now/finetuning/run_finetuning.py
```diff
@@ -72,74 +65,97 @@ def finetune_now(
...
-        finetune_settings.pre_trained_embedding_size
-        if not finetune_settings.bi_modal
-        else finetune_settings.pre_trained_embedding_size * 2
-    )
-    head = LinearHead(input_size, finetune_settings.finetune_layer_size)
-
-    finetuner.fit(
-        head,
+    experiment_name = 'now-finetuning-' + _get_random_string(8)
+    run_name = _get_random_string(12)
```
bwanglzu, line 103: minor thing, if you don't specify a `run_name`, finetuner will automatically create a random string for you, see: https://finetuner.jina.ai/walkthrough/basic-concepts/ 

> Please note that these two arguments are optional. If not supplied, Finetuner will use the current working directory as a default experiment_name, and generate a random run_name for you.

now/finetuning/run_finetuning.py
```diff
@@ -72,74 +65,97 @@ def finetune_now(
...
+            'hidden_sizes': finetune_settings.hidden_sizes,
+            'l2': True,
+            'bias': False if finetune_settings.bi_modal else True,
+        },
         train_data=finetune_ds.train,
+        experiment_name=experiment_name,
+        run_name=run_name,
         eval_data=finetune_ds.val,
-        epochs=finetune_settings.epochs,
-        learning_rate=finetune_settings.learning_rate,
```
bwanglzu, line 129: i suggest you to keep a learning rate, otherwise it will use the default learning rate of the optimizer, which often deliver bad results.

now/finetuning/run_finetuning.py
```diff
@@ -72,74 +65,97 @@ def finetune_now(
...
+            run_failed = run.status()['status'] == 'FAILED'
+            sleep(1)
+        if run_failed:
+            spinner.fail('👎')
+            print(run.logs())
+            sys.exit(0)
+        else:
+            spinner.ok('👍')
+
+    with yaspin(sigmap=sigmap, text='Running finetune job', color='green') as spinner:
```
bwanglzu, line 136: incase you wanted, i believe `rich` is more stable in these things, i'm note sure about yaspin, maybe it's also very good.
64C99EFD-3741-4E31-8A8B-B9D56E1469A3

now/finetuning/run_finetuning.py
joschkabraun: Thank you Bo!

florian-hoenicke:
Looks good. 
There is just the small issue where the matches can not be displayed.
The following approaches can solve this.
a) modify the playground to only care about uri in case the other attributes are not set
b) modify the playground to ignore uri in case it does not start with http
c) remove uri before indexing the dataset

I would recommend c) because it does not require to modify the playground and also there is less data indexed.

now/finetuning/run_finetuning.py
```diff
@@ -71,83 +60,97 @@ def finetune_now(
...
         EvaluationCallback(
             finetune_ds.val_query,
             finetune_ds.val_index,
             limit=finetune_settings.eval_match_limit,
-            num_workers=8,
-            metrics={'ndcg': (ndcg_at_k, {})},
+            # metrics=['ndcg'],
         ),
-        BestModelCheckpoint(monitor='ndcg', save_dir=save_dir),
+        # BestModelCheckpoint(monitor='ndcg', save_dir=save_dir, verbose=True),
```
makram93, line 87: We don't need this?

now/finetuning/run_finetuning.py
```diff
@@ -71,83 +60,97 @@ def finetune_now(
...
+            'hidden_sizes': finetune_settings.hidden_sizes,
+            'l2': True,
+            'bias': False if finetune_settings.bi_modal else True,
+        },
         train_data=finetune_ds.train,
+        experiment_name=experiment_name,
         eval_data=finetune_ds.val,
-        epochs=finetune_settings.epochs,
-        learning_rate=finetune_settings.learning_rate,
-        batch_size=finetune_settings.batch_size,
```
makram93, line 130: Are we relying on default values?

makram93 (commented):
Looks good to me. One observation I would like to put forth: Let's say somebody wants to use the fine-tuning option and finetuning is successful, however, at a later stage, some error occurs because of which the app deployment fails. Now, if the user re-runs the same app should he run the finetuning step again? Or should we allow him to re-use the artifact and skip finetuning step the second time?

now/finetuning/run_finetuning.py
joschkabraun: No, the best model is saved automatically and we don't need the others.

now/apps/music_to_music/ft-flow-music.yml
```diff
@@ -15,11 +15,24 @@ executors:
...
+    env:
+      JINA_LOG_LEVEL: DEBUG
   - name: linear_head
-    uses: ${{ ENV.LINEAR_HEAD_NAME }}
+    uses: jinahub+docker://FinetunerExecutor/v0.9.2
     uses_with:
-      finetune_layer_size: ${{ ENV.FINETUNE_LAYER_SIZE }}
-      pre_trained_embedding_size: ${{ ENV.PRE_TRAINED_EMBEDDINGS_SIZE }}
+      artifact: ${{ ENV.FINETUNE_ARTIFACT }}
+      token: ${{ ENV.JINA_TOKEN }}
```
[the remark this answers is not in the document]
joschkabraun: The user running the app. We need it to retrieve the artefact from Hubble.

joschkabraun:
> I would recommend c) because it does not require to modify the playground and also there is less data indexed.

Okay, I opted for c) as well with a to do to move it into the preprocessing executor.

now/finetuning/run_finetuning.py
makram93: If the epoch is less, like around 10-20, then I would say default is fine, if not then let's assign a default value that is less.

joschkabraun:
> Looks good to me. One observation I would like to put forth: Let's say somebody wants to use the fine-tuning option and finetuning is successful, however, at a later stage, some error occurs because of which the app deployment fails. Now, if the user re-runs the same app should he run the finetuning step again? Or should we allow him to re-use the artifact and skip finetuning step the second time?

That sounds like a great idea! I created a follow-up ticket for it #315.

joschkabraun:
> > I would recommend c) because it does not require to modify the playground and also there is less data indexed.
> 
> Thank you for the investigation! I opted for c) as well with a to do to move it into the preprocessing executor.

I have investigated the issue further. And it is a trickle down effect that we need to set the `embedding` attribute for the `FinetunerExecutor` in the `CastNMoveNowExecutor` which resets the `blob` attribute. I have updated the `CastNMoveNowExecutor` to convert `blob` into `uri` if it is present.

PR #287, opened 2022-06-29
fix: progress bar

florian-hoenicke:
before:
- got quickly stuck at 100 and then waiting
image
after:
- goes step by step
image


makram93:
There's no ETA shown on the progress bar from client?

florian-hoenicke:
> There's no ETA shown on the progress bar from client?

Good observation. I was reporting this to core.

PR #314, opened 2022-07-22
feat: refactor fine-tune logic

### LINKED ISSUE #313: feat: refactor after show & tell

joschkabraun:
Closes #313 by:
- removes the question to delete old WOLF flows before deploying new one (this is a suggestion, given that people can deploy multiple flows through the storefront)
- introduces additional encode flow for all apps to be used when embeddings are inferred before finetuning
- split of `deploy_flow` into an function only concerned with deploying flows (called `deploy_flow` in `flow.py`) and another function concerned with indexing data (`call_index` in `run_backend.py`)
- enables monitoring in flows where it is missing
- adds finetuning for `text-to-video` and `text-to-text`
- removes unused argumens in `run_bff_playground.run`
- moves finetuning logic into `run_finetuning.py` entirely
- renames `finetune_now` to `maybe_finetune` which is now main entry point of finetuning logic
- fixes bug that calculating embeddings for finetuning will always deploy a k8s locally even for WOLF deployment
- creates `now_common/utils.py` which contains
    - shared function to create flow environment variables dictionary for all apps (used to be in `run_backend.py`
    - a refactored `setup` function which is used by all current apps (main logic stems from `finetune_flow_setup` in `run_backend.py`)
      - note: that `music_to_music` app has for the demo datasets a slightly different flow as we need to integrate an external executor


now/finetuning/run_finetuning.py
```diff
@@ -39,18 +41,26 @@ def finetune_now(
...
     :return: artifact of finetuned model and token required for FinetunerExecutor
     """
-    if pre_trained_head_map is not None and user_input.data in pre_trained_head_map:
-        print(f'⚡️ Using cached hub model for speed')
-        return pre_trained_head_map[user_input.data]
+    if not finetune_settings.perform_finetuning:
+        return
+
+    print(f'🔧 Perform finetuning!')
+    app_instance.set_flow_yaml(encode=True)
```
makram93, line 56: If I am not wrong - You are passing `encode=True` to use the encode flow when calling `_maybe_add_embeddings` which will set `self.flow_yaml`. So, once this finetuning part is done, shouldn't we change it back to `finetune=True` to use the finetune flow when indexing?

makram93 (commented):
Looks good to me. Some minor comments

now/apps/image_to_text/app.py
```diff
@@ -45,11 +45,22 @@ def required_docker_memory_in_gb(self) -> int:
         return 8
 
     def set_flow_yaml(self, **kwargs):
+        finetuning = kwargs.get('finetuning', False)
+        encode = kwargs.get('encode', False)
+        if finetuning and encode:
+            raise ValueError(f"Can't set finetuning={finetuning} and encode={encode}")
```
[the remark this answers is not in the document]
joschkabraun: Good point, thank you!

now/finetuning/run_finetuning.py
joschkabraun: Yes, we need to change it back which is done in `setup_clip_music_apps` in `now_common/utils.py`.

makram93: Ah okay, I saw that but then I thought `setup_clip_music_apps` is used only for music2music app. My bad!

PR #318, opened 2022-07-28
Feat: Load from S3 Bucket

joschkabraun:
Implements loading of data from folder on S3 by copying folder structure and then creating docarray from that. This is an extension of: #317.


tests/unit/data_loading/test_data_loading.py
```diff
@@ -55,20 +58,20 @@ def test_da_pull(da: DocumentArray):
...
-def test_da_url_fetch(da: DocumentArray):
-    user_input = UserInput()
-    user_input.is_custom_dataset = True
-    user_input.custom_dataset_type = DatasetTypes.URL
-    user_input.dataset_url = 'https://some.url'
-
-    loaded_da = load_data('', user_input)
-
-    assert is_da_text_equal(da, loaded_da)
+# def test_da_url_fetch(da: DocumentArray):
```
makram93, line 66: If we don't need this function then lets remove it

makram93 (approved):
Looks good to me

PR #323, opened 2022-08-02
Feat add preprocess executor

joschkabraun:
Moves preprocessing to additional executor in flow such that those can be scaled up for indexing more data.

Also, fixes error of `text-to-text` search app: `requests.exceptions.HTTPError: 401 Client Error: Unauthorized for url: https://huggingface.co/api/models/sentence-transformers/e`.

Also, fixes that we needed to execute end-to-end tests in certain order. The issue was that in `deploy_flow` when loading the environment variables with `load_dotenv`, the environment variables are by default not overriden. This lead to the usage of the same environment variables for all subsequent tests on the same runner. I.e. using CLIP for text-to-text retrieval.

Introduces usage of medium quality CLIP model as medium quality model for `text-to-text` search. Otherwise, indexing 200k documents takes a long time.


now/data_loading/data_loading.py
```diff
@@ -60,7 +60,13 @@ def _pull_docarray(dataset_name: str):
         exit(1)
 
 
-def _load_from_disk(app: JinaNOWApp, dataset_path: str) -> DocumentArray:
+def _load_to_datauri_and_save_into_tags(d: Document) -> Document:
+    d.tags['uri'] = d.uri
+    return d.convert_uri_to_datauri()
```
makram93, line 65: Why do we need to convert it to datauri?

now/apps/text_to_video/app.py
```diff
@@ -94,34 +87,35 @@ def setup(
...
-                    elif d.tensor is not None:
-                        d.convert_tensor_to_blob()
-                sample_video(d)
-            except:
-                pass
-            return d
+    def preprocess(
+        self, da: DocumentArray, user_input: UserInput, is_indexing=False
+    ) -> DocumentArray:
+        if is_indexing:
```
makram93, line 93: This `preprocess` will be called only for the data which is to be loaded and indexed, then why we need this boolean variable?

now/data_loading/data_loading.py
joschkabraun: Because if we load from fish, we need to load files locally and then send them to the preprocessor. The pre processor doesn’t have access to local files

now/apps/text_to_video/app.py
joschkabraun: It will be called for indexing and searching. With our current setup we need to indicate whether we are indexing (eg preprocessing images) or if we are searching (eg preprocessing text).

makram93: But from the UI, when the image is selected it is directly sent to the Search Flow via BFF, so how or when does this preprocess comes into picture for searching?

joschkabraun: Yes and the BFF sends it directly to the search flow which includes as the first executor the preprocessor (after a possible security executor).

We could do the preprocessing in the BFF but I think there are a couple of reasons against it:

- Latency: sending actual images/gifs makes it a lot slower (previously indexing 10k GIFs took 3-5 hours, now indexing 30k GIFs without any replicas takes 2-2.5 hours)
- we already have a preprocessor in place such that it makes sense to have all the relevant code in the same location
- The BFF shouldn’t be equipped to do any heavy lifting in terms of preprocessing as this requires a different architecture
- We might change preprocessing over time such that the BFF needs to know according to which Jina version preprocessing is applied
- if we decide to write any additional features such as duplicate detection they could need to query large amounts of data such that efficiency of the search endpoint is particularly important

makram93 (approved):
Looks good to me! 🚀

PR #324, opened 2022-08-04
Fix bump streamlit

makram93:


joschkabraun (approved):
LGTM & works on my machine!

PR #329, opened 2022-08-08
Fix wrap error

makram93:
- [x] add exception handling in bff
- [x] handle the error message on UI

---
Before ->

---
Now -> 
image


joschkabraun (approved):
LGTM :)

PR #348, opened 2022-08-16
feat: enable gpu

azayz:


now/constants.py
```diff
@@ -5,7 +5,8 @@
 # DEMO_DATASET_DOCARRAY_VERSION = docarray_version
 DEMO_DATASET_DOCARRAY_VERSION = '0.13.17'
 DOCKER_BFF_PLAYGROUND_TAG = '0.0.105'
-NOW_PREPROCESSOR_VERSION = '0.0.35'
+NOW_PREPROCESSOR_VERSION = '0.0.34-enable-gpu'
+GPU_THRESHOLD = 50000
```
joschkabraun, line 9: as dicussed in pairing would increase this to larger than 200_000 such that demo datasets don't trigger GPU

joschkabraun (commented):
Nice PR! After fixing the minor comments, the other flow.yaml and adding a test it LGTM :)

florian-hoenicke (approved):
LGTM after joschkas feedback is addressed

florian-hoenicke:
I would recommend ignoring the music case for GPU because there is low business value behind currently.

florian-hoenicke:
In our current setup, it is always required to update the preprocessor version with each commit.

PR #357, opened 2022-08-17
Feat annlite indexer

### LINKED ISSUE #345: feat: annlite integration
Following [1 Mio. DocArray Benchmark](https://docarray.jina.ai/advanced/document-store/benchmark/#benchmark-result), using Quadrant would make the most sense given it's excellent query time with reasonable recall. But Qdrant, ElasticSearch and Weviate aren't supported by `jcloud` (we would need to host the actual database somewhere else). Additionally, there have been excellent results on AnnLite for millions of documents. Therefore, choose AnnLite.

### LINKED ISSUE #351: fix: runners on ci cd
Two of our runners don't receive any tests in the ci cd. But the ci cd would be much faster when splitting up the tests properly.

Example bellow:
Image

joschkabraun:
Closes #345 by using the following indexers in different scenarios:

- `DocarrayIndexerV2` for up to 50k documents being indexed or if it is a CI run (otherwise test fails)
-  `AnnLiteIndexer` with `cpu=1.0` and `memory=2G` for jcloud resources for between 50k and 250k documents being indexed (actually, code supports `cpu=0.1` but due to https://github.com/jina-ai/jina/issues/5075 we can't set `cpu` resources via environment variables; such that currently `cpu=1.0` is chosen in order to avoid having to replicate multiple `flow.yaml` files)
-  `AnnLiteIndexer` with `cpu=1.0` and `memory=4G` for jcloud resources for more than 250k documents being indexed

Also, closes #351 by updating pytest-split parameters to `--splitting-algorithm least_duration`. Due to moving text dataset tests to jcloud, overall test time still takes around 25 minutes.


florian-hoenicke (approved):
Just one test on wolf with annlite indexer is needed.
Great work

PR #365, opened 2022-08-18
feat: error handling in bff

florian-hoenicke:
At the moment, we only get a 500 back without more details. After this is merged, we get better output to investigate errors


makram93 (approved):
🚀

PR #367, opened 2022-08-18
Update JCloud and use latest jina auth

makram93:
In this PR we update the requirements and make use of the latest version of `jina-hubble-sdk` which is the central library for auth across Jina ecosystem.


florian-hoenicke (approved):
LGTM, nice that you simplified the _get_info_hubble method. Looks much better now

PR #376, opened 2022-08-19
Fix refined logging

makram93:
Catch any exception from BFF and print it in a clear format as shown in picture below:

image

Expanding this will show the complete stack trace -

image


florian-hoenicke (approved):
LGTM

PR #380, opened 2022-08-20
fix(gif-app): fix set flow yaml

joschkabraun:
Fixes bug in `text-to-video` search which made the preprocessor crash when starting the app.


florian-hoenicke (approved):
LGTM

PR #383, opened 2022-08-22
feat: enable gpu

azayz:


azayz:
one sec, don't merge

now_common/utils.py
```diff
@@ -66,13 +70,30 @@ def setup_clip_music_apps(
         finetune_datasets=app_instance.finetune_datasets,
     )
 
+    gpu = '0'
+    device = 'cpu'
+    gpu_threshold = 250000
+
+    if 'NOW_CI_RUN' in os.environ and 'NOW_CI_GPU' in os.environ:
+        gpu_threshold = 10000
```
florian-hoenicke, line 78: Is the threshold 10K correct?
We either index 1K or 1.5K if I remember correctly

tests/integration/test_end_to_end.py
```diff
@@ -116,6 +118,8 @@ def test_backend_demo_data(
         pytest.skip('Too time consuming, hence skipping!')
 
     os.environ['NOW_CI_RUN'] = 'True'
+    if gpu:
+        os.environ['NOW_CI_GPU'] = 'True'
```
florian-hoenicke, line 122: I think just having 'NOW_CI_RUN' would be enough.
Do you think `NOW_CI_GPU` is really needed?

florian-hoenicke (changes requested):

tests/integration/test_end_to_end.py
azayz: then gpu would be enabled for all tests, we would like to test with and also without gpu

florian-hoenicke: Okay, I see. But can't you not just use the `gpu` parameter for this?

now/constants.py
```diff
@@ -50,7 +50,7 @@ class Qualities(BetterEnum):
     Qualities.EXCELLENT: ('ViT-L14', 'openai/clip-vit-large-patch14'),
 }
 
-CLIP_USES = 'CLIPEncoder/v0.2.1'
+CLIP_USES = 'CLIPEncoder/0.3.1-gpu'
```
makram93, line 53: I believe this should be parameterized since we sometimes do not use the GPU.

now_common/utils.py
```diff
@@ -66,13 +70,30 @@ def setup_clip_music_apps(
         finetune_datasets=app_instance.finetune_datasets,
     )
 
+    gpu = '0'
+    device = 'cpu'
+    gpu_threshold = 250000
+
+    if 'NOW_CI_GPU' in os.environ:
+        gpu_threshold = 1000
```
joschkabraun, line 78: Instead of using `NOW_CI_GPU` I would simply use `NOW_CI_RUN` and then the test will be automatically applied to the `best-artworks` dataset as it is of 2k docs (cf. `data_loading/data_loading:load_data`)

tests/integration/test_end_to_end.py
joschkabraun: Can remove `NOW_CI_GPU` by comment in `utils.py`

joschkabraun (changes requested):
LGTM after minor comment

now/constants.py
azayz: I think it's still fine this way because even when we don't use gpu it stil works

makram93: Yes, it will work but it takes time to install the torch-gpu since it is a very large file. So, if a user is running it locally with slow internet, it will take time before the pods could be ready. I agree it is not a breaking change but just something to consider about the speed.

now_common/utils.py
```diff
@@ -66,13 +70,35 @@ def setup_clip_music_apps(
         finetune_datasets=app_instance.finetune_datasets,
     )
 
+    gpu = '0'
+    device = 'cpu'
+    gpu_threshold = 250000
+    if 'NOW_CI_CPU' in os.environ:  # test gpu in CI if len(dataset) > 1000
```
makram93, line 76: Did you mean `NOW_CI_RUN`?
```suggestion
    if 'NOW_CI_RUN' in os.environ:  # test gpu in CI if len(dataset) > 1000
```

azayz: yep

makram93 (approved):
🚀

joschkabraun (approved):

PR #385, opened 2022-08-22
fix: fix chunk level search annlite

joschkabraun:
Issue was that `AnnLiteIndexer` uses `index_traversal_paths` and `search_traversal_paths` and not `traversal_paths`, while `DocArrayIndexerV2` uses only `traversal_paths`. This also lead to that matched documents need to be accessed through `results[0].chunks[0].matches` instead of `results[0].matches` (as for `DocArrayIndexerV2`


deployment/bff/app/v1/routers/txt2video.py
```diff
@@ -65,4 +65,10 @@ def search(data: NowTextSearchRequestModel):
...
     )
 
-    return docs[0].matches.to_dict()
+    # DocArrayIndexerV2 returns matches on matches level, while AnnLite returns them on .chunks[0].matches level
+    if len(docs[0].matches) > 0:
+        return docs[0].matches.to_dict()
+    elif len(docs[0].chunks[0].matches) > 0:
+        return docs[0].chunks[0].matches.to_dict()
+    else:
+        return DocumentArray().to_dict()
```
azayz, line 74: in which case would you return an empty dict ?

joschkabraun: If e.g. no results have been returned

makram93, line 74:
```suggestion
    if len(docs[0].chunks[0].matches) > 0:
        return docs[0].chunks[0].matches.to_dict()
    else:
        return docs[0].matches.to_dict()
```

makram93 (approved):
LGTM! 🚀 
Just a suggestion but not important. Can be merged without it

deployment/bff/app/v1/routers/txt2video.py
joschkabraun: Good idea! Integrated a (safer) variant of that!

PR #398, opened 2022-08-24
feat: add tags

azayz:


now/data_loading/data_loading.py
```diff
@@ -54,6 +56,25 @@ def load_data(app: JinaNOWApp, user_input: UserInput) -> DocumentArray:
     return da
 
 
+def _load_tags_from_json(da: DocumentArray):
+
+    print('Loading tags!')
+    dic = {}
+    for i, d in enumerate(da):
+        folder = d.uri.split('/')[-2]
```
florian-hoenicke, line 64: it can still be the case that two subfolders are identical even though they are in different parent folders. This would lead to collisions
Instead, you can use `d.uri.rsplit('/', 1)[0]`

now/data_loading/data_loading.py
```diff
@@ -54,6 +56,25 @@ def load_data(app: JinaNOWApp, user_input: UserInput) -> DocumentArray:
...
+        if not d.uri.endswith('.json') and folder not in dic:
+            dic[folder] = i
+
+    for i, d in enumerate(da):
+        folder = d.uri.split('/')[-2]
```
florian-hoenicke, line 69: I think you don't have to determin the folder again

florian-hoenicke (changes requested):

now/data_loading/data_loading.py
azayz: I think I do because I go through the documents twice, one (L63) to put gif paths in the dictionary and second (L68) to determine if a json file's folder path corresponds to a gif (exists in the dict)

now/data_loading/data_loading.py
```diff
@@ -54,9 +56,28 @@
     return da
 
 
+def _load_tags_from_json(da: DocumentArray):
+
+    print('Loading tags!')
+    dic = {}
+    for i, d in enumerate(da):
+        folder = d.uri.rsplit('/', 1)[0]
+        if not d.uri.endswith('.json') and folder not in dic:
```
joschkabraun, line 65: loading the json file will fail as the json is on s3 or another cloud bucket

now/data_loading/data_loading.py
```diff
@@ -54,9 +56,28 @@
...
+            dic[folder] = i
+
```
joschkabraun, line 67: This logic will update the tags of the last document in the same folder as the json.
I would extend this to:
- test that only one file is in the same folder to make sure that we assign the tags to the right document (might even need to be extended to check that only the one file in the folder which matches the output modality of the app)
- if there are multiple files in folder, match json with file which has the same stem

now/data_loading/data_loading.py
```diff
@@ -54,9 +56,28 @@
...
+    for i, d in enumerate(da):
```
joschkabraun, line 68: I would filter returned docarray to not contain .json files anymore as they already have been been processed

joschkabraun (changes requested):

now/data_loading/data_loading.py
azayz: I believe this is on the client's side to ensure that his data follows the specific structure otherwise It would be hard to generalize

florian-hoenicke: > if there are multiple files in folder, match json with file which has the same stem
yes, that would be a good idea. If you want to skip it in this pr to keep it small it is also fine.
> test that only one file is in the same folder to make sure that we assign the tags to the right document
in that case, I would drop the file if it is just an annotation and keep if it is just a content file

florian-hoenicke (changes requested):

now/data_loading/data_loading.py
```diff
@@ -36,6 +37,7 @@ def load_data(app: JinaNOWApp, user_input: UserInput) -> DocumentArray:
             da = _load_from_disk(app, user_input)
         elif user_input.custom_dataset_type == DatasetTypes.S3_BUCKET:
             da = _list_files_from_s3_bucket(app=app, user_input=user_input)
+            da = _load_tags_from_json(da, user_input, True)
```
joschkabraun, line 40: add `if any([doc.uri.endswith('.json') for doc in da])`

now/data_loading/data_loading.py
```diff
@@ -36,6 +37,7 @@ def load_data(app: JinaNOWApp, user_input: UserInput) -> DocumentArray:
             da = _load_from_disk(app, user_input)
```
joschkabraun, line 37: with add `if any([doc.uri.endswith('.json') for doc in da])` we can also activate for local storage

joschkabraun (commented):
After comments, good to go! :rocket:

now/data_loading/data_loading.py
```diff
@@ -54,6 +59,70 @@ def load_data(app: JinaNOWApp, user_input: UserInput) -> DocumentArray:
...
+def _open_json(path: str):
+    with open(path, 'r') as f:
+        data = json.load(f)
+    return data
+
+
+def _open_s3_json(path: str, user_input: UserInput):
+    import boto3
+
+    path_splits = path.split('/')
```
florian-hoenicke, line 71:
```
'a/b/c/d/e.json'.split('/', 1)
Out[3]: ['a', 'b/c/d/e.json']
```

florian-hoenicke (approved):
LGTM

PR #434, opened 2022-08-30
feat: add data classes

jupyterjazz:
Quick PR to add data classes because we need them for other tasks.
Will probably be refactored in the near future. 

Using `pydantic` because `dataclasses` can't load nested classes from json.

### DIFF
--- a/now/now_dataclasses.py
+++ b/now/now_dataclasses.py
@@ -1,3 +1,4 @@
+from pydantic.dataclasses import dataclass as pydantic_dataclass
 from dataclasses import dataclass
 from typing import Dict, List, Optional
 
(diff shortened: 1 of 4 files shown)


now/now_dataclasses.py
```diff
@@ -33,3 +34,33 @@ class UserInput:
...
+    train_dataset_name: str
+    training_data_generation_methods: List[TrainDataGenerationConfig]
+
+
+@pydantic_dataclass
+class Task:
+    name: str
+    extracted_dataset: str
+    encoders: List[EncoderConfig]
+    add_indexer: bool
```
violenil, line 66:
```suggestion
    indexer_scope: Dict[str, str]
```

violenil (approved):
LGTM!

tests/resources/config.json
```diff
@@ -0,0 +1,52 @@
...
+          "target": {
+            "method": "image_normalization",
+            "parameters": {},
+            "scope": ["uris"]
+          }
+        }
+      ]
+    }
+  ],
+  "add_indexer": false
```
violenil, line 51: "indexer_scope": {'text': 'text', 'image': 'uris'}

Issue #430, opened 2022-08-30
test: unset `embedding` attribute in example datasets


joschkabraun:
Monkey patch `load_data` to unset the `embedding` attribute of the example datasets in order to test also `encoding` of executor. Might need to reduce test size.

PR #437, opened 2022-08-31
feat: use cas executor for clip

### LINKED ISSUE #427: fix: reuse CLIP executor
Either use CAS CLIP executor or have it running as an external executor.

Using CAS CLIP demo introduces additional latencies of 0.2-0.3 seconds.

Image

joschkabraun:
Closes #427 by reusing CaS CLIP executor for any remote deployment. Quality option won't be necessary as best model is used for remote. For local deployment smaller model is used in order to make it faster.

Also, CaS CLIP executor still calculates `embedding` no matter if it is set or not. Therefore, #430 is also closed by this PR.


florian-hoenicke:
Would it be possible to quickly evaluate the quality of the new clip encoder?
Does the embedding now have a different size?

joschkabraun:
> > I evaluated best-artworks and `totally-looks-lik datasets. Should I evaluate more?
> 
> I think it is enough. Did you see much differences in the search quality?

No I did not. Or rather I was very happy with the results of larger CLIP.

florian-hoenicke:
Great work! This will save us a lot of resources for hosting.

PR #442, opened 2022-09-01
fix: adapt annlite

azayz:


now/data_loading/data_loading.py
```diff
@@ -24,27 +24,30 @@ def load_data(app: JinaNOWApp, user_input: UserInput) -> DocumentArray:
     :return: The loaded DocumentArray.
     """
     da = None
-
+    tags = []
     if user_input.is_custom_dataset:
         if user_input.custom_dataset_type == DatasetTypes.DOCARRAY:
             print('⬇  Pull DocArray dataset')
             da = _pull_docarray(user_input.dataset_name)
+            tags = _extract_tags_annlite(da)
```
florian-hoenicke, line 32: Instead of getting the tags from each of these cases, you could just get it once independently of the case.
I see that you also get the tags from `_load_tags_from_json` but then the logic is duplicated.

now/data_loading/data_loading.py
```diff
@@ -119,9 +137,14 @@ def _load_tags_from_json(da: DocumentArray, user_input: UserInput):
                 data = _open_json(d.uri)
             for tag, value in data['tags'].items():
                 da[dic[folder]].tags[tag] = value['slug']
+                tags.add((tag, str(tag.__class__.__name__)))
     if len(ids_to_delete) > 0:
         del da[ids_to_delete]
-    return da
+    final_tags = list(tags)
```
florian-hoenicke, line 143: duplicated logic

florian-hoenicke (changes requested):

now/data_loading/data_loading.py
```diff
@@ -66,6 +67,19 @@ def _open_json(path: str):
     return data
 
 
+def _extract_tags_annlite(da: DocumentArray):
+    tags = set()
+    if any([len(doc.tags.keys()) > 0 for doc in da]):
```
florian-hoenicke, line 72: I think this if-condition is not required. It is handled within the for look

now/data_loading/data_loading.py
```diff
@@ -66,6 +67,19 @@ def _open_json(path: str):
     return data
 
 
+def _extract_tags_annlite(da: DocumentArray):
```
florian-hoenicke, line 70: I think the tag extraction should be not part of the data loading. Can we put it one level above?

now/data_loading/data_loading.py
```diff
@@ -66,6 +67,19 @@ def _open_json(path: str):
...
+    tags = set()
+    if any([len(doc.tags.keys()) > 0 for doc in da]):
+        for doc in da:
+            if len(doc.tags.keys()) > 0:
+                for tag, _ in doc.tags.items():
+                    tags.add((tag, str(tag.__class__.__name__)))
+    final_tags = list(tags)
```
florian-hoenicke, line 77: I think this should do the same job:

final_tags = [list(tag) for tag in tags]

florian-hoenicke (changes requested):

now/run_backend.py
```diff
@@ -28,9 +28,10 @@ def run(app_instance: JinaNOWApp, user_input: UserInput, kubectl_path: str):
     :return:
     """
     dataset = load_data(app_instance, user_input)
+    tags = _extract_tags_annlite(dataset)
 
     env_dict = app_instance.setup(
-        dataset=dataset, user_input=user_input, kubectl_path=kubectl_path
+        dataset=dataset, user_input=user_input, kubectl_path=kubectl_path, tags=tags
```
joschkabraun, line 34: I wouldn't change the interface of the `setup` function as all those information can be extracted from `dataset` parameter. I would move this into `now_common.utils.common_setup`

joschkabraun (changes requested):

now_common/utils.py
```diff
@@ -141,17 +144,27 @@ def get_indexer_config(num_indexed_samples: int) -> Dict:
...
     """
-    config = {'indexer_uses': 'AnnLiteNOWIndexer2/0.3.0'}
+    config = {'indexer_uses': 'AnnLiteNOWIndexer3/latest'}
     threshold1 = 50_000
     threshold2 = 250_000
     if 'NOW_CI_RUN' in os.environ:
         threshold1 = 1_500
     if num_indexed_samples <= threshold1:
-        config['indexer_uses'] = 'DocarrayIndexerV2'
+        config['indexer_uses'] = 'DocarrayIndexerV3/latest'
```
florian-hoenicke, line 153: Using latest will lead to problems. Assume you are using Jina NOW v1 with the latest tag.
In Jina NOW v2, we will upgrade our indexers. But then v1 will get the latest.
We need to fix all executor versions.

florian-hoenicke (changes requested):

joschkabraun (approved):
LGTM!

deployment/bff/app/v1/routers/helper.py
```diff
@@ -34,7 +34,7 @@ def process_query(
         filter_query = []
 
         for key, value in conditions.items():
-            filter_query.append({f'tags__{key}': {'$eq': value}})
+            filter_query.append({f'{key}': {'$eq': value}})
```
joschkabraun, line 37: need to add comments in heer on how the conditions are constructed with its interdependency with DocArrayIndexer

now_common/utils.py
```diff
@@ -141,17 +144,27 @@ def get_indexer_config(num_indexed_samples: int) -> Dict:
...
     else:
         config['indexer_resources'] = {'INDEXER_CPU': 1.0, 'INDEXER_MEM': '4G'}
 
     return config
+
+
+def _extract_tags_annlite(da: DocumentArray):
+    tags = set()
+    for doc in da:
+        if len(doc.tags.keys()) > 0:
```
florian-hoenicke, line 166: can be removed

PR #455, opened 2022-09-01
fix: file endings

florian-hoenicke:
the file endings were broken. It was not possible to load for instance images from disk


now/data_loading/data_loading.py
```diff
@@ -84,46 +83,56 @@ def _open_s3_json(path: str, user_input: UserInput):
...
+
+def _load_tags_from_json_if_needed(da: DocumentArray, user_input: UserInput):
+    if any([doc.uri.endswith('.json') for doc in da]):
+        return _load_tags_from_json(da, user_input)
+    else:
+        return da
+
+
+def _load_tags_from_json(da, user_input):
+    print('Loading tags!')
```
joschkabraun, line 111: I would print how the tags are loaded (, i.e., assumed folder structure), as otherwise the documents will simply disappear and user won't know why.

joschkabraun (approved):
I like the refactoring! After comment LGTM :rocket:

azayz (approved):

now/data_loading/data_loading.py
florian-hoenicke: done, thanks

PR #459, opened 2022-09-02
Refactor dialog app options

makram93:


now/now_dataclasses.py
```diff
@@ -1,32 +1,55 @@
...
-from typing import Dict, List, Optional
+from typing import Any, Callable, Dict, List, Optional, Union
 
 from now.constants import Apps, DatasetTypes
 
 
 @dataclass
 class UserInput:
     app: Optional[Apps] = None
+    app_instance: Optional['JinaNOWApp'] = None  # noqa: F821
```
joschkabraun, line 18: we should either have an attribute with `Apps` or `JinaNOWApp` but not both

now_common/options.py
```diff
@@ -0,0 +1,389 @@
...
+    name='user_emails',
+    prompt_message='Please enter the comma separated Email IDs'
+    'who will have access to this flow:',
+    prompt_type='input',
+    depends_on=ADDITIONAL_USERS,
+    conditional_check=lambda user_inp: user_inp.additional_user,
+)
+
+
+def _check_if_existing():
```
joschkabraun, line 268: Given the new location of the function, the name isn't descriptive enough anymore such that I would change it

now/apps/image_to_text/app.py
```diff
@@ -94,7 +94,7 @@ def preprocess(
             split_by_sentences = False
             if (
                 user_input
-                and user_input.is_custom_dataset
+                and user_input.data == 'custom'
```
joschkabraun, line 97: other apps also check `user_input.is_custom_dataset`

now/now_dataclasses.py
```diff
@@ -1,32 +1,55 @@
...
 
     # data related
     data: Optional[str] = None
-    is_custom_dataset: Optional[bool] = None
```
joschkabraun, line 13: this attribute is used in some tests and executors AFAIK

now_common/options.py
```diff
@@ -0,0 +1,389 @@
...
+    )()
+
+    if user_input:
+        user_input.app_instance = app_instance
+
+    return app_instance
+
+
+@time_profiler
+def _setup_cluster(user_input: UserInput, **kwargs) -> None:
```
joschkabraun, line 287: Is that correct? The name is very similar to `run_all_k8s.setup_cluster` ...

now/dialog.py
```diff
@@ -6,392 +6,54 @@
...
-    _configure_cluster(user_input, **kwargs)
-    app_instance.run_checks(user_input)
-    if user_input.deployment_type == 'remote':
-        if _configure_security(user_input, **kwargs):
-            if _configure_additional_user(user_input, **kwargs):
-                _configure_user_emails(user_input, **kwargs)
-    return app_instance, user_input
+    _configure_app_options(user_input, **kwargs)
 
+    return user_input.app_instance, user_input
```
joschkabraun, line 27: given that now `user_input` contains the app; we don't need to return both

tests/integration/test_end_to_end.py
```diff
@@ -76,34 +76,34 @@ def test_token_exists():
             DemoDatasets.BIRD_SPECIES,
             'local',
         ),
-        (
```
joschkabraun, line 79: I think you forgot to uncomment those again

tests/unit/test_dialog.py
```diff
@@ -213,10 +174,7 @@ def __call__(self, question: Dict):
             'deployment_type': 'local',
             'quality': 'medium',
         },
-        UserInput(
-            is_custom_dataset=False,
-            create_new_cluster=True,
-        ),
+        UserInput(),
```
joschkabraun, line 177: The last parameter is the `expected_user_input`; so we test them all to be empty? I.e. the tests says everything is good, when we don't actually have set anything?

joschkabraun (dismissed):

now/dialog.py
```diff
@@ -6,392 +6,54 @@
...
-            },
-        ],
-        prompt_message='What sort of search engine would you like to build?',
-        prompt_type='list',
-        **kwargs,
-    )
 
+def _configure_app_options(user_input, **kwargs):
+    # First we configure the app and initialize it
+    configure_option(JinaNOWApp.ask_app(), user_input, **kwargs)
```
joschkabraun, line 32: If we refactor this, we can get rid of having both `app` and `app_instance` in `user_input` which I believe is confusing to ask.
Also, an app shouldn't be able to configure this as we need to ask this basic question at the beginning of `jina now start`.

now/finetuning/run_finetuning.py
```diff
@@ -12,14 +12,14 @@
...
 from finetuner.callback import EarlyStopping, EvaluationCallback
+from now_common.options import UserInput
 
 from now.apps.base.app import JinaNOWApp
 from now.deployment.deployment import cmd, terminate_wolf
 from now.deployment.flow import deploy_flow
 from now.finetuning.dataset import FinetuneDataset, build_finetuning_dataset
 from now.finetuning.settings import FinetuneSettings
 from now.log import time_profiler, yaspin_extended
-from now.now_dataclasses import UserInput
```
joschkabraun, line 22: Why importing `UserInput` from `now_common.options` and not from `now.now_dataclasses` where it is actually located?

tests/unit/test_dialog.py
makram93: It is updated inside the function

now/now_dataclasses.py
makram93: Removed the occurences. We only rely on data='custom'

tests/unit/test_dialog.py
joschkabraun: Ok. Let's remove it from the `parametrize` then

now/dialog.py
```diff
@@ -6,392 +6,57 @@
...
-        user_input.aws_region_name = _prompt_value(
-            name='aws_region_name',
-            prompt_message='Please enter the AWS region:',
-            prompt_type='input',
-            **kwargs,
-        )
 
+def _configure_app_options(user_input, **kwargs):
+    # First we configure the app and initialize it
+    configure_option(JinaNOWApp.ask_app(), user_input, **kwargs)
```
joschkabraun, line 32: I would remove `ask_app` from `JinaNOWApp` as it's part of the framework and not a single app

now/apps/base/app.py
```diff
@@ -24,26 +26,30 @@
...
     @property
     def app_name(self) -> str:
         """
~        Name of the app. Should be an enum value set in now.constants.Apps
~        """
~        raise NotImplementedError()
~
~    @property
~    def is_enabled(self) -> bool:
~        """
```
joschkabraun: I think `property` makes more sense or lmk if `staticmethod` makes more sense

now/apps/base/app.py
```diff
@@ -24,26 +26,30 @@
         self.flow_yaml = ''
         self.set_flow_yaml()
 
+    @staticmethod
+    def ask_app():
```
joschkabraun, line 30: we can remove that after changing `dialog`

now/apps/base/app.py
```diff
@@ -24,26 +26,30 @@
...
     def app_name(self) -> str:
         """
~        Name of the app. Should be an enum value set in now.constants.Apps
~        """
~        raise NotImplementedError()
~
~    @property
~    def is_enabled(self) -> bool:
~        """
~        Set to True if this app is enabled for the end user.
```
joschkabraun: I would update the docstring to reflect how this property is now going to be used. Also, would show how it can be extended

now/run_backend.py
```diff
@@ -72,19 +71,22 @@
     return_results: Optional[bool] = False,
 ):
     request_size = estimate_request_size(dataset)
+    # Remove app_instance from parameters
+    parameters['user_input']['app'] = parameters['user_input']['app_instance'].app_name
+    parameters['user_input'].pop('app_instance')
```
joschkabraun, line 76: I don't think that is a good solution as this changes the interface and executor don't expect `app` attribute anymore

joschkabraun (changes requested):
We also need to update `.gitignore` to not save `.lock` files in `dask-worker-space`.

now/run_backend.py
```diff
@@ -52,7 +51,7 @@
 
     print(f"▶ indexing {len(dataset)} documents")
     params = {
-        'user_input': user_input.dict(),
+        'user_input': user_input.__dict__,
```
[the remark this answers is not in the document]
makram93: Yes, it works the same way since pydantic dataclasses under the hood has dataclasses.dataclass

now/run_backend.py
makram93: We can not send an class object to as param and currently all user_input is convert to dict and passed as param. We have to remove the app_instance, however, since `app` is not expected anymore I will remove it from that line. Thanks for pointing it out

now_common/options.py
```diff
@@ -0,0 +1,387 @@
...
+        },
+        {'name': '📝 ▶ 📝 text to text search', 'value': Apps.TEXT_TO_TEXT},
+        {
+            'name': '📝 ▶ 🎦 text to video search (gif only at the moment)',
+            'value': Apps.TEXT_TO_VIDEO,
+        },
+        {
+            'name': '🥁 ▶ 🥁 music to music search',
+            'value': Apps.MUSIC_TO_MUSIC,
+        },
```
violenil, line 51: should `Apps.TEXT_TO_TEXT_AND_IMAGE` be an option here?

makram93: Yes It should be but only added to this list once this feature is integrated and ready to use.

joschkabraun (approved):
LGTM! :rocket: Really nice refactoring :) I like how now arguments are exposed to the CLI

PR #460, opened 2022-09-02
fix: missing space in sentence

alexcg1:
Fixes `Huband` in this text:

```
Jina NOW trains a model, pushes it to Jina Huband deploys a Flow and a playground app in the cloud or locally.
```


joschkabraun (approved):
Good catch & thank you! 🚀

PR #461, opened 2022-09-02
Fix auth exec restart

joschkabraun:
Closes https://github.com/jina-ai/now.private/issues/138


now_executors/NOWAuthExecutor/executor.py
```diff
@@ -112,19 +114,39 @@ def __init__(
         self.api_keys = api_keys if api_keys else []
         self._user = None
 
+        self.api_keys_path = (
+            os.path.join(self.workspace, 'api_keys.json') if self.workspace else None
```
makram93, line 118: In case of pod restarts, will the `workspace` be persistent? Which storage is it pointing to?

now_executors/NOWAuthExecutor/executor.py
```diff
@@ -112,19 +114,39 @@ def __init__(
...
+        )
+        self.user_emails_path = (
+            os.path.join(self.workspace, 'user_emails.json') if self.workspace else None
+        )
+
+        if self.api_keys_path and os.path.exists(self.api_keys_path):
+            with open(self.api_keys_path, 'r') as fp:
+                self.api_keys = json.load(fp)
```
makram93, line 126: Should we override `api_keys` here? Can we append it the value from file to `self.api_keys` or read from files only when `api_keys` is not provided?

now_executors/NOWAuthExecutor/executor.py
```diff
@@ -112,19 +114,39 @@ def __init__(
...
 
     @secure_request(on='/admin/updateApiKeys', level=SecurityLevel.ADMIN)
     def update_api_keys(self, parameters: Dict = None, **kwargs):
         """
         Update the api keys during runtime. That way, we don't have to restart it.
         """
         self.api_keys = parameters['api_keys']
+        if self.api_keys_path:
+            with open(self.api_keys_path, 'w') as fp:
+                json.dump(self.api_keys, fp)
```
makram93, line 149: Why is `update` -> overwriting existing values? Shouldn't it append new values to the old values?

now_executors/NOWAuthExecutor/executor.py
joschkabraun: Yes, the workspace is persistent to restarts.

now_executors/NOWAuthExecutor/executor.py
joschkabraun: Even though I didn't implement this logic, I think that is the right solution to always overwrite. Only appending would require another endpoint which allows deletion of api keys. This solution is briefer.

now_executors/NOWAuthExecutor/executor.py
joschkabraun: No, `self.api_keys_path` and `self.user_emails_path` only exist if there have been updates since the pod was started the first time. After a restart the executor is initialised with the old `api_keys` and `user_mails` it got when it was created the first time. This would lead to inconsistent states, therefore we need to overwrite.

makram93 (approved):
LGTM!

PR #462, opened 2022-09-02
refactor: ci for preprocessor

joschkabraun:
Makes the filters for building of the preprocessor more restrictive in order to avoid unnecessary updates of preprocessor constants.

Note: from now on any shared preprocessing functions reside in `now_common/preprocess.py`


florian-hoenicke (approved):
LGTM 
Moving the code to a separated package makes a lot of sense.
Why was the data loading in that list before?

joschkabraun:
> LGTM
> Moving the code to a separated package makes a lot of sense.
> Why was the data loading in that list before?

There was some dependency for loading from S3, but this isn't up-to date anymore.

PR #465, opened 2022-09-05
fix: secure preprocessor

joschkabraun:
- [x] closes part 2 of https://github.com/jina-ai/now.private/issues/124 by letting `NOWPreprocessor` extend `NOWAuthExecutor`
- [x] closes https://github.com/jina-ai/now.private/issues/151
- [x] refactor: set Jina version the same as is installed in all flows
- [x] reported bug in core for loading empty lists from environment variables into flow yaml: https://github.com/jina-ai/jina/issues/5142
- [x] reported same bug in jcloud after core fixed it: https://github.com/jina-ai/jcloud/issues/81


now_executors/NOWAuthExecutor/executor.py
```diff
@@ -48,56 +53,47 @@ def _check_user(kwargs, level, user_emails, admin_emails, api_keys):
...
-        # TODO hack - user_info['_id'] should not be used. Use email once hubble provides it
-        if user_info['_id'] == email:
-            return
-
-        if 'identity' in jwt and jwt['identity']['userInfo']['email'] == email:
-            if not jwt['identity']['userInfo']['email_verified']:
-                raise PermissionError(
-                    f'User {email} needs to be verified. Check if you received verification link.'
-                )
+        if user_info.get('email') == email:
```
makram93, line 58: We should still check if the email is verified. I can create an account with your email ID and send my token whose email ID will match the saved IDs. So, verification check should still be in place

deployment/bff/app/v1/routers/img2img.py
```diff
@@ -25,7 +25,6 @@ def index(data: NowImageIndexRequestModel):
     `base64` encoded using human-readable characters - `utf-8`.
     """
     index_docs = DocumentArray()
-    jwt = data.jwt
```
makram93, line 28: Is this intended? How do we pass jwt then from BFF?

now/deployment/flow.py
```diff
@@ -114,11 +114,8 @@ def _extend_flow_yaml(flow_yaml, tmpdir, secured, admin_emails, user_emails):
   - name: security_check
     uses: jinahub+docker://NOWAuthExecutor/v{NOW_AUTH_EXECUTOR_VERSION}
     uses_with:
-      admin_emails: {admin_emails if admin_emails else []}
-      user_emails: {user_emails if user_emails else []}
-    jcloud:
-      resources:
-        memory: 1G
```
makram93, line 121: Did we remove this jcloud resource because that is the default resource allocated?

now_executors/NOWAuthExecutor/executor.py
```diff
@@ -33,6 +35,9 @@ def wrapper(*args, **kwargs):
 
 
 def _check_user(kwargs, level, user_emails, admin_emails, api_keys):
+    if user_emails == [] and admin_emails == [] and api_keys == []:
```
makram93, line 38: When would this scenario be? We only attach security executor when the flow is to be secured and then we pass email/api. So, any of these attributes will always have some value if this executor is added to the flow?

now_executors/NOWPreprocessor/executor.py
```diff
@@ -6,8 +6,10 @@
 from typing import Dict, Optional
 
 import boto3
-from jina import Document, DocumentArray, Executor, requests
+from jina import Document, DocumentArray
 from now_common.options import _construct_app
+from now_executors import NOWAuthExecutor as Executor
```
makram93, line 11: Is this import possible? NOWPreprocessor is an independent executor and when pushed, it will have no reference to the now_executors

now_executors/NOWAuthExecutor/executor.py
joschkabraun: Email won't be sent back anymore from Hubble for a given token if it isn't verified.

deployment/bff/app/v1/routers/img2img.py
joschkabraun: Yes, this is actually handled in `bff/app/v1/routers/helper:jina_client_post`, the `jwt` variable was never used.

now/deployment/flow.py
joschkabraun: No, default resource allocation is 100 mb, but I think 1 gb is too much for authentication executor.

now_executors/NOWAuthExecutor/executor.py
joschkabraun: We need this to let the `NOWPreprocessor` extend `NOWAuthExecutor`. We don't always have securitisation of a flow but we always have a preprocessor.

now_executors/NOWPreprocessor/executor.py
joschkabraun: Yes. `NOWPreprocessor` has in its `requirements.txt` the `now` package. From there it imports the `NOWAuthExecutor` but it doesn't need the pushed and built executor.

makram93 (approved):
LGTM! 
Let's wait for florian's review as well. During the initial implementation, he suggested not relying on just email and was against passing tokens only. I am not sure if it is still valid. 
cc @florian-hoenicke

joschkabraun:
> LGTM!
> Let's wait for florian's review as well. During the initial implementation, he suggested not relying on just email and was against passing tokens only. I am not sure if it is still valid.
> cc @florian-hoenicke

Thank you for the feedback! Just remembered that I discussed this in an earlier pairing already, so I will discuss it once the `jcloud` bug is resolved and CI is green.

florian-hoenicke:
@makram93 Back then we couldn't get the email from j cloud. But now we get it so it is fine

PR #468, opened 2022-09-06
fix: loading tags

joschkabraun:
- [x] fixes issue that `json` files are not loaded when listing files from S3
- [x] speeds up loading of tags from S3 & disk

Also,
- removes unused function `_download_from_s3_bucket ` (implemented in `NOWPreprocessor`)
- fixes missing `traversal_path` to `text-to-video` app flows


now/apps/text_to_video/flow-video-clip-high-performance.yml
```diff
@@ -48,7 +50,6 @@ executors:
       index_traversal_paths: '@c'
       search_traversal_paths: '@c'
       dim: ${{ ENV.PRE_TRAINED_EMBEDDINGS_SIZE }}
-      columns: ${{ ENV.COLUMNS }}
```
azayz, line 51: why is this being deleted ? are we using only docarray indexer or we don't expect to do filtering in this case ?

now/data_loading/data_loading.py
```diff
@@ -96,6 +108,7 @@ def merge_documents(user_input, content_file, json_file):
             tags = _open_json(json_file)
     else:
         tags = {}
+    tags = flatten_dict(tags)
```
azayz, line 111: the news tags files contain nested dictionaries ?

now/data_loading/data_loading.py
```diff
@@ -116,17 +129,24 @@ def _load_tags_from_json(da, user_input):
         folder = d.uri.rsplit('/', 1)[0]
         folder_to_files[folder].append(d.uri)
     merged_documents = []
-    for _, files in folder_to_files.items():
-        document = get_document(files, user_input)
-        if document:
-            merged_documents.append(document)
+    with ThreadPoolExecutor() as thread_executor:
```
azayz, line 132: I assume this is for speeding things up ?

now/apps/text_to_video/flow-video-clip-high-performance.yml
joschkabraun: Good catch, wasn't supposed to be there. Thanks!

now/data_loading/data_loading.py
joschkabraun: Yes, they still do. I think we should have this solution in place for the general case. If the `tags` are 'flat', then they won't be changed.

now/data_loading/data_loading.py
joschkabraun: Yes

now/apps/text_to_video/flow-video-clip-high-performance.yml
```diff
@@ -30,13 +30,15 @@ executors:
...
       traversal_paths: '@c'
+      access_paths: '@c'
       name: ${{ ENV.PRE_TRAINED_MODEL_NAME }}
     env:
       JINA_LOG_LEVEL: DEBUG
   - name: postprocessor
     uses: jinahub+docker://NOWPostprocessor/v0.0.5
     uses_with:
       traversal_paths: '@c'
+      access_paths: '@c'
```
makram93, line 41: I don't know about this but why do we have/need two paths arguments?

makram93 (approved):
LGTM!

now/apps/text_to_video/flow-video-clip-high-performance.yml
joschkabraun: Yes, some executors have adopted to new naming and some don't.

PR #470, opened 2022-09-06
feat: adapt finetuning to nest

jupyterjazz:
Adapt the finetuning module to nest with **minimal** changes.
We can refactor this module better, but i don't think it's a priority now. 

~waiting for https://github.com/jina-ai/now/pull/459 to be merged to have access on app instance from the user_input.~


now/finetuning/run_finetuning.py
```diff
@@ -51,207 +51,222 @@
...
-            'l2': True,
-            'bias': False if finetune_settings.bi_modal else True,
-        },
+        model=finetune_settings.model_name,
+        model_options=_get_model_options(finetune_settings),
+        loss=finetune_settings.loss,
         train_data=finetune_ds.train,
         experiment_name=experiment_name,
         eval_data=finetune_ds.val,
         callbacks=callbacks,
```
joschkabraun, line 135: would leave it as `_maybe_add_embeddings` as it only will add embeddings if they don't exist

joschkabraun (dismissed):
We should move any app dependent logic from `settings.py` into the apps to have it all in the same place.

now/finetuning/settings.py
```diff
@@ -67,12 +77,16 @@ def _is_bi_modal(user_input: UserInput, dataset: DocumentArray) -> bool:
...
     finetune_datasets: Tuple = (),
+    pre_trained_embedding_size: Optional[int] = None,
+    encoder_type: Optional[str] = None,
 ) -> FinetuneSettings:
     """This function parses the user input configuration into the finetune settings"""
     return FinetuneSettings(
-        pre_trained_embedding_size=pre_trained_embedding_size,
         perform_finetuning=_is_finetuning(user_input, dataset, finetune_datasets),
         bi_modal=_is_bi_modal(user_input, dataset),
+        model_name=user_input.app.finetuning_model_name(encoder_type),
```
violenil, line 88: how will this work when we want to finetune two models (sbert and clip)? Will we get two `FinetuneSettings`?

jupyterjazz: Exactly!

now/apps/text_to_text_and_image/app.py
```diff
@@ -64,3 +64,23 @@ def setup(
         # only implements data generation
         data = DataBuilder(dataset=dataset, config=user_input.task_config).build()
         return {}
+
+    @property
+    def finetuning_model_name(self, encoder_type: Optional[str] = None) -> str:
```
florian-hoenicke, line 69: This property makes the assumption that each app uses a model to fine-tune.

florian-hoenicke: I think finetuning should not be a property of a base app

now/apps/base/app.py
```diff
@@ -252,3 +252,16 @@ def bff_mapping_fns(
     def index_query_access_paths(self) -> str:
         """Gives access paths for indexing and searching."""
         return '@r'
+
+    def finetuning_model_name(self, encoder_type: Optional[str] = None) -> str:
+        """Name of the model used in case of fine-tuning."""
+        return 'mlp'
+
+    def loss_function(self, encoder_type: Optional[str] = None) -> str:
```
florian-hoenicke, line 260: I don't think finetuning should be part of the base app for two reasons:
1. an app might not need fine-tuning
2. an app might use multiple loss functions and models. So it is hard to generalize.

The first point could be tackled by introducing a fine-tuning Mixin

florian-hoenicke (changes requested):
I think it is good that you are tackling this.
It is a necessary refactoring.
The open point of discussion is where to put the fine-tuning logic.
I think the base app should be very general and not include app-specific things like fine-tuning. We should rather put it into our common package for now.

At a later point in time I would suggest to have mixins (similar to docarray) for all the things we have in the common package.

joschkabraun:
I think we can move all that logic to the `setup` function. Some things are shared by many apps which then can go into `now_common`. The new losses, etc. for `text-to-text-and-image` go in that app folder.

PR #471, opened 2022-09-07
feat: add list endpoint

florian-hoenicke:
For listed documents, we can get all the meta data but ignore the content


florian-hoenicke:
Thanks for your suggestion @makram93 It is now much more performant

PR #473, opened 2022-09-07
feat: add filtering playground

azayz:
closes https://github.com/jina-ai/now.private#102


now_common/flow/flow-clip.yml
```diff
@@ -29,7 +29,7 @@ executors:
     env:
       JINA_LOG_LEVEL: DEBUG
   - name: postprocessor
-    uses: jinahub+docker://NOWPostprocessor/v0.0.5
+    uses: jinahub+docker://NOWPostprocessorV2/v0.0.5
```
joschkabraun, line 32: any `flow.yaml` needs to be updated to the new post-processor

deployment/playground/playground.py
```diff
@@ -120,9 +120,21 @@ def deploy_streamlit():
             from better_profanity import profanity
 
             profanity.load_censor_words()
-
         setup_design()
 
+        if params.host:
+            client = Client(host=params.host)
```
joschkabraun, line 126: This seems a bit like an antipattern given that we usually call the BFF and now we call the client.. what do you think?

now_executors/NOWPostprocessor/executor.py
```diff
@@ -38,11 +50,24 @@ def maybe_drop_blob_tensor(
...
                     except FileNotFoundError:
                         continue
+
         return docs
 
+    def close(self):
+        with open('./tags.pkl', 'wb') as f:
+            pickle.dump(self.tags_values, f)
+
+    @requests(on='/tags')
```
joschkabraun, line 60: Should use `secure_request` as NOWPreprocessor does in order to secure the data

now_executors/NOWPostprocessor/executor.py
```diff
@@ -1,11 +1,14 @@
...
+from collections import defaultdict
 from typing import Dict
 
 from jina import Document, DocumentArray, Executor, requests
 
 CLOUD_BUCKET_PRE_FIXES = ['s3://']
 
 
-class NOWPostprocessor(Executor):
+class NOWPostprocessorV2(Executor):
```
joschkabraun, line 11: post processor needs to extend `NOWAuthExecutor` given that is handling data

joschkabraun (changes requested):
Given that we are getting rid of docarray indexer, we should move all the logic / changes in the post processor into the `NOWAnnLiteIndexer` then we also don't have to worry about building new Postprocessor and securitizing it

deployment/playground/playground.py
azayz: I do agree with you but it felt unnecessary to create new BFF endpoint which will not be exposed to the clients (/tags), a direct call from the playground to the flow seemed more logical  to retrieve just the tags and their values

joschkabraun (changes requested):
I wouldn't change the `NOWPostprocessor` but move all those changes into `NOWAnnLiteIndexer` as this is the only indexer moving forward after https://github.com/jina-ai/now/pull/491 has been merged. You can modify `NOWAnnLiteIndexer` already and changes will be automatically built on CI.

deployment/playground/src/search.py
```diff
@@ -130,11 +133,12 @@ def call_flow(url_host, data, attribute_name, domain):
     return docs
 
 
-def search_by_text(search_text, jwt) -> DocumentArray:
-    return search('text', search_text, jwt)
+def search_by_text(search_text, jwt, filter_selection) -> DocumentArray:
+    print('search by text : ', filter_selection)
```
joschkabraun, line 137: would remove print

deployment/playground/playground.py
```diff
@@ -120,9 +121,22 @@ def deploy_streamlit():
             from better_profanity import profanity
 
             profanity.load_censor_words()
-
         setup_design()
 
+        filters = None
+        if params.host:
+            client = Client(host=params.host)
+            response = client.post(on='/tags')
```
joschkabraun, line 129: in case the fow is secured, we need to also provide the `token` of the user

now_executors/NOWAnnLiteIndexer/executor.py
```diff
@@ -92,6 +95,12 @@ def __init__(
 
         self.da = DocumentArray(sorted([d for d in self.da], key=lambda x: x.id))
 
+        if os.path.isfile('./tags.pkl'):
```
joschkabraun, line 98: Nice that you persist this! I would persist it though in `self.workspace` if it exists (some logic is in `NOWAuthExecutor`)

now_executors/NOWAnnLiteIndexer/executor.py
```diff
@@ -127,8 +136,22 @@ def index(
...
+
+        for doc in flat_docs:
+            for tag, value in doc.tags.items():
+                self.tags_values[tag].add(value)
+
+        self.save_tags()
         return DocumentArray([])
 
+    @secure_request(on='/tags', level=SecurityLevel.USER)
+    def get_tags_and_values(self, **kwargs):
```
joschkabraun, line 148: can you please add some docs here on what it returns and the format of it

joschkabraun (commented):
The PR looks good! For consistency, we should also handle the tags for the `update` and `delete` endpoint`.

makram93 (commented):
Playground changes look good. Nice work! 👍 
For Indexer, I have left some comments, let me know if the suggestion makes sense.

now_executors/NOWAnnLiteIndexer/executor.py
```diff
@@ -164,6 +211,15 @@ def delete(self, parameters: dict = {}, **kwargs):
             filtered_docs = deepcopy(self.da.find(filter=filter))
             self.delete_inmemory_docs(filtered_docs)
             self._index.delete(filtered_docs)
+            for doc in filtered_docs:
+                for tag, value in doc.tags.items():
+                    if value in self.tags_values[tag]:
+                        self.tags_values[tag].remove(value)
```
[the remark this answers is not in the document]
azayz: I agree, thanks for the valuable comment, I changed the code accordingly

now_executors/NOWAnnLiteIndexer/executor.py
```diff
@@ -127,8 +136,35 @@ def index(
 
         self._index.index(flat_docs)
         self.extend_inmemory_docs(flat_docs)
+
+        for doc in flat_docs:
+            for tag, value in doc.tags.items():
+                self.tags_values[tag].add(value)
```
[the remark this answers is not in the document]
azayz: I don't see how moving the logic inside will reduce iterations, can you elaborate ?

makram93 (approved):
LGTM! 🚀

now_executors/NOWAnnLiteIndexer/executor.py
```diff
@@ -126,9 +142,33 @@ def index(
...
+                Document(.., tags={'color':'blue'}),
+                Document(.., tags={'greeting':'hello'}),
+            ])
+
+        the resulting response would be a document array with
+        one document containg a dictionary in tags like the following:
+        {'tags':{'color':['red', 'blue'], 'greeting':['hello']}}
+        """
+        tag_values = defaultdict(set)
+        for id, tags in self.doc_id_tags.items():
```
joschkabraun, line 165: I would replace with `self.doc_id_tags.values()`

joschkabraun (approved):
LGTM! :rocket:

PR #474, opened 2022-09-07
feat: add elastic indexer

violenil:
This PR introduces the ElasticIndexer which should be usable by any NOW app. It should accommodate the following:

- [x] allow indexing of multi-modal documents in an Elasticsearch index (index endpoint)
- [x] make script-score queries that combine BM25 and vector search (search endpoint)
- [x] list all documents in index (list endpoint)


joschkabraun:
I am not sure if you already plan on doing that. But when you are ready to push the first version, can put building of the indexer on CI similar to how it is done for the `NOWPreprocessor`, `NOWAuthExecutor` and `NOWAnnLiteIndexer` (https://github.com/jina-ai/now/blob/main/.github/workflows/ci.yml#L89). For that it's also important to log out of the jina account before pushing.

now_executors/elastic_indexer/elastic_indexer.py
```diff
@@ -0,0 +1,279 @@
...
+                print("Modality not found")
+                raise
+        if embeddings and len(embeddings) == 2:
+            return embeddings
+        else:
+            print("No/not all embeddings extracted")
+            raise
+
+
+if __name__ == "__main__":
```
florian-hoenicke, line 215: Could you please create an integration test out of that code snippet instead of having it as the main function?

now_executors/elastic_indexer/elastic_indexer.py
```diff
@@ -0,0 +1,279 @@
...
+            "index": "true",
+        },
+    }
+}
+
+
+class ElasticIndexer(Executor):
+    def __init__(
+        self,
+        es_connection_str: Optional[str] = 'http://localhost:9200',
```
florian-hoenicke, line 40: How is access to private remote clusters handled? Is everything we require contained in the connection string?

now_executors/elastic_indexer/elastic_indexer.py
```diff
@@ -0,0 +1,279 @@
...
+        distance: str = 'cosine',
+        index_name: str = 'nest',
+        es_mapping: Optional[Dict] = None,
```
florian-hoenicke, line 43: is there a reasonable default value? I think by default, it should assume we just insert documents with single embeddings.

now_executors/elastic_indexer/elastic_indexer.py
```diff
@@ -0,0 +1,279 @@
...
+            "dims": 512,
+            "similarity": "cosine",
+            "index": "true",
+        },
+    }
+}
+
+
+class ElasticIndexer(Executor):
+    def __init__(
```
florian-hoenicke, line 38: we should also add a limit parameter

now_executors/elastic_indexer/elastic_indexer.py
```diff
@@ -0,0 +1,279 @@
...
+
+        self.es_connection_str = es_connection_str
+        self.distance = distance
+        self.index_name = index_name
+        if not es_mapping:
+            self.es_mapping = MAPPING
+        else:
+            self.es_mapping = es_mapping
+
+        self.es = Elasticsearch(es_connection_str, verify_certs=False)
```
florian-hoenicke, line 64: verify certs = False does look a bit wrong. Is this introducing some security issue or is it fine?

now_executors/elastic_indexer/elastic_indexer.py
```diff
@@ -0,0 +1,279 @@
...
+        :param index_name: ElasticSearch Index name used for the storage
+        :param es_mapping: Mapping for new index.
+        """
+        super().__init__(**kwargs)
+
+        self.es_connection_str = es_connection_str
+        self.distance = distance
+        self.index_name = index_name
+        if not es_mapping:
+            self.es_mapping = MAPPING
```
florian-hoenicke, line 60: MAPPING can directly be taken as default value

now_executors/elastic_indexer/elastic_indexer.py
florian-hoenicke, line 38: I think it would make sense to also add `dim` and explain in the docstring that `dim` and `metric` are only used in case no mapping is defined

now_executors/elastic_indexer/elastic_indexer.py
```diff
@@ -0,0 +1,279 @@
...
+        },
+    }
+}
+
+
+class ElasticIndexer(Executor):
+    def __init__(
+        self,
+        es_connection_str: Optional[str] = 'http://localhost:9200',
+        distance: str = 'cosine',
```
florian-hoenicke, line 41:
```suggestion
        metric: str = 'cosine',
```

now_executors/elastic_indexer/elastic_indexer.py
florian-hoenicke, line 38: please add `traversal_paths`

now_executors/elastic_indexer/elastic_indexer.py
```diff
@@ -0,0 +1,279 @@
...
+        """
+        es_docs = self._transform_docs_to_es(docs)
+        bulk(self.es, es_docs, refresh="wait_for")
+        return (
+            DocumentArray()
+        )  # prevent sending the data back by returning an empty DocumentArray
+
+    @requests(on="/search")
+    def search(
+        self, docs: Union[Document, DocumentArray], limit: Optional[int] = 20, **kwargs
```
florian-hoenicke, line 84: allow filtering by tags as well similar to annlite

now_executors/elastic_indexer/elastic_indexer.py
```diff
@@ -0,0 +1,279 @@
...
+
+    @requests(on="/filter")
+    def filter(self, parameters: dict = {}, **kwargs):
+        """
+        TODO: implement filter query for Elasticsearch
+
+        :returns: filtered results in root, chunks and matches level
+        """
+        raise NotImplementedError()
+
```
florian-hoenicke, line 149: We need a list endpoint for the sync job to get all existing file URIs.

now_executors/elastic_indexer/elastic_indexer.py
```diff
@@ -0,0 +1,279 @@
...
+    def _transform_docs_to_es(self, docs: DocumentArray) -> List[Dict]:
```
florian-hoenicke, line 150: I think this function should also support the simple case where we just have one document with a single embedding. If this is supported, all the other NOW apps can use this indexer.  Maybe, you could make the transformation based on the provided mapping. But if it is too complicated, then we need another way to support single documents.

now_executors/elastic_indexer/elastic_indexer.py
```diff
@@ -0,0 +1,279 @@
...
+                index=self.index_name,
+                query=query,
+                fields=['text'],
+                source=False,
+                size=limit,
+            )["hits"]["hits"]
+            doc.matches = self._transform_es_results_to_matches(result)
+        return docs
+
+    def _build_es_query(
```
florian-hoenicke, line 103: Score function should also handle the simple case where we just have one document with one embedding

florian-hoenicke (changes requested):
This is a good step to add the elastic search indexer to the NOW repo.
It will give all NOW apps more reliability.
I requested some changes which are required to make the elastic search indexer work with the other apps.

girishc13 (commented):
Hi there, don't mind me randomly popping in to provide some suggestions. Feel free to ignore the some suggestions if it doesn't support the immediate goal.

now_executors/elastic_indexer/elastic_indexer.py
violenil: Is it not sufficient to have a limit parameter on the search endpoint?

joschkabraun (changes requested):
Can you put building of this indexer on CI similar to how it is done for the `NOWPreprocessor`, `NOWAuthExecutor` and `NOWAnnLiteIndexer` (https://github.com/jina-ai/now/blob/main/.github/workflows/ci.yml#L89). For that it's also important to log out of the jina account before pushing to get a secret to build it.

now_executors/elastic_indexer/elastic_indexer.py
violenil: resolved using `es_config`

now_executors/elastic_indexer/elastic_indexer.py
```diff
@@ -0,0 +1,205 @@
...
+        self.es_mapping = es_mapping
+
+        self.es = Elasticsearch(hosts=self.hosts, **es_config)
+        if not self.es.indices.exists(index=self.index_name):
+            self.es.indices.create(index=self.index_name, mappings=self.es_mapping)
+
+    @requests(on="/index")
+    def index(self, docs: DocumentArray, **kwargs) -> DocumentArray:
+        """
+        Index new `Document`s by adding them to the Elasticsearch indeex.
```
jupyterjazz, line 66:
```suggestion
        Index new `Document`s by adding them to the Elasticsearch index.
```

now_executors/elastic_indexer/elastic_indexer.py
```diff
@@ -0,0 +1,205 @@
...
+        for c in doc.chunks:
+            if c.modality == 'text':
+                embeddings['text_embedding'] = c.embedding
+            elif c.modality == 'image':
+                embeddings['image_embedding'] = c.embedding
+            else:
+                print("Modality not found")
+                raise
+        if not embeddings:
+            print("No embeddings extracted")
```
jupyterjazz, line 204: should we raise an error here?

now_executors/elastic_indexer/elastic_indexer.py
```diff
@@ -0,0 +1,205 @@
...
+
+        :param doc: A `Document` containing text and image chunks.
+        :return: A list of dictionaries containing text, text embedding and image embedding
+        """
+        new_docs = list()
+        for doc in docs:
+            new_doc = dict()
+            new_doc['_id'] = doc.id
+            for chunk in doc.chunks:
+                if chunk.modality == "text":
```
jupyterjazz, line 162: let's stay consistent with single quotes

jupyterjazz (commented):
This is so exciting!

added few minor comments

now_executors/elastic_indexer/elastic_indexer.py
```diff
@@ -0,0 +1,205 @@
...
+    'l2_norm': 'l2norm',
+}
+
+MAPPING = {
+    "properties": {
+        "id": {"type": "text", "analyzer": "standard"},
+        "text": {"type": "text", "analyzer": "standard"},
+        "text_embedding": {
+            "type": "dense_vector",
+            "dims": 5,
```
jupyterjazz, line 23: why do we have 5 here?

violenil: Good question 😅 I think that was from one of my tests. Will change to 768.

now_executors/elastic_indexer/elastic_indexer.py
violenil: Good point, I guess we should raise an error if we assume by default there should be at least one embedding, and none are extracted.

now_executors/elastic_indexer/elastic_indexer.py
```diff
@@ -0,0 +1,279 @@
...
+        self, docs: Union[Document, DocumentArray], limit: Optional[int] = 20, **kwargs
+    ):
+        """Perform traditional bm25 + vector search.
+
+        :param docs: query `Document`s.
+        :param limit: return top `limit` results.
+        """
+        for doc in docs:
+            query = self._build_es_query(doc)
+            result = self.es.search(
```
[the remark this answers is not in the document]
violenil: Thanks for your suggestion @girishc13! I will take note of this multi-search api and perhaps we can use this in future. For now, I think this solution is sufficient. 

Can you suggest how to 'unmask' error from this search call though? That would be really helpful.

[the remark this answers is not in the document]
girishc13: The method automatically raises one the errors specified [here](https://elasticsearch-py.readthedocs.io/en/v8.4.1/exceptions.html).

now_executors/elastic_indexer/manifest.yml
```diff
@@ -0,0 +1,5 @@
+manifest_version: 1
+name: ElasticIndexer
+description: indexer for elasticsearch
+url: none
+keywords: [index, trivial]
```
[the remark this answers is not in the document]
violenil: Done.

florian-hoenicke (changes requested):
Great implementation. I can not await to see this live. Good work.
The following things need to be done to have it work with the current apps:
- [ ] test-end-to-end should work with one of the apps.
- [x] ~cluster needs to be deployed (choose some cost-efficient deployment)~
  - using local deployment for now
- [x] ~serialize the whole document when indexing and retrieve it on search~
  - saving all parts of document as fields in index (dynamic mapping + chunk fields predefined)
- [x] remove embedding when indexing the documents
- [x] ~create a follow-up ticket for delete endpoint (deleting by tags)~
  - implemented in [7bfd1b5](https://github.com/jina-ai/now/pull/474/commits/7bfd1b57e65efc2dbe2d1e5d7af7786b6eae4d8e)

now_executors/elastic_indexer/tests/test_integration.py
```diff
@@ -0,0 +1,148 @@
...
+                    Document(
+                        embedding=np.array([1, 2, 3, 4, 5]),
+                    ),
+                ],
+            )
+        ]
+    )
+
+
+def test_indexing(multimodal_da):
```
joschkabraun, line 69: All tests of executors are now under `tests/now_executors/...` such that they will executed under CI

.github/workflows/ci.yml
```diff
@@ -261,6 +261,50 @@ jobs:
...
+        env:
+          NOW_ELASTIC_INDEXER_VERSION: ${{ env.NOW_ELASTIC_INDEXER_VERSION }}
+        run: |
+          if jina hub pull jinahub+docker://i7asjny7/v$NOW_ELASTIC_INDEXER_VERSION; then
+            echo "NOW Elastic Indexer version is used already. Please update the tag"
+            exit 1
+          else
+            echo "NOW Elastic Indexer version isn't used already, continue to build..."
+          fi
+      - name: Build and push executor
```
joschkabraun, line 302: Similar to the other executors, I would execute the tests before building it.

now/apps/text_to_text_and_image/app.py
```diff
@@ -138,6 +151,50 @@ def setup(
...
         env_dict['APP'] = self.app_name
+
+        indexer_config = get_indexer_config(len(dataset), elastic=True)
+        # get elastic host
+        num_retries = 0
+        es_password, error_msg = '', b''
+        while num_retries < MAX_RETRIES:
+            es_password, error_msg = cmd(
+                [
+                    "kubectl",
```
florian-hoenicke, line 162: custom kubectl path should be taken to have this run for users without kubectl installed

now_common/options.py
```diff
@@ -53,7 +53,7 @@
         {
             'name': '📝 ▶ 📝+🏞 text to text+image search',
             'value': Apps.TEXT_TO_TEXT_AND_IMAGE,
-            'disabled': AVAILABLE_SOON,
+            # 'disabled': AVAILABLE_SOON,
```
florian-hoenicke, line 56: needs to be disabled for now

now/apps/text_to_text_and_image/app.py
violenil: Have changed this in `utils.py` where this is called for the `text-to-text` app. Currently running the `text-to-text-and-image` app.

now/apps/text_to_text/flow-sbert.yml
```diff
@@ -41,7 +41,8 @@ executors:
   - name: indexer
     uses: ${{ ENV.INDEXER_NAME }}
     uses_with:
-      dim: ${{ ENV.N_DIM }}
+      hosts: ${{ ENV.HOSTS }}
+      dims: ${{ ENV.N_DIM }}
```
joschkabraun, line 45: I would suggest to add `dim` as well to keep backward compatability with AnnLite. Except for we are switching every app to Elasticsearch

now/apps/text_to_text/app.py
```diff
@@ -44,18 +44,23 @@
~        flow_dir = os.path.abspath(os.path.join(__file__, '..'))
~
~        if finetuning:
~            self.flow_yaml = os.path.join(flow_dir, 'ft-flow-sbert.yml')
~        else:
~            self.flow_yaml = os.path.join(flow_dir, 'flow-sbert.yml')
~
```
joschkabraun: Is there a particular reason to only use `elastic` for the `text-to-text` app and not the others?

now_common/utils.py
```diff
@@ -58,13 +65,39 @@
         **encoder_with,
         **indexer_resources,
     }
+
+    if elastic:
```
joschkabraun, line 69: That is duplicate code to `setup` in `text_to_image_and_text`, I would suggest to put that into an own function

now_common/utils.py
```diff
@@ -72,16 +105,16 @@
...
~    if 'NOW_EXAMPLES' in os.environ:
~        valid_app = DEFAULT_EXAMPLE_HOSTED.get(user_input.app_instance.app_name, {})
~        is_demo_ds = user_input.data in valid_app
~        if is_demo_ds:
~            config[
~                'CUSTOM_DNS'
~            ] = f'now-example-{user_input.app_instance.app_name}-{user_input.data}.dev.jina.ai'
~            config['CUSTOM_DNS'] = config['CUSTOM_DNS'].replace('_', '-')
~    return config
~
```
joschkabraun: Given that we not only get the indexer configuration but also spin up Elasticsearch, I would change it to `setup_indexer_get_config` or something similar.

joschkabraun (commented):
Some minor comments, otherwise LGTM :)

now/apps/text_to_text/app.py
violenil: I have not tested it with the other apps, plus the deployment can only run locally (not remote) so I have limited the use of this indexer to the text-to-text app for now. I also suspect there might be unforeseen errors that pop up when using this indexer with other apps. I think this can be tested/introduced in a separate (smaller) pr.

now_common/utils.py
violenil: True. I had separated the process of creating the `env_dict` for the text-to-text-and-image app from the common setup functions, and not considered that there are some overlaps. I have refactored this in [8ef8981](https://github.com/jina-ai/now/pull/474/commits/8ef8981094641ec63070375446d64170cda5cc9b)

florian-hoenicke (approved):

PR #476, opened 2022-09-07
feat: index delete by filter

florian-hoenicke:
this pr allows deleting by filter conditions. It will be used by the sync job to remove old documents

### DIFF
--- a/now_executors/ann_lite_now/executor.py
+++ b/now_executors/ann_lite_now/executor.py
@@ -1,3 +1,4 @@
+from copy import deepcopy
 from sys import maxsize
 from typing import Dict, List, Optional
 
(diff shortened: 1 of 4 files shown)


now_executors/ann_lite_now/executor.py
```diff
@@ -151,26 +152,17 @@ def update(
...
-        traversal_paths = parameters.get('traversal_paths', self.index_traversal_paths)
-        flat_docs = docs[traversal_paths]
-        if len(flat_docs) == 0:
-            return
-        self.delete_inmemory_docs(flat_docs)
-        self._index.delete(flat_docs)
+        Delete endpoint to delete document/documents from the index.
+        Filter conditions can be passed to select documents for deletion.
+        """
+        filter = parameters.get("filter", {})
```
joschkabraun, line 160: don't we also need `traversal_paths`?

joschkabraun (approved):
LGTM after comment :)

now_executors/ann_lite_now/executor.py
florian-hoenicke: thanks for pointing that out. There is an inconsistency between the other endpoints. But this gives us the freedom filtering by any conditions

PR #477, opened 2022-09-08
feat: set jina log level of gateway to debug

joschkabraun:
Set `JINA_LOG_LEVEL=DEBUG` in gateway in order to get logs there.


now/apps/text_to_text_and_image/flow.yml
```diff
@@ -1,4 +1,9 @@
 jtype: Flow
+with:
+  name: nowapi
+  monitoring: true
```
makram93, line 4: Should we set `port_expose` and `prefetch` params like in other flows?

joschkabraun: I think prefetch is going to be app individual. But I think `port_expose` is good. Thanks!

now/apps/base/flow.yml
```diff
@@ -1,6 +1,10 @@
 jtype: Flow
 version: '1'
 with:
+  name: nowapi
```
jupyterjazz, line 4: is this flow used somewhere?

jupyterjazz (approved):
LGTM!

now/apps/base/flow.yml
makram93: Though, It is used in the base/app.py inside `set_flow_yaml()` but every JinaApp should override this function so in effect this flow is not used. It just serves as a reference template.

makram93 (approved):

now/apps/base/flow.yml
joschkabraun: Yes exactly!

PR #483, opened 2022-09-08
fix: update cli parser

makram93:
Fix the cli
- and also fixes the bug introduced in refactoring

### DIFF
--- a/now/finetuning/run_finetuning.py
+++ b/now/finetuning/run_finetuning.py
@@ -3,6 +3,7 @@ import os
 import random
 import string
 import sys
+from copy import deepcopy
 from time import sleep
 from typing import Dict, Tuple
 
@@ -228,7 +229,7 @@ def _maybe_add_embeddings(
     result = call_index(
         client=client,
         dataset=documents_without_embedding,
-        parameters={'user_input': user_input.__dict__},
+        parameters={'user_input': deepcopy(user_input.__dict__)},
         return_results=True,
     )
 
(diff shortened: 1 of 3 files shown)


now/finetuning/run_finetuning.py
```diff
@@ -228,7 +229,7 @@ def _maybe_add_embeddings(
     result = call_index(
         client=client,
         dataset=documents_without_embedding,
-        parameters={'user_input': user_input.__dict__},
+        parameters={'user_input': deepcopy(user_input.__dict__)},
```
azayz, line 232: why is deep copy necessary ?

makram93: Because when I pop `app_instance` from the parameters, it is removed from the `user_input` as well. Hence making a deepcopy

now/run_backend.py
```diff
@@ -71,8 +77,10 @@ def call_index(
     return_results: Optional[bool] = False,
 ):
     request_size = estimate_request_size(dataset)
-    # Remove app_instance from parameters
-    parameters['user_input'].pop('app_instance')
+
+    # Deep copy of the user_input without app_instance from parameters
+    if 'app_instance' in parameters['user_input']:
+        parameters['user_input'].pop('app_instance')
```
jupyterjazz, line 83:
```suggestion
    parameters['user_input'].pop('app_instance', None)
```

jupyterjazz (commented):
what was the bug in the cli? i.e. what was fixed by the changes introduced in the `parser.py`?

now/run_backend.py
makram93: > what was the bug in the cli? i.e. what was fixed by the changes introduced in the parser.py?
The default app options were not populated in the cli parser. I was adding app specific but forgot to add the default ones.

jupyterjazz (approved):
LGTM!

PR #489, opened 2022-09-09
Fix local finetuning

makram93:
When a finetune-enabled dataset is used to deploy locally, it fails to deploy pods for embedding creation. This fixes the namespace issue and also filters the data based on the output modality


now_executors/NOWPreprocessor/executor.py
```diff
@@ -141,9 +141,10 @@ def index(
         :return: preprocessed documents which are ready to be encoded and indexed
         """
         if docs is None or len(docs) == 0:
-            return
+            return DocumentArray()
+        is_indexing = parameters.get('is_indexing', True)
         self._set_user_input(parameters=parameters)
-        return self._preprocess_maybe_cloud_download(docs=docs, is_indexing=True)
+        return self._preprocess_maybe_cloud_download(docs=docs, is_indexing=is_indexing)
```
joschkabraun, line 147: I would suggest write a new endpoint & function which is called `encode` which concatenates the output of `index` and `search`. Introducing this `is_indexing` parameter is very confusing for someone calling the `index` endpoint.

makram93: I just realized that we do not need it as a param at all or to write a different function for `/encode` as the existing logic works fine. I wanted to parameterize it but I will revert this change.

joschkabraun: What changed?
By latest changes in `main` we skip the this encoding step as we reuse precomputed `embedding` again.

jupyterjazz (approved):
LGTM!

joschkabraun (changes requested):
Using this will remove any `input_modality` documents from the demo datasets and they will be disregarded in the finetuning. For the `text-to-image` app: This leads to the finetuned MLP to only see input samples of the fom (tensor, 0) (from images) but never (0, tensor) (from text). Therefore, the MLP which maps the text to the final embedding will be left with the initialized weights. I would suggest to
- implement `encode` endpoint in `NOWPreprocessor` to combine results form `index` and `search` endpoint

makram93:
> Using this will remove any `input_modality` documents from the demo datasets and they will be disregarded in the finetuning. For the `text-to-image` app: This leads to the finetuned MLP to only see input samples of the fom (tensor, 0) (from images) but never (0, tensor) (from text). Therefore, the MLP which maps the text to the final embedding will be left with the initialized weights. I would suggest to
> 
> * implement `encode` endpoint in `NOWPreprocessor` to combine results form `index` and `search` endpoint

Thanks for the suggestion, it makes the code more readable. I have updated this PR accordingly

joschkabraun (approved):
Nice fix! LGTM :rocket:

PR #490, opened 2022-09-09
feat: preprocessor gets tags

florian-hoenicke:


now_executors/NOWPreprocessor/executor.py
```diff
@@ -82,7 +76,24 @@ def convert_fn(d: Document) -> Document:
...
+                region_name=self.user_input.aws_region_name,
+            )
+            bucket = session.resource('s3').Bucket(d.uri.split('/')[2])
+            d.uri = download(bucket, d.uri)
+            if 'tag_uri' in d.tags:
+                d.tags['tag_uri'] = download(bucket, d.tags['tag_uri'])
+                with open(d.tags['tag_uri'], 'r') as fp:
+                    tags = json.load(fp)
+                    tags = flatten_dict(tags)
+                    d.tags.update(tags)
```
joschkabraun, line 96: I would delete `tag_uri` from `d.tags`

joschkabraun (approved):
LGTM after small comment! Curious to see how much faster tag loading becomes.

now_executors/NOWPreprocessor/executor.py
florian-hoenicke: Thanks, makes sense

PR #494, opened 2022-09-11
fix: add api keys

florian-hoenicke:
Since we are using replicas, the propagation of the API key can not be guaranteed. We increase the likelihood that all replicas will get a key by sending more requests.
Assume, we have 10 replicas and 100 requests of type `add_api_key`.
We assume that the probability of what replica is used is evenly distributed.
It means that the likelihood that a replica has an api_key is:
`1-(9/10)^100` ~ 1


joschkabraun (approved):
LGTM

PR #495, opened 2022-09-11
refactor: remove security executor deployment

florian-hoenicke:
the other executors inherit from the security executor, so it is not needed alone

### DIFF
--- a/now/deployment/flow.py
+++ b/now/deployment/flow.py
@@ -4,7 +4,7 @@ import pathlib
 import tempfile
 from os.path import expanduser as user
 from time import sleep
-from typing import Dict, List
+from typing import Dict
 
 from jina import Flow
 from jina.clients import Client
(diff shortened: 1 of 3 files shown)


makram93:
Why are we removing security exectuor?

now/deployment/flow.py
```diff
@@ -104,44 +104,15 @@ def deploy_k8s(f, ns, tmpdir, kubectl_path):
...
-
-
 @time_profiler
 def deploy_flow(
     deployment_type: str,
     flow_yaml: str,
     ns: str,
     env_dict: Dict,
     kubectl_path: str,
-    secured: bool = False,
```
makram93, line 137: Where or how are these informations passed on to the flow?

florian-hoenicke: good observation. I think that could be the reason why the local test is not passing. I will investigate

florian-hoenicke: In case, there are no admin emails, user emails and api_keys, the flow knows it is not secured

PR #496, opened 2022-09-11
fix: no email ignore error

florian-hoenicke:
At the moment the program execution just crashes in case the email is not contained in the hubble response. Instead, we should print an error message and continue the execution

### DIFF
--- a/now/utils.py
+++ b/now/utils.py
@@ -361,7 +361,13 @@ def get_info_hubble(user_input):
         user_token = config_val['auth_token']
     client = hubble.Client(token=user_token, max_retries=None, jsonify=True)
     response = client.get_user_info()
-    user_input.admin_emails = [response['data']['email']]
+    user_input.admin_emails = (
+        [response['data']['email']] if 'email' in response['data'] else []
+    )
+    if not user_input.admin_emails:
+        print(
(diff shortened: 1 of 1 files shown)


now/utils.py
```diff
@@ -361,7 +361,13 @@ def get_info_hubble(user_input):
         user_token = config_val['auth_token']
     client = hubble.Client(token=user_token, max_retries=None, jsonify=True)
     response = client.get_user_info()
-    user_input.admin_emails = [response['data']['email']]
+    user_input.admin_emails = (
+        [response['data']['email']] if 'email' in response['data'] else []
+    )
+    if not user_input.admin_emails:
```
makram93, line 367: - Email is only needed when the user wants to secure the flow. But at this stage, this dialog has not yet been asked. So, should we raise the flag for unverified email when he selects yes for the secured remote flow option or here? I am okay with it being at either place.
- Shouldn't we raise the error and stop the execution, to enforce account verification?

florian-hoenicke: I think you are right. Tha problem is just that I did not receive a verification email from hubble. So I had to make this workaround.

Furthermore, the responsibility of checking the verification can not be part of the now package since it is not adding a security mechanism. Everyone could just deploy flows without verification on jcloud. So I think It will be jclouds responsibility.

PR #502, opened 2022-09-12
test: add tests for nest

jupyterjazz:
Covers testing:

- Preprocessing function
- Option for reading the task configuration
- Finetuning settings construction


now/apps/text_to_text_and_image/app.py
```diff
@@ -147,6 +147,6 @@ def _model_name(encoder_type: Optional[str] = None) -> str:
     @staticmethod
     def _loss_function(encoder_type: Optional[str] = None) -> str:
         """Loss function used during fine-tuning."""
-        if encoder_type == 'text_to_image':
+        if encoder_type == 'text-to-image':
```
florian-hoenicke, line 150: is image-to-text needed here as well?

tests/unit/apps/test_preprocessing.py
```diff
@@ -83,3 +86,28 @@ def test_music_preprocessing(is_indexing):
     assert len(da) == 1
     assert len(da[0].chunks) == 0
     assert da[0].blob != b''
+
+
+@pytest.mark.parametrize('is_indexing', [False, True])
+def test_nest_preprocessing(is_indexing, get_nest_config_path):
```
florian-hoenicke, line 92: I would recommend removing the name nest from our codebase and instead being explicit about the input and output modalities used. This will also make it easier for new people getting onboarded and helps us us with the integration.

now/apps/text_to_text_and_image/app.py
jupyterjazz: not for this app

tests/unit/apps/test_options.py
```diff
@@ -0,0 +1,18 @@
+from now.apps.text_to_text_and_image.app import TextToTextAndImage
+from now.dialog import configure_option
+from now.now_dataclasses import UserInput
+
+
+def test_text_to_text_and_image_options(mocker, get_nest_config_path):
```
violenil, line 6: what is this `get_nest_config_path`? And should we think about renaming to remove the word `nest` here?

violenil (approved):
Looks good! Just posed one question for understanding + comment on naming.

PR #505, opened 2022-09-13
fix: cd

azayz:
fixes CD fail


joschkabraun (approved):
After CI is green, LGTM! Nice quick fix!

Issue #352, opened 2022-08-16
feat: make playground login look more professional


florian-hoenicke:
It is great that we can login into the playground.
In the next iteration, let's focus on making the login experience better by having a more professional looking button

This is how it looks right now:
![Image](https://user-images.githubusercontent.com/11627845/184920940-b8391016-fcb0-4cae-8a18-45fefdc4ea68.png)

Suggestions to beauty it up:
- use Jina NOW color scheme
- center button
- button design in typical streamlit design

florian-hoenicke:
as discussed in the refinement, @fogx will ask the design team  for input

fogx:
can't we just redirect to the login page directly, instead of showing this intermediate page? Then we don't need any design at all.

florian-hoenicke:
You are awesome. Yes, we can

PR #511, opened 2022-09-15
fix: logout issue

azayz:
fixes https://github.com/jina-ai/now/issues/352 and https://github.com/jina-ai/now.private/issues/107


florian-hoenicke (approved):
Thanks for showing me how this works. I think it is very nice that we smoothened the user interaction.

PR #521, opened 2022-09-16
Fix: playground tags secured

azayz:
Reverts jina-ai/now#520

### DIFF
--- a/deployment/playground/playground.py
+++ b/deployment/playground/playground.py
@@ -2,6 +2,7 @@ import base64
 import io
 import json
 import os
+from collections import OrderedDict
 from copy import deepcopy
 from urllib.parse import quote, unquote
 from urllib.request import urlopen
@@ -13,6 +14,7 @@ import streamlit as st
 import streamlit.components.v1 as components
 from better_profanity import profanity
 from docarray import Document, DocumentArray
+from jina import Client
 from src.constants import (
     BUTTONS,
     JWT_COOKIE,
@@ -97,7 +99,6 @@ def deploy_streamlit():
 
     # Retrieve query params
     params = get_query_params()
-
     redirect_to = render_auth_components(params)
 
     _, mid, _ = st.columns([0.8, 1, 1])
@@ -124,8 +125,25 @@ def deploy_streamlit():
             from better_profanity import profanity
 
             profanity.load_censor_words()
-
         setup_design()
+
+        filters = None
+        if params.host:
+            client = Client(host=params.host)
+            if params.secured == 'true':
+                response = client.post(
+                    on='/tags', parameters={'jwt': {'token': st.session_state.jwt_val}}
+                )
+            else:
+                response = client.post(on='/tags')
+            filters = OrderedDict(response[0].tags['tags'])
+
+        filter_selection = {}
+        if filters:
+            st.sidebar.title("Filters")
+            for tag, values in filters.items():
+                values.insert(0, 'All')
+                filter_selection[tag] = st.sidebar.selectbox(tag, values)
 
         if params.input_modality == 'image':
             media_type = st.radio(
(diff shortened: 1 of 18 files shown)


deployment/playground/playground.py
```diff
@@ -124,9 +125,26 @@ def deploy_streamlit():
             from better_profanity import profanity
 
             profanity.load_censor_words()
-
         setup_design()
 
+        filters = None
```
makram93, line 130: Please use st.session_sate to store the filters since, we would want to load all the tags only once when the page loads.

deployment/playground/playground.py
```diff
@@ -124,9 +125,26 @@ def deploy_streamlit():
...
+        if params.host:
+            client = Client(host=params.host)
+            if params.secured.lower() == 'true':
+                response = client.post(
+                    on='/tags', parameters={'jwt': {'token': st.session_state.jwt_val}}
+                )
+            else:
+                response = client.post(on='/tags')
+            filters = OrderedDict(response[0].tags['tags'])
```
makram93, line 139: Either wrap it in try-catch or check first if the response has docs and `tags` and only then access it

deployment/playground/playground.py
```diff
@@ -326,19 +344,25 @@ def render_image(da_img):
             with txt:
                 if st.button('Search', key=doc.id, on_click=clear_match):
                     st.session_state.matches = search_by_image(
-                        document=doc, jwt=st.session_state.jwt_val
+                        document=doc,
```
makram93, line 347: Please make sure that you are passing filter_selection as params wherever `search_by_image` is called(there should be 3 places) and similarly for other search functions. Or make this param optional if it is not needed to be called at other places

makram93 (commented):
Some minor comments

now_common/utils.py
```diff
@@ -167,10 +173,73 @@ def get_indexer_config(num_indexed_samples: int) -> Dict:
...
-    for doc in da:
-        for tag, _ in doc.tags.items():
-            tags.add((tag, str(tag.__class__.__name__)))
+    for tag, _ in d.tags.items():
+        tags.add((tag, str(tag.__class__.__name__)))
     final_tags = [list(tag) for tag in tags]
     return final_tags
+
+
+def download(bucket, uri, tmpdir):
```
joschkabraun, line 190: the name is too ambiguous and the function doesn't carry its context anymore. I would
- name the function to something more descriptive
- add document on what it does and what `bucket`, `uri`, and `tmpdir` are

now_common/utils.py
```diff
@@ -167,10 +173,73 @@ def get_indexer_config(num_indexed_samples: int) -> Dict:
...
+        base64.b64encode(bytes(path_s3, "utf-8")).decode("utf-8"),
+    )
+    bucket.download_file(
+        path_s3,
+        path_local,
+    )
+    return path_local
+
+
+def convert_fn(d: Document, user_input, tmpdir) -> Document:
```
joschkabraun, line 203: Naming of function is very ambiguous. I would move it into `_maybe_download_from_s3` as it's only used there.

now_common/utils.py
```diff
@@ -167,10 +173,73 @@ def get_indexer_config(num_indexed_samples: int) -> Dict:
...
+        d.tags['tag_uri'] = download(bucket, d.tags['tag_uri'], tmpdir)
+        with open(d.tags['tag_uri'], 'r') as fp:
+            tags = json.load(fp)
+            tags = flatten_dict(tags)
+            d.tags.update(tags)
+        del d.tags['tag_uri']
+    return d
+
+
+def _maybe_download_from_s3(
```
joschkabraun, line 223: I would move the `_maybe_download_from_s3` from `NOWPreprocessor` here (or into `now/data_loading` or `now/utils`) and then reuse it here

joschkabraun (changes requested):

now/utils.py
```diff
@@ -421,3 +426,69 @@ def _prompt_value(
         qs['choices'] = choices
         qs['type'] = 'list'
     return maybe_prompt_user(qs, name, **kwargs)
+
+
+def _maybe_download_from_s3(
+    docs: DocumentArray, tmpdir: tempfile.TemporaryDirectory, user_input
```
joschkabraun, line 432: We are missing using `max_workers` as they are specified in `NOWPreprocessor`.

joschkabraun (approved):
After CI is green, LGTM! :rocket

joschkabraun (approved):
LGTM!

PR #523, opened 2022-09-16
feat: introduce Elasticsearch options to CLI

jupyterjazz:
- add dialog options for Elasticsearch
- implement and test ES data loading
- generate task configuration based on the user input 

The feature is still disabled because it only works for text to text+image app.
I'll make it work for all the apps in the next PR.


now/apps/text_to_text_and_image/app.py
```diff
@@ -49,23 +49,47 @@
...
-            prompt_type='input',
-            post_func=lambda user_input, **kwargs: self._read_task_config(
-                kwargs['task_config'], user_input
-            ),
-        )
-        return [task_config_option]
-
     @staticmethod
-    def _read_task_config(task_config_path: str, user_input: UserInput):
-        with open(task_config_path) as f:
```
joschkabraun, line 66: does that `task_config.json` work for any task or does it need to be modified?

joschkabraun (approved):
LGTM! :rocket:

now/apps/text_to_text_and_image/app.py
jupyterjazz: it needs to be modified by adding field names, that's the purpose of `_create_task_config()`

PR #525, opened 2022-09-19
fix: playground url inside try catch

makram93:
When loading samples, if the link is not correct then it will throw HTTP error. Instead of that, catch the error silently and do not show the samples.

### DIFF
--- a/deployment/playground/playground.py
+++ b/deployment/playground/playground.py
@@ -261,8 +261,8 @@ def load_example_queries(data, output_modality):
                 output_modality_dir = 'text'
                 data_dir = root_data_dir + output_modality_dir + '/'
                 da_txt = load_data(data_dir + data + f'.txt10-{docarray_version}.bin')
-        except HTTPError:
-            pass
+        except HTTPError as exc:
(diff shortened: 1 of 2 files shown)


deployment/playground/playground.py
```diff
@@ -248,19 +245,24 @@ def _do_logout():
...
+                da_img, da_txt = load_data(
+                    data_dir + data + f'.img10-{docarray_version}.bin'
+                ), load_data(data_dir + data + f'.txt10-{docarray_version}.bin')
+            elif output_modality == 'text':
+                # for now deactivated sample images for text
+                output_modality_dir = 'text'
+                data_dir = root_data_dir + output_modality_dir + '/'
+                da_txt = load_data(data_dir + data + f'.txt10-{docarray_version}.bin')
+        except HTTPError:
+            pass
```
azayz, line 265: maybe add logs here, it will be easier to debug, pass will make this extremely hard to figure it out

azayz (changes requested):

azayz (approved):

PR #528, opened 2022-09-20
feat: init commit

florian-hoenicke:


now_executors/NOWQdrantIndexer/executor.py
```diff
@@ -0,0 +1,212 @@
...
+
+from docarray import DocumentArray
+from jina import Executor, requests
+from jina.logging.logger import JinaLogger
+
+
+class QdrantIndexer3(Executor):
+    """QdrantIndexer3 indexes Documents into a Qdrant server using DocumentArray  with `storage='qdrant'`"""
+
+    def __init__(
```
JoanFM, line 14: why is this needed and not being used the ones from `jinahub`?

florian-hoenicke: With this, we can host the qdrant indexer on wolf without setting up a qdrant cluster

PR #540, opened 2022-09-26
fix: update list endpoint and test cases

makram93:
List endpoint currently returns the flat docs which is wrong when we have chunks of documents. This PR solves it by removing the duplicates and only returning single docs for all chunks of the same parent.

### DIFF
--- a/now_executors/NOWAnnLiteIndexer/executor.py
+++ b/now_executors/NOWAnnLiteIndexer/executor.py
@@ -91,7 +91,7 @@ class NOWAnnLiteIndexer(Executor):
         self.da = DocumentArray()
         for cell_id in range(self._index.n_cells):
             for docs in self._index.documents_generator(cell_id, batch_size=10240):
-                self.extend_inmemory_docs(docs)
+                self.extend_inmemory_docs_and_tags(docs)
 
         self.da = DocumentArray(sorted([d for d in self.da], key=lambda x: x.id))
 
@@ -104,7 +104,9 @@ class NOWAnnLiteIndexer(Executor):
     def extend_inmemory_docs_and_tags(self, docs):
         """Extend the in-memory DocumentArray with new documents"""
         for d in docs:
-            self.da.append(Document(id=d.id, uri=d.uri, tags=d.tags))
+            self.da.append(
+                Document(id=d.id, uri=d.uri, tags=d.tags, parent_id=d.parent_id)
+            )
             self.doc_id_tags[d.id] = d.tags
         self.save_tags()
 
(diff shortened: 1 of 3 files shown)


tests/now_executors/NOWAnnLiteIndexer/test_executor.py
```diff
@@ -79,28 +80,38 @@ def test_list(tmpdir, offset, limit):
...
+                'chunks_size': 2,
+            }
+            if has_chunk
+            else {
+                'offset': offset,
+                'limit': limit,
             }
-            if offset is not None
-            else {}
         )
```
joschkabraun, line 95: I think we should change test to only add `offset` and `limit` in the case they are actually given / not `None`

tests/now_executors/NOWAnnLiteIndexer/test_executor.py
```diff
@@ -79,28 +80,38 @@ def test_list(tmpdir, offset, limit):
...
         if l > 0:
-            assert list_res[0].id == str(offset) if offset is not None else '0'
-            assert list_res[0].uri == 'my-parent-uri'
-            assert len(list_res[0].chunks) == 0
-            assert list_res[0].embedding is None
-            assert list_res[0].text == ''
-            assert list_res[0].tags == {'parent_tag': 'value'}
+            if has_chunk:
+                assert len(list_res[0].chunks) == 0
+                assert len(set([d.id for d in list_res])) == l
```
joschkabraun, line 107: I think `has_chunk` should also test for `tags`. So, we need to change `gen_docs` function

joschkabraun (commented):
Nice fix!
I think we should change the behavior of the parameter as I outlined in the comments. Let me know what you think

now_executors/NOWAnnLiteIndexer/executor.py
```diff
@@ -220,10 +222,34 @@ def list(self, parameters: dict = {}, **kwargs):
         - offset (int): number of documents to skip
         - limit (int): number of retrieved documents
         """
-
-        limit = int(parameters.get('limit', maxsize))
-        offset = int(parameters.get('offset', 0))
-        return self.da[offset : offset + limit]
+        limit = (
+            int(parameters.get('limit'))
+            if parameters.get('limit') is not None
```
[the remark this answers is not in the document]
makram93: I agree that `None` may not be the case but somehow the params passed is received as `float` inside the `/list` endpoint hence the type casting. Will update the test cases accordingly

tests/now_executors/NOWAnnLiteIndexer/test_executor.py
makram93: Thanks, have added the additional test check

joschkabraun (approved):
LGTM!

PR #541, opened 2022-09-26
fix: bug unpickling error

azayz:


now/data_loading/utils.py
```diff
@@ -22,11 +24,17 @@ def _fetch_da_from_url(
     if not os.path.exists(data_path):
         download(url, data_path)
 
-    with yaspin_extended(
-        sigmap=sigmap, text="Extracting dataset from DocArray", color="green"
-    ) as spinner:
+    try:
```
florian-hoenicke, line 27: Maybe we could do a loop two times over everything.
to avoid code duplication. But not sure. I would leave it to you. I think it also works as it is. Just a thought

now/data_loading/utils.py
```diff
@@ -22,11 +24,17 @@ def _fetch_da_from_url(
...
+        with yaspin_extended(
```
florian-hoenicke, line 28: yaspin can be removed

now/data_loading/utils.py
florian-hoenicke: Or better leave it as it is

florian-hoenicke (approved):

now/data_loading/utils.py
```diff
@@ -22,11 +23,13 @@ def _fetch_da_from_url(
     if not os.path.exists(data_path):
         download(url, data_path)
 
-    with yaspin_extended(
```
makram93, line 25: Why do we remove this block? Try catch can still be put inside it. no?

azayz: the argument was the data extraction was too fast to even see the progress bar so we suggested to remove it

makram93 (approved):
LGTM!

joschkabraun (approved):
LGTM!

Issue #542, opened 2022-09-27
bug: fail to import update_api_key in sync_job


makram93:
When calling `update_api_keys` inside the sync job, it fails to load the function as the `update_api_keys` module tries to import the function which is defined inside the test folder... and this test folder is not packaged with Jina NOW hence python is unable to find it.

![Image](https://user-images.githubusercontent.com/6537525/192476993-3b77d9fb-dedc-4778-9576-f0deabb47c56.png)

PR #543, opened 2022-09-27
Refactor remove jc secret

makram93:
This PR refactors the jc_secret and removes all its occurrences. In the integration test, we still need to keep track of the deployed flow for post-cleanup. We cannot list all flows and deploy them as they may be in use in other parallel tests. 

This PR also solves #542

### DIFF
--- a/now/run_all_k8s.py
+++ b/now/run_all_k8s.py
@@ -1,13 +1,11 @@
-import json
 import os
-from os.path import expanduser as user
 
 import cowsay
 from now_common.options import _get_context_names
 
 from now import run_backend, run_bff_playground
 from now.cloud_manager import setup_cluster
-from now.constants import DOCKER_BFF_PLAYGROUND_TAG, JC_SECRET, SURVEY_LINK
+from now.constants import DOCKER_BFF_PLAYGROUND_TAG, SURVEY_LINK
 from now.deployment.deployment import cmd, list_all_wolf, status_wolf, terminate_wolf
 from now.dialog import configure_app, configure_user_input
 from now.log import yaspin_extended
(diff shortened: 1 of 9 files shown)


now/run_all_k8s.py
```diff
@@ -57,31 +50,32 @@
...
~                sigmap=sigmap,
~                text=f"Remove namespace nowapi NOW from {cluster}",
~                color="green",
~            ) as spinner:
~                cmd(f'{kwargs["kubectl_path"]} delete ns nowapi')
                 spinner.ok('💀')
             cowsay.cow(f'nowapi namespace removed from {cluster}')
     elif 'wolf.jina.ai' in cluster:
-        flow_details = get_remote_flow_details()
-        flow_id = flow_details['flow_id']
```
joschkabraun, line 78: Don't we also need to call `cleanup` in any case and not only for `remote` deployment? E.g. databases need to be closed etc. when deleting a local deployment.

tests/integration/test_end_to_end.py
```diff
@@ -173,18 +186,24 @@ def assert_deployment_queries(
     kwargs,
     output_modality,
     test_search_image,
+    response,
 ):
     url = f'http://localhost:30090/api/v1'
+    host = None
+    if 'host' in response:
+        host = response['host']
```
joschkabraun, line 194: can shorten to `host = response.get('host', None)`

makram93: Ah yes, thanks. Will update it

now/run_all_k8s.py
joschkabraun: I think it is a good solution to fetch the app from the `cluster` one wants to remove. Is it possible to also get the app from the environment variables from the namespace in kubectl? 

I agree we don't use the `cleanup` function yet, I still believe that we need to keep it consistent to avoid forgetting about it in the future. I can imagine in the `cleanup` function that a report can be created of all the queries and their performance etc. This would also be interesting in the case of local deployment.

makram93: >Is it possible to also get the app from the environment variables from the namespace in kubectl?
I am not sure. I remember we had a similar request to store metadata in the flow itself so that we can query to extract information. However, it is currently not being done. Anyways, I will revert the cleanup to the original code and we can create another ticket to discuss this

florian-hoenicke (approved):
LGTM

Issue #539, opened 2022-09-26
bug: initialize package too often


florian-hoenicke:
I was running the text-to-text+image app. The following issues occurred:
- [x] multiple initializations of our package
![Image](https://user-images.githubusercontent.com/11627845/192268192-46c93e8b-a301-4838-8028-ff984703bcf3.png)
- [ ] if finetuner API is down, it should be skipped instead of aborting the process.  The user should just be informed that fine-tuning is skipped.
```
🔧 Perform finetuning!
💪 fine-tuning:
Traceback (most recent call last):
  File "/Users/florianhonicke/jina/now/now/cli/__init__.py", line 147, in <module>
    cli()
  File "/Users/florianhonicke/jina/now/now/cli/__init__.py", line 143, in cli
    return run_k8s(os_type=os_type, arch=arch, **args)
  File "/Users/florianhonicke/jina/now/now/run_all_k8s.py", line 168, in run_k8s
    **kwargs,
  File "/Users/florianhonicke/jina/now/now/run_all_k8s.py", line 111, in start_now
    ) = run_backend.run(app_instance, user_input, kubectl_path=kwargs['kubectl_path'])
  File "/Users/florianhonicke/jina/now/now/log/log.py", line 58, in profiled_fun
    result = fun(*args, **kwargs)
  File "/Users/florianhonicke/jina/now/now/run_backend.py", line 33, in run
    dataset=dataset, user_input=user_input, kubectl_path=kubectl_path
  File "/Users/florianhonicke/jina/now/now/apps/text_to_text_and_image/app.py", line 125, in setup
    kubectl_path=kubectl_path,
  File "/Users/florianhonicke/jina/now/now/log/log.py", line 58, in profiled_fun
    result = fun(*args, **kwargs)
  File "/Users/florianhonicke/jina/now/now/finetuning/run_finetuning.py", line 78, in finetune
    return _finetune_model(finetune_ds, finetune_settings)
  File "/Users/florianhonicke/jina/now/now/log/log.py", line 58, in profiled_fun
    result = fun(*args, **kwargs)
  File "/Users/florianhonicke/jina/now/now/finetuning/run_finetuning.py", line 109, in _finetune_model
    finetuner.login()
  File "/Users/florianhonicke/jina/now/env/lib/python3.7/site-packages/finetuner/__init__.py", line 32, in login
    ft.login()
[...]

PR #545, opened 2022-09-27
fix: multiple initialisation prints

jupyterjazz:
Fixes the first part of https://github.com/jina-ai/now/issues/539

Multiprocessing imports the `__init__` module for each process, resulting in printing `Initialising Jina NOW` several times. It's fixed by moving print statement into the `cli()` function.

### DIFF
--- a/now/cli/__init__.py
+++ b/now/cli/__init__.py
@@ -11,8 +11,6 @@ from now import __version__ as version
 from now.deployment.deployment import cmd
 
 warnings.filterwarnings("ignore")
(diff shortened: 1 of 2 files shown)


makram93 (commented):
The problem with this is that it takes time before it shows the `initializing...` message, because when starting the app it imports a lot of things and takes time before it enters the `cli()`. That's the reason we moved it outside... Do you notice this latency with the new changes or is the effect the same?

joschkabraun (approved):
LGTM

jupyterjazz:
@makram93 

Nothing really changed in terms of speed. Shows the text immediately. 

Screenshot 2022-09-27 at 16 56 35

makram93 (approved):

Issue #482, opened 2022-09-08
refactor: embed demo datasets


joschkabraun:
In order to cost save & quickly show demos host (some) demo datasets permanently on a static URL in docs.

(we're doing this to have example deployments on the storefront)
UAC 

- [ ] embed demo datasets in a section in the docs

- [x] #556

  - when someone chooses demo data, then they will be pointed towards the static URL in the docs for that particular demo
- [x] #569
  - need to add to `flow-clip.yml`, `flow-sbert.yml` and any other relevant `flow.yaml` the following line under jcloud: `custom_dns: now-example-${{ ENV.APP }}-${{ ENV.DEMO_DATA }}.dev.jina.ai`
    - cf. more:
        
        [https://jinaai.slack.com/archives/C03434J5BND/p1662473489965669?thread_ts=1662119465.049149&cid=C03434J5BND](https://jinaai.slack.com/archives/C03434J5BND/p1662473489965669?thread_ts=1662119465.049149&cid=C03434J5BND)
  - also need to add `CNAME` record to Route53 for `dev.jina.ai`
 - [x] #570

fogx:
how much does it cost to keep a flow alive for one week? and how much does it cost to deploy a new flow?

PR #548, opened 2022-09-27
refactor: dialog

joschkabraun:
Closes https://github.com/jina-ai/now/issues/556 of https://github.com/jina-ai/now/issues/482 and additionally:

- [x] removes redundancy in `UserInput`
- [x] changes every data source (except Elasticsearch) to use `data` as the origin of the data


now/client/__init__.py
```diff
@@ -67,7 +67,7 @@ def send_request(self, endpoint: str, **kwargs):
         jina_response = jina_client_post(
             app_request,
             endpoint,
-            jina_request.data,
```
azayz, line 70: why is this changing ?

now_common/options.py
```diff
@@ -328,11 +339,6 @@ def _construct_app(app_name: str):
     )()
 
 
-@time_profiler
-def _check_requirements(user_input: UserInput, **kwargs) -> None:
-    user_input.app_instance.run_checks(user_input)
```
azayz, line 333: calling run_checks directly wouldn't use the decorator time profiler is this intended ?

azayz (commented):
Mostly LGTM besides some few things I would like to understand more

now/client/__init__.py
joschkabraun: Thanks for catching this! Got accidentally renamed.

now_common/options.py
joschkabraun: Yes, but I don't think profiling it is actually necessary. If we want that, we can always move it to the base app. What do you think?

azayz: do you know the reason why we used time profiling in the first place ? If you believe that it's unnecessary then feel free to remove it as it's nit extremely important anyway

now_common/options.py
```diff
@@ -62,23 +66,11 @@
     description='What sort of search engine would you like to build?',
 )
 
-DATA = DialogOptions(
```
azayz, line 65: if these changes update the cli options and dialog, can you also update the docs ?

makram93 (commented):
Nice refactoring. I like it overall. My only concern is about the cli parser. @florian-hoenicke and I had a discussion long back where he suggested to have only `--data` args for cli and parse it based on the value of it. It would be nice to still have that as the only argument exposed and add a parser function that will set the `user_input` accordingly. If we wish to continue with the current changes then it will have a breaking changes in the`now_api`, docs, jina.ai website, etc
cc @florian-hoenicke

now_common/options.py
joschkabraun: no

now_common/options.py
joschkabraun: Good catch. Thanks!

florian-hoenicke:
> Nice refactoring. I like it overall. My only concern is about the cli parser. @florian-hoenicke and I had a discussion long back where he suggested to have only `--data` args for cli and parse it based on the value of it. It would be nice to still have that as the only argument exposed and add a parser function that will set the `user_input` accordingly. If we wish to continue with the current changes then it will have a breaking changes in the`now_api`, docs, jina.ai website, etc cc @florian-hoenicke

@makram93 I remember our discussion. Back then I was arguing for a single --data attribute. It fits our story: just provide your data and we provide an API. 
Meanwhile, I have not a strong opinion anymore. The main usage of our product will be via setup assistance on cli and via jina cloud.
`jina now --data --...` is not something we promote at the moment. Therefore I'm happy with making our code more readable and removing redundant attributes.

PR #553, opened 2022-09-28
fix: convert user emails to list

makram93:
comma separated user_emails to list of user_emails

### DIFF
--- a/now_common/options.py
+++ b/now_common/options.py
@@ -312,7 +312,14 @@ USER_EMAILS = DialogOptions(
     prompt_type='input',
     depends_on=ADDITIONAL_USERS,
     conditional_check=lambda user_inp: user_inp.additional_user,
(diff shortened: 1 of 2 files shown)


now_common/options.py
```diff
@@ -312,9 +312,18 @@ def _parse_image_fields(user_input: UserInput):
     prompt_type='input',
     depends_on=ADDITIONAL_USERS,
     conditional_check=lambda user_inp: user_inp.additional_user,
+    post_func=lambda user_input, **kwargs: _add_additional_users(user_input, **kwargs),
```
azayz, line 315: LGTM! just one question, how was additional emails being handled before ?

makram93: When I was refactoring the user dialog, I forgot to add this post-processing of emails, which should split the string to list

PR #560, opened 2022-10-03
fix: music to music app

azayz:


tests/integration/test_end_to_end.py
```diff
@@ -111,6 +126,20 @@ def test_token_exists():
             DemoDatasets.TUMBLR_GIFS_10K,
             'local',
         ),
+        (
+            Apps.MUSIC_TO_MUSIC,
+            Modalities.MUSIC,
+            Modalities.MUSIC,
+            DemoDatasets.MUSIC_GENRES_ROCK,
+            'local',
```
makram93, line 134: I would recommend adding only a remote test case since it could increase CI time..

now/data_loading/data_loading.py
```diff
@@ -55,6 +55,13 @@ def load_data(app: JinaNOWApp, user_input: UserInput) -> DocumentArray:
             da = da[:300]
         else:
             da = da[:300]
+    if (
+        user_input.data == DemoDatasets.MUSIC_GENRES_MIX
+        or user_input.data == DemoDatasets.MUSIC_GENRES_ROCK
+    ):
+        for doc in da:
+            if 'genre_tags' in doc.tags and isinstance(doc.tags['genre_tags'], list):
+                doc.tags['genre_tags'] = ' '.join(doc.tags['genre_tags'])
```
makram93, line 64: Why do we need to join the list of tags to a string?

tests/integration/test_end_to_end.py
azayz: okay makes sense

now/data_loading/data_loading.py
azayz: annlite indexer doesn't accept list type as one of the columns therefore I must transform this into a string and in the playground transform back into a list

makram93: Okay, this makes sense for this demo app that we deploy. However, for users is it clear that they can not specify the list of values for a single column? Either we should handle this silently in the background (in indexer - concatenating the list of values when inserting and then returning the list of values) or mention this explicitly. For this, we can have another ticket, not needed to be addressed in this.

makram93 (approved):
LGTM!

PR #567, opened 2022-10-05
Fix add static url

makram93:


now/apps/base/app.py
```diff
@@ -223,6 +224,27 @@ def preprocess(
...
+        hosted_ds = DEFAULT_EXAMPLE_HOSTED.get(self.app_name, {})
+        if (
+            hosted_ds
+            and user_input.data in hosted_ds
+            and user_input.deployment_type == 'remote'
+            and 'NOW_EXAMPLES' not in os.environ
+            and 'NOW_CI_RUN' not in os.environ
+        ):
+            client = Client(
+                host=f'grpcs://noow-example-{self.app_name}-{user_input.data}.dev.jina.ai'.replace(
```
violenil, line 237: Is this double 'o' in `f'grpcs://noow-example-...'` intentional or a typo?

makram93: Thanks for pointing that out.. It is indeed a typo :D

PR #574, opened 2022-10-07
feat: github actions for example deployment

makram93:


now/admin/deploy_examples.py
```diff
@@ -1,56 +1,119 @@
...
         with ProcessPoolExecutor() as thread_executor:
-            futures = []
-            for app, data in params:
-                f = thread_executor.submit(deploy, app, data)
-                futures.append(f)
-            for f in futures:
-                results.append(f.result())
+            # call delete function with each flow
+            thread_executor.map(lambda x: terminate_wolf(x), flow_ids)
+        print('Deploying all examples!!')
```
violenil, line 89: Is it possible that some apps should also be re-deployed when `deployment_type == 'all'`?

makram93: When `deployment_type == 'all'` then all example apps will be re-deployed else only those apps will be re-deployed which are down. Does this answer your question?

florian-hoenicke:
LGTM

PR #578, opened 2022-10-10
refactor: remove post exec

makram93:


now/constants.py
```diff
@@ -6,9 +6,8 @@
 DEMO_DATASET_DOCARRAY_VERSION = '0.13.17'
 DOCKER_BFF_PLAYGROUND_TAG = '0.0.128-add-cname-1'
 
-NOW_PREPROCESSOR_VERSION = '0.0.91-feat-add-qdrant-indexer-35'
-NOW_ANNLITE_INDEXER_VERSION = '0.0.12-feat-add-qdrant-indexer-35'
```
azayz, line 10: why is annlite version deleted ?

now/apps/text_to_text/ft-flow-sbert.yml
```diff
@@ -53,10 +53,6 @@ executors:
         memory: 4G
     env:
       JINA_LOG_LEVEL: DEBUG
-  - name: postprocessor
```
azayz, line 56: I recall @joschkabraun  saying that post processor is still needed, but I believe you moved all post processor behaviour to the base indexer am I correct ?

makram93: Yes... Postprocessor only removes blob/tensor and this logic is now in base indexer

PR #586, opened 2022-10-12
feat: dialog options for search and filter fields

jupyterjazz:


jupyterjazz:
Merged with https://github.com/jina-ai/now/pull/587

PR #596, opened 2022-10-14
Fix deploy example script

makram93:
Add an endpoint to get the status of the host.
Script calls this status to check the status and then deploy

### DIFF
--- a/.github/workflows/force-deploy-examples.yml
+++ b/.github/workflows/force-deploy-examples.yml
@@ -12,6 +12,8 @@ on:
         required: false
         default: PARTIAL
 
+  pull_request:
+
 jobs:
   python-script:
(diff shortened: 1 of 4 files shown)


joschkabraun (approved):
LGTM!

Out of curiosity: Does `jina_client.is_flow_ready` not work or does it throw an error when you try to check if flow exists?

makram93:
> LGTM!
> 
> Out of curiosity: Does `jina_client.is_flow_ready` not work or does it throw an error when you try to check if flow exists?

It works but has problem with multiprocessing, since `Client` also use threadings. It is discussed in this [thread](https://jinaai.slack.com/archives/C018F60RBL5/p1665401944958959)

PR #597, opened 2022-10-17
test: run multimodal demo in end-to-end-tests

violenil:
This PR makes the following changes:

- relevant routers are added to the BFF for the text-to-text-and-image app
- finetuning of models when deploying the `text-to-text-and-image` app is disabled for now
  - reason: `access_paths` are not supported by FinetunerExecutor (hosts finetuned models for encoding), which we cannot do without since our documents are inherently nested ([issue](https://github.com/jina-ai/finetuner-core/issues/363) has been created)
- the `text-to-text-and-image` is using ElasticIndexer, though this means deployments can only be local (remote option not possible at this time)
- the ElasticIndexer supports `docs_matrix` as input, and has a merging function that selects relevant documents at chunk level depending on whether the request is indexing or searching

Future work:

- Frequently, a deployment cannot be cleaned up particularly when deployment is local. This results in the following error message (see below). Because the cluster is not removed, further unit tests in the BFF (in the same session) will be able to establish a connection on the default port (31080) and host (localhost) when they should not be able to. These are the tests like `test_[...]_fails_with_no_flow_running`.
```
There is no current event loop in thread 'MainThread'.
```
- We need to ensure that local deployments are properly cleaned up after tests ([issue](https://github.com/jina-ai/now/issues/634) created)
- ElasticIndexer currently expects query document chunks and index document chunks to have the same order, and be indexed by the same encoders ie.
  - document has SBert embedding in first chunk, CLIP-vision embedding in second chunk
  - query also has SBert embedding in first chunk, CLIP-text embedding in second chunk
  - if the order is inconsistent, ElasticIndexer is not robust enough to handle this
- allow remote deployment with ElasticIndexer


now/app/text_to_text_and_image/flow.yml
```diff
@@ -28,6 +28,7 @@ executors:
   uses_with:
     artifact: ${{ ENV.SBERT_ARTIFACT }}
     token: ${{ ENV.JINA_TOKEN }}
+    is_onnx: True
```
florian-hoenicke, line 31: why is this flag needed? Can we do without?

violenil: We are using an onnx model artifact, without this flag the executor cant be loaded. Perhaps we can use a pytorch model instead (this is supported with FinetunerExecutor)

now/app/text_to_text_and_image/app.py
```diff
@@ -149,9 +149,11 @@ def setup(
             kubectl_path=kubectl_path,
             deployment_type=user_input.deployment_type,
         )
-        env_dict['HOSTS'] = indexer_config['hosts']
+        env_dict['HOSTS'] = indexer_config.get('hosts')
```
azayz, line 152: why get ?

now/app/text_to_text_and_image/app.py
```diff
@@ -187,7 +189,7 @@ def _construct_finetune_settings(
         # temporary adjustments to work with small text+image dataset
         finetune_settings.epochs = 2
         finetune_settings.num_val_queries = 5
-        finetune_settings.train_val_split_ration = 0.8
+        finetune_settings.train_val_split_ration = 0.75
```
azayz, line 192: any reason for this change ?

violenil: yes because the test split was too small for finetuner, I think it was only 30 instances, while finetuner requires a minimum 32 samples to make batch.

now/app/text_to_text_and_image/app.py
violenil: this will handle errors better, while fetching a key like `dict[key]` will throw a KeyError

deployment/bff/app/v1/models/text_and_image.py
```diff
@@ -0,0 +1,31 @@
...
+
+# Response Model
+class NowTextAndImageResponseModel(BaseSearchResponseModel):
+    text: Optional[str] = Field(description='Matching text result.', default='')
+    blob: Optional[str] = Field(
+        description='Base64 encoded image in `utf-8` str format'
+    )
+
+
+NowTextAndImageResponseModel.update_forward_refs()
```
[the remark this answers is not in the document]
violenil: I am not sure, honestly. The `NowImageResponseModel` has this too, so that is why I made sure to include this call too.

.test_durations
```diff
@@ -59,7 +59,7 @@
     "tests/integration/data_loading/test_demo_datasets.py::test_dataset_is_available[text-pop-lyrics]": 0.19309283047914505,
     "tests/integration/data_loading/test_demo_datasets.py::test_dataset_is_available[text-rap-lyrics]": 0.19753794418647885,
     "tests/integration/data_loading/test_demo_datasets.py::test_dataset_is_available[text-rock-lyrics]": 0.2014372106641531,
-    "tests/integration/data_loading/test_demo_datasets.py::test_dataset_is_available[text_and_image-extracted-data-online-shop-50-flat]": 0.20864115236327052,
+    "tests/integration/data_loading/test_demo_datasets.py::test_dataset_is_available[text-and-image-extracted-data-online-shop-50-flat]": 0.20864115236327052,
```
azayz, line 62: just out of curiosity why does this have to change ?

deployment/bff/app/v1/routers/text2text.py
```diff
@@ -71,11 +71,14 @@ def suggestion(data: NowTextSearchRequestModel):
...
     """
-    query_doc, _ = process_query(text=data.text, uri=data.uri, conditions=data.filters)
+    query_doc, filter_query = process_query(
+        text=data.text, uri=data.uri, conditions=data.filters
+    )
 
     docs = jina_client_post(
         data=data,
         inputs=query_doc,
+        parameters={'limit': data.limit, 'filter': filter_query, 'apply_bm25': True},
```
azayz, line 81: I think the apply_bm25 is not needed for this endpoint

now/executor/indexer/elastic/elastic_indexer.py
```diff
@@ -145,7 +162,17 @@ def search(
...
-            return
+            return docs
+        if docs_matrix:
+            if len(docs_matrix) > 1:
+                docs = self._join_docs_matrix_into_chunks(
+                    docs_matrix=docs_matrix, on='search'
+                )
+            else:
+                docs = docs_matrix[0]
+        docs.summary()
```
azayz, line 173: these were used for debugging so you can remove

.test_durations
violenil: I changed the formatting of the output modality (changed underscores to dashes because underscores are not permitted in URLs) and so this test naming also changed.

tests/unit/test_finetuning/test_run_finetuning.py
```diff
@@ -43,7 +43,8 @@ def test_nest_construct_finetune_settings(encoder_type, get_task_config_path):
         user_input=user_input, dataset=dataset, encoder_type=encoder_type
     )
 
-    assert settings.perform_finetuning
+    # TODO: uncomment when finetuning is enabled for text-to-text-and-image
+    # assert settings.perform_finetuning
```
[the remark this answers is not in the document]
violenil: Currently, the text-to-text-and-image app has finetuning disabled, so it cannot be tested. Finetuning is disabled for this app because FinetunerExecutor currently does not support `access_paths`, which are required for encoding nested documents (inherently the data structure of `text-to-text-and-image` app).

now/app/text_to_text/flow-sbert.yml
```diff
@@ -49,4 +49,5 @@ executors:
         memory: ${{ ENV.INDEXER_MEM }}
         cpu: 1.0
     env:
-      JINA_LOG_LEVEL: DEBUG
\ No newline at end of file
+      JINA_LOG_LEVEL: DEBUG
+    disable_reduce: true
```
jupyterjazz, line 53: what's `disable_reduce`?

now/app/text_to_text_and_image/app.py
```diff
@@ -149,9 +167,11 @@ def setup(
             kubectl_path=kubectl_path,
             deployment_type=user_input.deployment_type,
         )
-        env_dict['HOSTS'] = indexer_config['hosts']
+        env_dict['HOSTS'] = indexer_config.get('hosts')
         env_dict['INDEXER_NAME'] = f"jinahub+docker://{indexer_config['indexer_uses']}"
-        env_dict['INDEXER_MEM'] = indexer_config['indexer_resources']['INDEXER_MEM']
+        env_dict['INDEXER_MEM'] = indexer_config.get('indexer_resources').get(
+            'INDEXER_MEM'
+        )
```
jupyterjazz, line 174: if you want to avoid `KeyError` and return `None` if there's no indexer resources:
```suggestion
        env_dict['INDEXER_MEM'] = indexer_config.get('indexer_resources', {}).get(
            'INDEXER_MEM'
        )
```

now/app/text_to_text_and_image/app.py
```diff
@@ -160,12 +180,16 @@ def setup(
             ] = 0  # JCloud will delete after 24hrs of being idle if not deleted in CI
         else:
             env_dict['RETENTION_DAYS'] = -1  # for user deployment set it to 30 days
+        env_dict['N_DIM'] = [768, 512]
```
jupyterjazz, line 183: would be nice to have constants for the dimensions

now/executor/indexer/elastic/elastic_indexer.py
```diff
@@ -102,9 +106,18 @@ def index(
...
+            if len(docs_matrix) > 1:
+                docs = self._join_docs_matrix_into_chunks(
+                    docs_matrix=docs_matrix, on='index'
+                )
+            else:
+                docs = docs_matrix[0]
         if not parameters:
             parameters = {}
+        docs.summary()
+        docs[0].summary()
```
jupyterjazz, line 120: these summaries should also be removed

now/executor/indexer/elastic/elastic_indexer.py
```diff
@@ -376,8 +403,9 @@ def _transform_es_results_to_matches(self, es_results: List[Dict]) -> DocumentAr
             matches.append(d)
         return matches
 
+    @staticmethod
     def _extract_embeddings(
-        self, doc: Document, traversal_paths: str
+        doc: Document, traversal_paths: str
```
jupyterjazz, line 408: instead of 
`if 'r' in traversal_paths:
`
why not `traversal_paths == '@r'`

jupyterjazz: this might cause some problems in the future when traversal paths will contain field names..

now/executor/indexer/elastic/elastic_indexer.py
```diff
@@ -396,3 +424,52 @@ def _extract_embeddings(
...
+                    raise Exception('Embedding size not supported')
+                new_da.append(new_doc)
+        else:
+            for doc1, doc2 in zip(*docs_matrix):
+                new_doc = Document()
+                text_chunks = [
+                    c for c in doc1.chunks if c.text is not None and c.text != ''
+                ]  # doc1 from SBert with text embedding
+                image_chunks = [
+                    c for c in doc2.chunks if c.uri is not None and c.uri != ''
```
jupyterjazz, line 469: I think this could also work?
```suggestion
                    c for c in doc1.chunks if c.text
                ]  # doc1 from SBert with text embedding
                image_chunks = [
                    c for c in doc2.chunks if c.uri
```

deployment/bff/app/v1/routers/text2text.py
violenil: Although it is not strictly necessary, I think it is quite a nice feature to allow for text-to-text search, since this is one of the few applications with input+output modality that allow bm25 scoring. So I think we should keep it :) 
If an indexer is used (eg. Qdrant) that does not support bm25 scoring, it will just ignore this parameter anyway.

tests/executor/indexer/elastic/conftest.py
```diff
@@ -85,3 +85,59 @@ def text_da():
...
+@pytest.fixture
+def docs_matrix_index():
+    return [
+        DocumentArray(
+            Document(
+                chunks=[
+                    Document(
+                        text='this', embedding=np.ones(768)
+                    ),  # embedding from SBERT
+                    Document(uri='https://jina.ai'),  # not encoded by SBERT
```
jupyterjazz, line 99: why are we passing 'https://jina.ai' as an uri?

tests/unit/data_loading/test_data_loading.py
```diff
@@ -133,8 +133,8 @@ def test_es_online_shop_ds(da: DocumentArray):
 
     for doc in loaded_da:
         assert doc.chunks
-        assert doc.chunks[0].text
-        assert doc.chunks[1].uri
+        for c in doc.chunks:
+            assert c.text is not None or c.uri is not None
```
jupyterjazz, line 137:
```suggestion
            assert c.text or c.uri
```

jupyterjazz (commented):
Great work! :rocket: :partying_face:

left some minor comments

now/app/text_to_text/flow-sbert.yml
violenil: This parameters allows an executor to gather results (or DocumentArrays) from multiple previous executors (specified with `needs`), ie. it gets a `docs_matrix` consisting of a list of DocumentArrays, where each DocumentArray comes from a different executor. If this parameter is not set, the Indexer only receives output from the last executor in the `needs` parameter.

now/app/text_to_text_and_image/app.py
violenil: Good idea!

now/executor/indexer/elastic/elastic_indexer.py
violenil: The traversal path could contain 'r' as well as 'c', for example '@r,c' in which case this condition would not hold. We will need to refactor this in the future when traversal paths contain field names. But I don't think  `traversal_paths == '@r'` solves this issue for now.

tests/executor/indexer/elastic/conftest.py
violenil: Will replace with random but realistic uri.

now/executor/indexer/elastic/elastic_indexer.py
violenil: Done.

now/executor/indexer/elastic/elastic_indexer.py
violenil: Done.

now/app/text_to_text_and_image/app.py
violenil: Done.

now/app/text_to_text_and_image/app.py
violenil: Done.

tests/unit/data_loading/test_data_loading.py
violenil: Done.

deployment/bff/app/v1/routers/txt2txt_and_img.py
```diff
@@ -0,0 +1,75 @@
...
+    "/index",
+    summary='Add more text and image data to the indexer',
+)
+def index(data: NowTextAndImageIndexRequestModel):
+    """
+    Append the list of image data to the indexer. Each image data should be
+    `base64` encoded using human-readable characters - `utf-8`.
+    """
+    index_docs = DocumentArray()
+    for text, image, uri, tags in zip(data.texts, data.images, data.uris, data.tags):
```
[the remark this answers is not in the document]
violenil: > this will error if they don't have same length

That would be expected, as we expect these to have the same length.

now/executor/indexer/elastic/elastic_indexer.py
```diff
@@ -144,37 +160,44 @@
~                - 'traversal_paths' (str): traversal paths for the docs
                 - 'limit' (int): nr of matches to get per Document
         """
         if not docs:
-            return
+            return docs
+        if docs_matrix:
+            if len(docs_matrix) > 1:
+                docs = self._join_docs_matrix_into_chunks(
```
joschkabraun, line 167: Is now `docs_matrix` deterministic?

now/executor/indexer/elastic/elastic_indexer.py
```diff
@@ -144,37 +160,44 @@
...
+                    docs_matrix=docs_matrix, on='search'
+                )
+            else:
```
joschkabraun, line 170: this doesn't work with documents for which we directly uploaded `blob` instead of `uri`

deployment/bff/app/v1/routers/text2text.py
joschkabraun, line 81: Why do we need `apply_bm25`, `limit` and `filter` for the `suggestion` endpoint?

deployment/bff/app/v1/routers/text2text.py
```diff
@@ -56,7 +56,7 @@ def search(data: NowTextSearchRequestModel):
     docs = jina_client_post(
         data=data,
         inputs=query_doc,
-        parameters={'limit': data.limit, 'filter': filter_query},
+        parameters={'limit': data.limit, 'filter': filter_query, 'apply_bm25': True},
```
joschkabraun, line 59: For remote deployment for `text2text` we use Qdrant, right? I assume `apply_bm25` will then simply be ignored, right?

now/app/text_to_text_and_image/pretrained_flow.yml
```diff
@@ -0,0 +1,76 @@
...
+  uses_with:
+    model_name: sentence-transformers/msmarco-distilbert-base-v4
+    pretrained_model_name_or_path: sentence-transformers/msmarco-distilbert-base-v4
+    access_paths: '@c'
+  env:
+    JINA_LOG_LEVEL: DEBUG
+  needs: preprocessor
+  jcloud:
+    resources:
+      memory: 4G
```
joschkabraun, line 42: Are `4G` actually needed? E.g. `text2video` uses `1G`

now/common/options.py
```diff
@@ -58,7 +58,6 @@
         {
             'name': '📝 ▶ 📝+🏞 text to text+image search',
             'value': Apps.TEXT_TO_TEXT_AND_IMAGE,
-            'disabled': AVAILABLE_SOON,
```
joschkabraun, line 61: Wuhu!!

now/common/preprocess.py
```diff
@@ -11,12 +11,12 @@
~        try:
~            if d.tensor is None:
~                if d.blob != b'':
~                    d.convert_blob_to_image_tensor()
~                elif d.uri:
~                    d.load_uri_to_image_tensor(timeout=10)
~            return to_thumbnail_jpg(d)
```
joschkabraun: Can be simplified to `if d.text`

now/common/preprocess.py
```diff
@@ -11,12 +11,12 @@
...
~        except:
~            return d
~
~    for d in da:
```
joschkabraun: Can be simplified to `if d.text`

now/common/utils.py
```diff
@@ -75,8 +75,8 @@
~            "pretrained_model_name_or_path"
~        ]
~    if finetune_settings.perform_finetuning:
~        config['FINETUNE_ARTIFACT'] = finetune_settings.finetuned_model_artifact
```
joschkabraun: Doesn't that deactivate the `text2text` app remotely?

joschkabraun (changes requested):
Really great work! :rocket:
Some small comments

now/common/utils.py
```diff
@@ -199,6 +199,10 @@ def get_indexer_config(
             'indexer_uses': f'ElasticIndexer/{NOW_ELASTIC_INDEXER_VERSION}',
             'hosts': setup_elastic_service(kubectl_path),
         }
+    elif elastic and deployment_type == 'remote':
+        raise ValueError(
+            'ElasticIndexer is currently not supported for remote deployment. Please use local deployment.'
```
joschkabraun, line 204: Doesn't that deactivate the `text2text` app remotely?

now/common/preprocess.py
```diff
@@ -67,7 +67,16 @@ def _get_sentence_docs(batch):
     if split_by_sentences:
         da = DocumentArray(d for d in gen_split_by_sentences())
 
-    return DocumentArray(d for d in da if d.text and d.text != '')
+    result_da = DocumentArray()
+    for d in da:
+        if d.text and d.text != '':
```
joschkabraun, line 72: Can be simplified to if `d.text`

now/common/options.py
florian-hoenicke: 🥳 🥳 🥳

now/app/text_to_text_and_image/flow.yml
```diff
@@ -23,42 +23,41 @@ executors:
 - name: sbert_encoder
   uses: jinahub+docker://FinetunerExecutor
   uses_requests:
-    '/query_text': encode
-    '/index_text': encode
+    /search: encode
```
florian-hoenicke, line 26: is it still necessary to define these two endpoints or can we also do without?

now/executor/indexer/elastic/elastic_indexer.py
violenil: When I was implementing this, only one app was passing a `docs_matrix` to the indexer, which is the multimodal app with multiple encoders. In this sense, I was assuming that this `docs_matrix` would be deterministic, and also assumed a strict order for the DocumentArrays in the `docs_matrix`. Now that I have looked at [your PR](https://github.com/jina-ai/now/pull/601) I realize this may not be the case if OCR is enabled. 
Can you think of a way to better structure this?

now/app/text_to_text_and_image/flow.yml
violenil: It is not necessary, though this flow.yml is not currently used in the app. It would be used again when finetuning is enabled for text-to-text-and-image, and these endpoints (and also the whole yml) can then be refactored.

now/app/text_to_text_and_image/pretrained_flow.yml
violenil: Ah ok, I was not sure what would be needed. I will amend this :)

deployment/bff/app/v1/routers/text2text.py
violenil: Indeed! Should just be ignored.

deployment/bff/app/v1/routers/text2text.py
violenil: The app was failing when I had not passed these parameters also to the suggestion endpoint. I cannot remember the reason exactly, will take another look.

azayz (approved):
LGTM after comments above

now/common/utils.py
joschkabraun: But isn't `elastic=True` in `setup` for `text2text`?

now/executor/indexer/elastic/elastic_indexer.py
violenil: Would it make sense to check the `modality` or `mime_type` instead?

now/common/utils.py
violenil: Changing this default to `elastic=False`

now/app/text_to_text_and_image/flow.yml
violenil: I have refactored this in anticipation of enabling finetuning again in this app in a future PR :)

now/executor/indexer/elastic/elastic_indexer.py
violenil: We can do some refactoring on this once core find a solution to [this issue](https://github.com/jina-ai/jina/issues/5330).

PR #612, opened 2022-10-21
docs: reword, number readme points

alexcg1:
1. Numbered every point to guide user effectively.
2. Removed duplicated information
3. Trimmed down text
4. Fixed some spelling and grammar


alexcg1:
Numbers are in bold, otherwise markdown sees them as start of a new list. That means even if I write:

```
1. foo

bar baz qux

2. plonk
````

It would come out as:

```
1. foo

bar baz qux

1. plonk
````

Wrapping the numbers in `**` gets around this

azayz (approved):

PR #615, opened 2022-10-24
fix: bff + refactor + feat

florian-hoenicke:
There are some issues with the bff.
Example: when you index a batch of images in combination with `tags`, all images get the same tags.
image


florian-hoenicke:
@azayz Is this what you expect for the suggestion endpoint?
image

azayz:
> @azayz Is this what you expect for the suggestion endpoint? image

yes, one question though, what happens if you pass extra arguments (like filters and stuff like that) ? because previously I use the same request model as search, and for response yeah we only need tags

florian-hoenicke:
> what happens if you pass extra arguments
It will fail in that case. I guess I would have to change the playground accordingly.

PR #627, opened 2022-10-31
Fix add sync

makram93:


now/common/options.py
```diff
@@ -304,6 +305,20 @@
...
 
+
+API_KEY = DialogOptions(
+    name='api_key',
+    prompt_message='Do you want to generate an api_key to access this deployment?',
+    prompt_type='list',
+    choices=[
+        {'name': '✅ yes', 'value': True},
+        {'name': '⛔ no', 'value': False},
+    ],
```
joschkabraun, line 316: I would not let it check that the user has set the api key. Otherwise other users don't have access to the playground without extra calling the flow again

now/common/options.py
```diff
@@ -312,8 +327,8 @@
```
joschkabraun: Did you mean `if not user_input.api_key:`?

joschkabraun: Did you mean `if not user_input.api_key`?

now/common/options.py
```diff
@@ -304,6 +305,20 @@
     conditional_check=lambda user_inp: user_inp.deployment_type == 'remote',
 )
 
+
+API_KEY = DialogOptions(
+    name='api_key',
+    prompt_message='Do you want to generate an api_key to access this deployment?',
```
joschkabraun, line 311: I think we can change `'value'` to `uuid.uuid4().hex` for `True` and to `None` for `False`. Then we don't need to set it to bool.

now/now_dataclasses.py
```diff
@@ -116,6 +116,7 @@ class UserInput(BaseModel):
     admin_emails: Optional[List[str]] = None
     user_emails: Optional[List[str]] = None
     additional_user: Optional[StrictBool] = None
+    api_key: Optional[Union[StrictBool, str]] = None
```
joschkabraun, line 119: By comment in `options.py` we can remove type `StrictBool`

tests/integration/test_end_to_end.py
```diff
@@ -165,6 +164,7 @@ def test_backend_demo_data(
         'dataset_name': dataset,
         'cluster': cluster,
         'secured': deployment_type == 'remote',
+        'api_key': False,
```
florian-hoenicke, line 167: should it be None as default value?

florian-hoenicke: ah okay, nevermind. I got it. It is a boolean asking if they want to secure it via api_key

joschkabraun (commented):
Nice PR! Some minor comments

florian-hoenicke:
image
I don't get asked anymore if I want to specify email addresses

florian-hoenicke (changes requested):

now/common/options.py
makram93: Nice idea

now/common/options.py
makram93: Can you elaborate what you mean about the users making an extra call to the flow?

tests/integration/test_end_to_end.py
makram93: Updated it

now/common/options.py
joschkabraun: It's similar to what Florian said that one doesn't get asked anymore about adding user emails. One would have to update user emails after program termination.

now/common/options.py
```diff
@@ -192,7 +193,7 @@
~
~
~def _parse_text_fields(user_input: UserInput):
~    user_input.es_text_fields = [
~        field.strip() for field in user_input.es_text_fields.split(',')
~    ]
~
```
joschkabraun: it's API key

now/run_backend.py
```diff
@@ -55,30 +60,81 @@
...
+
+
+def trigger_scheduler(user_input, host):
+    """
+    This function will trigger the scheduler which will sync the bucket with the indexer once a day
+    """
+    print('Triggering scheduler to index data from S3 bucket')
+    # check if the api_key exists. If not then create a new one
+    if user_input.secured and not user_input.api_key:
+        user_input.api_key = uuid.uuid4().hex
```
joschkabraun, line 89: I would add a print out to let the user know that this is the API key. For the case that they update the api keys in the future, they will need to keep that one for the sync job.

now/run_backend.py
```diff
@@ -55,30 +60,81 @@
...
+        # Also call the bff to update the api key
+        update_api_keys(user_input.deployment_type, user_input.api_key, host)
```
joschkabraun, line 91: We need to let this run as a for loop for now in order to ensure every replica of an executor receives the updated API keys if we have multiple replicas of the same executor.

tests/integration/test_end_to_end.py
joschkabraun, line 167: I would change it to `None`

joschkabraun (approved):
Minor comments, then LGTM!

now/run_backend.py
makram93: Yes at the end this will be printed along with playground & bff url

now/app/music_to_music/demo-data-flow-music.yml
```diff
@@ -19,6 +19,7 @@ executors:
       app: ${{ ENV.APP }}
       admin_emails: ${{ ENV.ADMIN_EMAILS }}
       user_emails: ${{ ENV.USER_EMAILS }}
+      api_keys: ${{ ENV.API_KEY }}
```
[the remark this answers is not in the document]
makram93: Nice suggestion! Have updated my code

azayz (approved):
LGTM

PR #630, opened 2022-10-31
fix: add timeout

### LINKED ISSUE #631: fix indexing taking a lot of time
it has been observed that for some datasets indexing may take a lot of time and this is because of loading uri from various websites takes time

azayz:
close #631

### DIFF
--- a/deployment/playground/playground.py
+++ b/deployment/playground/playground.py
@@ -440,7 +440,7 @@ def render_matches(OUTPUT_MODALITY):
 
                 if OUTPUT_MODALITY == 'text':
                     if match.text == '' and match.uri != '':
-                        match.load_uri_to_text()
+                        match.load_uri_to_text(parameters={'timeout': 30})
                     display_text = profanity.censor(match.text).replace('\n', ' ')
                     body = f"<!DOCTYPE html><html><body><blockquote>{display_text}</blockquote>"
                     if match.tags.get('additional_info'):
@@ -469,7 +469,7 @@ def render_matches(OUTPUT_MODALITY):
 
                 elif OUTPUT_MODALITY == 'music':
                     if match.uri:
-                        match.load_uri_to_blob()
+                        match.load_uri_to_blob(parameters={'timeout': 30})
                     display_song(c, match)
 
                 elif OUTPUT_MODALITY in ('image', 'video'):
--- a/deployment/playground/playground.py
+++ b/deployment/playground/playground.py
@@ -440,7 +440,7 @@ def render_matches(OUTPUT_MODALITY):
 
                 if OUTPUT_MODALITY == 'text':
                     if match.text == '' and match.uri != '':
-                        match.load_uri_to_text(parameters={'timeout': 30})
+                        match.load_uri_to_text(timeout=10)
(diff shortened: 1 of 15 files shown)


deployment/playground/playground.py
```diff
@@ -440,7 +440,7 @@ def render_matches(OUTPUT_MODALITY):
 
                 if OUTPUT_MODALITY == 'text':
                     if match.text == '' and match.uri != '':
-                        match.load_uri_to_text()
+                        match.load_uri_to_text(timeout=10)
```
makram93, line 443: The version needs to be updated in the playground as well. Make sure it is not failing after the version upgrade by doing local tests and opening streamlit

deployment/playground/src/search.py
```diff
@@ -155,7 +155,7 @@ def search_by_image(document: Document, jwt, filter_selection) -> DocumentArray:
         if query_doc.tensor is not None:
             query_doc.convert_image_tensor_to_blob()
         elif (query_doc.uri is not None) and query_doc.uri != '':
-            query_doc.load_uri_to_blob()
+            query_doc.load_uri_to_blob(timeout=10)
```
makram93, line 158: Does it throw an error on timeout or just suppress it and continue the execution?

makram93, line 158: Also, I would suggest reducing the timeout value.... maybe to 5? It will greatly factor in indexing large datasets with lots of invalid urls

makram93 (changes requested):
Some minor changes, otherwise it looks good.

jupyterjazz (commented):
Do we have some experimentation results? i.e. how this change affected speed of indexing certain datasets

deployment/playground/src/search.py
azayz: I believe 5 is too fast, I think this is also a tradeoff of how fast you want to index your data vs how incomplete your indexed data will be, I believe having complete indexed data is more important than speed, what do you think ?

azayz: surpress and continue the execution

azayz: throws an error 
    return self.read(nbytes, buffer)
  File "/usr/lib/python3.8/ssl.py", line 1099, in read
    return self._sslobj.read(len, buffer)
socket.timeout: The read operation timed out 

do you think this might be problematic ?

makram93: If it throws an error and terminates the program then we should add it in try-except block so as to continue with other docs

makram93: Okay 👍

azayz: I see but the indexing continues when I tried it, I think this is because I added continue on error argument in indexing

deployment/playground/playground.py
azayz: I tested locally with text to image with cars dataset and playground works fine

florian-hoenicke (approved):
LGTM

florian-hoenicke:
> Do we have some experimentation results? i.e. how this change affected speed of indexing certain datasets

@jupyterjazz It will only have positive effects. Before, it was not timing out and led to permanent blocking of the indexing process.

PR #637, opened 2022-11-07
fix: format the script-score query source correctly

violenil:
In the ElasticIndexer, a query is built and sent to Elasticsearch for retrieving documents. The format of the query is flawed when `apply_bm25 = False` is set (the current default setting of this parameter). This is because the placement of the plus sign when adding different segments to the source of the query is incorrect. This PR aims to solve this by moving the plus sign to the end of each component added to the source of the query.

In order to confirm that the query is built correctly, this PR also adds a test to the executor tests, which only performs vector search.

Summary:

- [ ] correctly format script-score query source
- [ ] add test for executor using only vector search (no bm25)


florian-hoenicke (approved):
LGTM

PR #652, opened 2022-11-10
docs: explain input format

florian-hoenicke:
better explanation  for input format

### DIFF
--- a/docs/user-guides/input_format.md
+++ b/docs/user-guides/input_format.md
@@ -1,0 +1,77 @@
+# Input Formats
+
+When deploying your app, you will be asked to select the input format of your data. 
+This is the format of the data that you will be sending to the API. 
+We currently support the following input options for custom input: DocumentArray, DocumentArrayUrl, local path, S3 bucket.
+In the following, we will explain each of these options in more detail.
+
+## DocumentArray
+If you have your data as a `DocumentArray` you can use this option.
+More information about how to create and push your own `DocumentArray` can be found [here](https://docarray.jina.ai/).
+
+## DocumentArray Url
+Similar to the `DocumentArray` option, but instead of pushing your `DocumentArray` to the API, you can provide a URL to your data.
+
+## Local Folder
+If you have your data stored locally, you can provide the path to the folder containing your data.
+The folder should contain all files that you want to index.
+
+Example structure for a text-to-image search app:
+```
+usr
+├── data
+│   ├── images
+│   │   ├── 1.jpg
+│   │   ├── 2.jpg
+│   │   ├── 3.jpg
+│   │   ├── 4.jpg
+```
+The local path you provide should be `/usr/data/images`.
+
+## AWS S3 bucket
+If you have your data stored in an AWS S3 bucket, you can provide the S3 URI, your `AWS access key ID` and `AWS secret key`.
+Similar to the local folder option, the S3 bucket should contain all files that you want to index.
+The only difference is that the S3 Uri should be in the following format: `s3://<bucket-name>/<path-to-data>`.
+Taking the example structure from above, the S3 URI would be `s3://my-<bucket-name>/usr/data/images`.
+
+
+# Support for Tags
+For most professional applications, you will want to provide additional information about your data, such as price, creator, brand, etc.
+This information can be provided in the `tags` of your `Documents` in case you selected the `DocumentArray` or `DocumentArrayUrl` option.
+If you selected the `local folder` or `S3 bucket` option, you can provide this information in a `json` file for each document.
+The example from above would look like this:
+```
+root
+├── data
+│   ├── images
+│   │   ├── 1
+│   │   │   ├── 1.jpg
+│   │   │   ├── 1.json
+│   │   ├── 2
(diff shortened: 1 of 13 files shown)


docs/user-guides/input_format.md
```diff
@@ -0,0 +1,77 @@
...
+```
+
+
+
+
+# supported file formats
+Here is an overview of the supported file formats for each modality:
+
+- Text: `.txt` (can also have a different extension, but has to be plain text)
+- Image: `.jpg`, `.png`, ... (everything supported by `PIL`)
```
joschkabraun, line 75: The app only mentions `['gif', 'png', 'jpg', 'jpeg']`. We could change the app requirements

joschkabraun (approved):
LGTM!

docs/user-guides/input_format.md
```diff
@@ -0,0 +1,77 @@
+# Input Formats
+
+When deploying your app, you will be asked to select the input format of your data. 
+This is the format of the data that you will be sending to the API. 
+We currently support the following input options for custom input: DocumentArray, DocumentArrayUrl, local path, S3 bucket.
+In the following, we will explain each of these options in more detail.
+
+## DocumentArray
+If you have your data as a `DocumentArray` you can use this option.
+More information about how to create and push your own `DocumentArray` can be found [here](https://docarray.jina.ai/).
```
jupyterjazz, line 10: To me this part is not informative enough. I'm wondering if we can make it more helpful.

Does it make sense to show minimal examples of docarray usage in our guides? Or maybe direct users to more specific pages in docarray documentation. Especially after we allow multimodal input structure, it feels like we have to communicate about it a bit more.

docs/user-guides/input_format.md
```diff
@@ -0,0 +1,77 @@
...
+{
+    "price": 100,
+    "brand": "Jina"
+}
+```
+
+
+
+
+# supported file formats
```
jupyterjazz, line 71:
```suggestion
# Supported file formats
```

docs/user-guides/input_format.md
florian-hoenicke: Thanks, I extended the number of supported files. However, we need to iterate over this in the future and extend the list even more.

docs/user-guides/input_format.md
florian-hoenicke: thanks

docs/user-guides/input_format.md
```diff
@@ -0,0 +1,77 @@
+# Input Formats
+
+When deploying your app, you will be asked to select the input format of your data. 
+This is the format of the data that you will be sending to the API. 
+We currently support the following input options for custom input: DocumentArray, DocumentArrayUrl, local path, S3 bucket.
+In the following, we will explain each of these options in more detail.
+
+## DocumentArray
```
fogx, line 8: This should be `## DocumentArray Name` or `DocArray Name`?
Is this all the information required? I can upload any kind of valid docArray and it will automatically work? If not please specify what constraints/requirements exist.
Team Core just finished a more [detailed description/how-to on creating docArrays](https://github.com/docarray/docarray/pull/728) - lets link to there once it's released

CatStark:
I have some comments :D 

1. It's confusing for an user that the documentation say we support local files but that's not true for the storefront
2. The part of Docarray is super brief, and it doesn't explain anything concrete. If this will be handled by docarray team, at least the link should be to the specific part on how to create docarrays for Jina NOW, all the rest is not useful for users
3. Document Url, should be URL? it looks bad without being capitalised 
4. The Support for tags is the same as docarray, it should link to the specific part on Docarray documentation since it's not clear here

docs/user-guides/input_format.md
```diff
@@ -0,0 +1,77 @@
...
+If you have your data as a `DocumentArray` you can use this option.
+More information about how to create and push your own `DocumentArray` can be found [here](https://docarray.jina.ai/).
+
+## DocumentArray Url
+Similar to the `DocumentArray` option, but instead of pushing your `DocumentArray` to the API, you can provide a URL to your data.
+
+## Local Folder
```
CatStark, line 15: Can I have multiple folders inside?

docs/user-guides/input_format.md
```diff
@@ -0,0 +1,77 @@
...
+The local path you provide should be `/usr/data/images`.
+
+## AWS S3 bucket
+If you have your data stored in an AWS S3 bucket, you can provide the S3 URI, your `AWS access key ID` and `AWS secret key`.
+Similar to the local folder option, the S3 bucket should contain all files that you want to index.
+The only difference is that the S3 Uri should be in the following format: `s3://<bucket-name>/<path-to-data>`.
+Taking the example structure from above, the S3 URI would be `s3://my-<bucket-name>/usr/data/images`.
+
+
+# Support for Tags
```
CatStark, line 38: We either capitalised all words on titles or we don't, but the mix is weird

fogx:
> It's confusing for an user that the documentation say we support local files but that's not true for the storefront
@CatStark this documentation is in the NOW documentation and we will have a subset of it in the Cloud documentation.

PR #654, opened 2022-11-11
fix: executor inheritance issues

florian-hoenicke:


now/executor/abstract/base_indexer/base_indexer.py
```diff
@@ -5,420 +5,446 @@
...
+                    doc_encoded.tags[TAG_OCR_DETECTOR_TEXT_IN_DOC] = doc_text.pop(
+                        TAG_OCR_DETECTOR_TEXT_IN_DOC, ''
+                    )
+                    flat_docs.append(doc_encoded)
+            else:
+                flat_docs = docs[traversal_paths]
+            if len(flat_docs) == 0:
+                return
+
+            self.set_tags_if_text_in_doc_matching_title(flat_docs)
```
joschkabraun, line 161: Similar to the search endpoint, we can also only execute this function `if self.columns and TAG_INDEXER_DOC_HAS_TEXT in [
                col[0] for col in self.columns
            ]:`
to avoid any side effects and surprising behavior

joschkabraun (approved):
LGTM!

PR #655, opened 2022-11-11
fix: split_by_sentences not working in txt2txt app

violenil:


now/data_loading/data_loading.py
```diff
@@ -42,13 +42,13 @@ def load_data(app: JinaNOWApp, user_input: UserInput) -> DocumentArray:
     elif user_input.dataset_type == DatasetTypes.DEMO:
         print('⬇  Download DocumentArray dataset')
         url = get_dataset_url(user_input.dataset_name, app.output_modality)
-        da = fetch_da_from_url(url)
+        da = fetch_da_from_url(url).shuffle()
```
makram93, line 45: Just curious why we are shuffling the documents.

makram93 (approved):
LGTM! Just one question

now/data_loading/data_loading.py
violenil: Previously this shuffle happened further down, when `NOW_CI_RUN` is true. The shuffle is originally done because `best artworks` and `bird species` demo datasets used for testing apps don't have the correct modality as first document, so the indexing is not successful. The shuffling seems to ensure that this does not happen. Instead of always shuffling in CI though, now only the demo datasets are shuffled.

joschkabraun (approved):
LGTM!

PR #658, opened 2022-11-11
feat: add app name

azayz:


scripts/deploy_examples.py
```diff
@@ -50,7 +50,7 @@ def deploy(app_name, app_data):
         'deployment_type': 'remote',
         'proceed': True,
         'secured': False,
-        'ns': NAMESPACE,
+        'flow_name': NAMESPACE,
```
joschkabraun, line 53: As dicussed in pairing, can you make sure there is consistent behavior between `namespace`, `ns`, and `flow_name`?

jupyterjazz (approved):
LGTM!

since it's used in so many places, I'd suggest to create a constant for the default flow name `nowapi`

PR #674, opened 2022-11-16
feat: add migration to jcloud

azayz:


deepankarm (changes requested):
https://github.com/jina-ai/now/blob/e2702669a7d48f4522629edd98e709712712563a/now/deployment/deployment.py#L9

:point_up: needs update as `name` and `env_file` are not supported anymore.

requirements.txt
```diff
@@ -12,7 +12,7 @@ docker==5.0.3
 requests==2.27.1
```
joschkabraun, line 12: why are we not using jina version 3.9.4 here

joschkabraun (approved):
LGTM!

requirements.txt
azayz: you're right I missed this

deepankarm (approved):
:+1:

PR #680, opened 2022-11-17
Fix deploy examples

makram93:
If the demo is accessible then no new deployment will be made for demo datasets and also user won't be asked for login

image

For local demo deployments, either set `NOW_CI_RUN` or `NOW_TESTING` in the environment variable to avoid checking the demo availability
- Fixes the playground bug where it raises error if any one chunk does not contain the content
- Enable `All of the above` option for index fields as well
- Retry if the docarray name is not provided
- No login needed for demo dataset (if it is available, else treated as a new deployment and will ask for login)
-  Enable deploying demo with all fields as indexed fields (This is possible but not yet activate in this PR as there are bug tickets which needs to be fixed first #976 #990 )
- Solve the below issue seen in CI/CD by removing the `raise error` statement in the bff
image


.github/workflows/force-deploy-examples.yml
```diff
@@ -36,7 +43,8 @@ jobs:
         env:
           WOLF_TOKEN: ${{ secrets.WOLF_EXAMPLES_TOKEN }}
       - name: Deploy examples
-        run: python scripts/deploy_examples.py
+        working-directory: ./scripts
```
azayz, line 46: why do you need to specify working directory ?

scripts/deploy_examples.py
```diff
@@ -82,47 +82,34 @@ def deploy(app_name, app_data):
...
 
     if deployment_type == 'all':
         # List all deployments and delete them
         flows = list_all_wolf(namespace=None)
-        flow_ids = [f['id'].replace('jflow-', '') for f in flows]
-        with ProcessPoolExecutor() as thread_executor:
-            # call delete function with each flow
-            thread_executor.map(lambda x: terminate_wolf(x), flow_ids)
-        print('Deploying all examples!!')
+        flow_name = {f['id']: f['id'].split('-')[-1] for f in flows['flows']}
```
azayz, line 95: there is a function to do this directly in now/utils I think

florian-hoenicke:
Is this pr still worked on?

makram93:
@florian-hoenicke Currently it is on hold due to restructuring/refactoring of the code. Once the MMDocs changes are done in the NOW, we can resume this.

.github/workflows/force-deploy-examples.yml
makram93: because the deployment file is inside scripts folder and the runner runs from root.. so either we specify the working directory or give absolute path to the deployment file

.github/workflows/force-deploy-examples.yml
```diff
@@ -12,9 +12,16 @@ on:
         required: false
         default: PARTIAL
 
+#  pull_request:
+
```
jupyterjazz, line 16: this is not needed I guess

now/demo_data.py
```diff
@@ -39,67 +42,81 @@
         DemoDataset(
```
joschkabraun, line 42: We could alternatively also index everything and set the default scoring in the elastic indexer such that only the `image` index fields will be used. Then we don't have to check if this particular combination of index fields has been deployed for a dataset but have everything which saves costs and simplifies the code. The user can play with the scores in the playground if they want to test out using all fields.

jupyterjazz (commented):
do we have an option to deploy all of them at once?
how will choosing the index work? how do we know which index corresponds to which dataset?

joschkabraun (commented):
Awesome that the demos are back! :rocket:

makram93:
> do we have an option to deploy all of them at once?

Yes running the Github Actions should trigger deploying all the examples in parallel

> how will choosing the index work? how do we know which index corresponds to which dataset?

I think what Joschka suggested would work. I can index all the fields and then the user can play around with the semantic scores from the UI

now/demo_data.py
makram93: >set the default scoring in the elastic indexer such that only the image index fields will be used

I would not recommend this as this might need changes in the elastic indexer which will be a bit hacky. However, I like the idea of indexing all fields(setting equal score weightage for all fields in ES) and from the UI user can adjust the semantic scores to see the result variations.

joschkabraun: I think it makes sense to have such an endpoint which allows to set a default semantic score. E.g. in the case of once you optimised your search results, you won't need to sent them anymore (or if different teams work on optimising search vs integrating the search into a frontend, then they can both work in parallel). So we could have an endpoint in elastic which allows that. Then these [lines](https://github.com/jina-ai/now/blob/main/now/executor/indexer/elastic/elastic_indexer.py#L222) would become
```
semantic_scores = parameters.get('semantic_scores', None) or self.default_semantic_scores or generate_semantic_scores(docs_map, self.encoder_to_fields)
```

joschkabraun: I also fear that setting all index_fields with equal score weights will lead to bad results, which wouldn't be a great demo.

now/common/options.py
```diff
@@ -132,6 +135,19 @@ def _get_demo_data_choices(user_input: UserInput, **kwargs):
     ]
 
 
+def _check_if_demo_available(user_input: UserInput, **kwargs):
+    # If the demo is available then do not continue with the dialog and break
+    if user_input.app_instance.is_demo_available(user_input):
+        raise DemoAvailableException(
+            'Demo is available. Not continuing with further dialog'
```
violenil, line 142:
```suggestion
            'Demo is available. Suspending the remaining dialog.'
```

violenil (approved):
Yay! Great to have the demos back.

florian-hoenicke:
In our pairing we now found out that the following things left to do:
- index only relevant fields for the search
- merge latest changes to support auto scaling for the preprocessor for efficient hosting

makram93:
>index only relevant fields for the search

Only the relevant field is indexed though the script can be easily extended to index all fields. 

Following are the fields indexed in the current hosted demo examples
- For image dataset - 'image'
- For text dataset - 'lyrics'
- For gif dataset - 'video'

now/app/search_app/app.py
```diff
@@ -53,20 +50,17 @@ def finetune_datasets(self) -> [Tuple]:
...
-                host=f'grpcs://now-example-{self.app_name}-{user_input.dataset_name}.dev.jina.ai'.replace(
-                    '_', '-'
-                )
+                host=f'grpcs://{DEMO_NS.format(user_input.dataset_name.split("/")[-1])}.dev.jina.ai'
             )
             try:
                 client.post('/dry_run', timeout=2)
-            except Exception:
-                return False
+            except Exception as e:  # noqa E722
```
azayz, line 62: I think this is dangerous, passing an exception it will be impossible to debug

now/common/options.py
```diff
@@ -213,7 +228,12 @@ def _get_demo_data_choices(user_input: UserInput, **kwargs):
     choices=lambda user_input, **kwargs: [
         {'name': field, 'value': field}
         for field in user_input.index_field_candidates_to_modalities.keys()
-    ],
+    ]
+    + (
+        [{'name': 'All of the above', 'value': '__all__'}]
```
azayz, line 233: any reason for adding __, I think you can set the value to anything, why not simply 'all' ?

now/common/options.py
```diff
@@ -256,15 +279,26 @@ def get_models_dialog(user_input: UserInput, **kwargs):
                 choices=MODALITY_TO_MODELS[
                     user_input.index_field_candidates_to_modalities[index_field]
                 ],
-                post_func=lambda user_input, option_name=option_name, **kwargs: update_model_choice(
-                    user_input, option_name, **kwargs
+                post_func=lambda user_inp, opt_name=option_name, **kw: update_model_choice(
```
azayz, line 282: can we keep kwargs as naming kw is less intuitive ?

azayz (commented):
Great job, does this happen when request is coming from JAC only or always ? because I would still like to test my code locally using demo datasets

makram93:
> does this happen when request is coming from JAC only or always ? because I would still like to test my code locally using demo datasets

The demo check will be done whenever now is executed - JAC or CLI. But for locally using demo datasets,  either set `NOW_CI_RUN` or `NOW_TESTING` in the environment variable

now/app/search_app/app.py
makram93: This is not concerned with general execution but just for checking the availability of our demo examples hosted (ideally all the demos should be live all the time, if not so then we have to make sure and fix it). However, if there's an exception raised at this place then the new deployment of the demo dataset will be developed so no problem for the end users and also is not needed for debug

now/common/options.py
makram93: It is just to make sure the value is unique as the there could also be an index or filter fields called `all` and this will conflict

now/common/options.py
makram93: Yes, definitely.

now/common/options.py
azayz: I see

azayz (approved):
LGTM

PR #682, opened 2022-11-18
feat: curation

florian-hoenicke:


joschkabraun (commented):
How will the user provide the curated list?

florian-hoenicke:
> How will the user provide the curated list?

I will handle the bff and the documentation in the other pr. This is just for us to do it manually.
I added a script in the admin folder.

PR #688, opened 2022-11-21
feat: deactivate telemetry on ci

florian-hoenicke:
prevent telemetry when executing tests


now/app/base/flow.yml
```diff
@@ -5,6 +5,7 @@ with:
   monitoring: true
   env:
     JINA_LOG_LEVEL: DEBUG
+    JINA_OPTOUT_TELEMETRY: ${{ ENV.NOW_CI_RUN }}
```
samsja, line 8: just to understand how is the value of `ENV.NOW_CI_RUN` is set ?

florian-hoenicke: When the test is starting.
https://github.com/jina-ai/now/blob/ccb3f5ccbaa379b0abba2b82e4b0b92ffe1ded1c/tests/integration/test_end_to_end.py#L153

samsja: then looks good

nan-wang (changes requested):
this implementation will always disable telemetry, because `JINA_OPTOUT_TELEMETRY` will always be set. Once `JINA_OPTOUT_TELEMETRY` is always set, the telemetry is disabled https://github.com/jina-ai/jina/blob/030a1e90ce5289c20c250084320a5ba1302da21e/jina/helper.py#L1704

nan-wang:
Here is MWE to show the issue.

```python
from jina import Flow, Executor, requests
from docarray import DocumentArray

import os


class MyExecutor(Executor):
    @requests(on='/hello')
    def hello(self, **kwargs):
        # no matter `NOW_CI_RUN` is set or not, `JINA_OPTOUT_TELEMETRY` is always set
        # once `JINA_OPTOUT_TELEMETRY` is always set, the telemetry is disabled
        # https://github.com/jina-ai/jina/blob/030a1e90ce5289c20c250084320a5ba1302da21e/jina/helper.py#L1704
        print(f'JINA_OPTOUT_TELEMETRY: {os.environ.get("JINA_OPTOUT_TELEMETRY", None)}')

# This is set in https://github.com/jina-ai/now/blob/ccb3f5ccbaa379b0abba2b82e4b0b92ffe1ded1c/tests/integration/test_end_to_end.py#L153
# os.environ['NOW_CI_RUN'] = 'True'
f = Flow.load_config(
'''
jtype: Flow
version: 1
with:
  env:
    JINA_OPTOUT_TELEMETRY: ${{ ENV.NOW_CI_RUN }}
executors:
  - uses: MyExecutor
'''
)
with f:
    f.post(on='/hello', inputs=DocumentArray.empty(1))
```

The output is 
```output
JINA_OPTOUT_TELEMETRY: $NOW_CI_RUN
hello
```

florian-hoenicke:
Thanks @nan-wang that was helpful. You are right. I will make a change.

PR #701, opened 2022-11-24
fix: playground bug

azayz:


makram93 (approved):
LGTM! Just remove the reference of JWT_COOKIE

PR #708, opened 2022-11-28
feat: add score breakdown for matches using ElasticIndexer

violenil:


now/executor/indexer/elastic/elastic_indexer.py
```diff
@@ -313,6 +325,9 @@ def _build_es_queries(
                     linear_weight,
                 ) in scores.get_scores(encoder):
                     field_doc = getattr(doc, query_field)
+                    docs[doc.id].tags['embeddings'][
+                        f'{query_field}-{document_encoder}'
+                    ] = field_doc.embedding
```
maximilianwerk, line 330: Do we only need this, when we have `get_score_breakdown == True`?

now/executor/indexer/elastic/elastic_indexer.py
```diff
@@ -449,5 +468,48 @@ def _transform_es_results_to_matches(self, es_results: List[Dict]) -> DocumentAr
         for result in es_results:
             d = self._transform_es_to_da(result)[0]
             d.scores[self.metric] = NamedScore(value=result['_score'])
```
maximilianwerk, line 470: That should become the `else` clause.

now/executor/indexer/elastic/elastic_indexer.py
```diff
@@ -449,5 +468,48 @@ def _transform_es_results_to_matches(self, es_results: List[Dict]) -> DocumentAr
...
+        """
+        Calculate the score breakdown for a given retrieved document. Each SemanticScore in the indexers
+        `default_semantic_scores` should have a corresponding value, returned inside a list of scores in the documents
+        tags under `score_breakdown`.
+
+        :param query_doc: The query document. Contains embeddings for the semantic score calculation at tag level.
+        :param retrieved_results: The Elasticsearch results, containing embeddings inside the `_source` field.
+        :return: List of integers representing the score breakdown.
+        """
+        for semantic_score in self.default_semantic_scores:
```
maximilianwerk, line 488: It is a matter of taste, but you can unpack the semantic score directly here. That's why it is a tuple: To allow nice unpacking. It makes the rest of the function better readable.

maximilianwerk (commented):
looks good. minor improvements suggested.

now/executor/indexer/elastic/elastic_indexer.py
violenil: I think it's still OK to have the total score returned under `self.metric` even when we want to see a score breakdown.

now/executor/indexer/elastic/elastic_indexer.py
violenil: True!

now/executor/indexer/elastic/elastic_indexer.py
```diff
@@ -439,15 +463,75 @@ def _get_bm25_fields(self, doc: Document):
...
+                )
+            ] = NamedScore(value=score)
+
+        # calculate bm25 score
+        vector_total = sum(
+            [v.value for k, v in retrieved_doc.scores.items() if k != self.metric]
+        )
+        overall_score = retrieved_doc.scores[self.metric].value
+        bm25_score = overall_score - vector_total
+        retrieved_doc.scores['bm25'] = NamedScore(value=bm25_score)
```
joschkabraun, line 533: as discussed in the pairing, we should also return the 'unnormalized' bm25 scores

joschkabraun: `bm25_raw` and `bm25_normalized`

PR #709, opened 2022-11-28
Fix top k

makram93:


violenil (approved):
LGTM!

deployment/playground/playground.py
```diff
@@ -265,7 +269,7 @@ def _do_logout():
     st.session_state.avatar_val = None
     st.session_state.token_val = None
     st.session_state.login = True
-    response = requests.post(
+    requests.post(
```
azayz, line 272: we don't use the response ?

azayz (approved):

deployment/playground/playground.py
makram93: No, it was unused hence I removed it

PR #712, opened 2022-11-28
feat: add schema functions

azayz:


now/common/options.py
```diff
@@ -42,6 +44,93 @@ def _create_app_from_output_modality(user_input: UserInput, **kwargs):
     user_input.app_instance = construct_app(app_name)
 
 
+def _get_schema_docarray(user_input: UserInput, **kwargs):
```
florian-hoenicke, line 47: I think the schema functionality should be in a separate file because it is not directly related to options.

now/common/options.py
```diff
@@ -42,6 +44,93 @@ def _create_app_from_output_modality(user_input: UserInput, **kwargs):
...
+    else:
+        client = Client(token=user_input.jwt['token'])
+
+    resp = client.get_artifact_info(name='test_subset_laion')
+    if resp.json()['code'] == 200:
+        field_names = resp.json()['data']['metaData']['summary'][3]['value']
+        for el in ['embedding', 'id', 'mime_type']:
+            if el in field_names:
+                field_names.remove(el)
+        user_input.field_names = field_names
```
florian-hoenicke, line 61: I think it is better to always create a fresh list instead of removing elements.
Something like this:
```
ignored_fieldnames = ['embedding', 'id', 'mime_type']
user_input.field_names = [field_name for field_name in field_names if field_name not in ignored_fieldnames]
```

now/common/options.py
```diff
@@ -42,6 +44,93 @@ def _create_app_from_output_modality(user_input: UserInput, **kwargs):
...
+
+    if user_input.jwt is None:
+        client = Client()
+    else:
+        client = Client(token=user_input.jwt['token'])
+
+    resp = client.get_artifact_info(name='test_subset_laion')
+    if resp.json()['code'] == 200:
+        field_names = resp.json()['data']['metaData']['summary'][3]['value']
+        for el in ['embedding', 'id', 'mime_type']:
```
florian-hoenicke, line 58: Why is it required to filter out these names?

florian-hoenicke (changes requested):
A great feature you are implementing. Here are my first comments:

now/common/options.py
azayz: not really required but the user won't want embeddings , or document id or mime type as a search field

now/common/detect_schema.py
```diff
@@ -0,0 +1,107 @@
...
+    path = os.path.abspath(path)
+    root = os.path.abspath(root)
+    return os.path.relpath(path, root).count(os.path.sep) == 1
+
+
+def _get_schema_local_folder(user_input: UserInput, **kwargs):
+
+    dataset_path = user_input.dataset_path.strip()
+    if os.path.isfile(dataset_path):
+        return []
```
florian-hoenicke, line 80: Why do we return an empty list in that case?

florian-hoenicke: Since the function is get schema local folder, why would it be a file?

now/common/detect_schema.py
```diff
@@ -0,0 +1,107 @@
...
+        if all_files:
+            user_input.field_names = []
+        else:
+            first_path = ''
+            for path in sorted(glob.glob(os.path.join(dataset_path, '**/**'))):
+                if not first_path:
+                    first_path = path
+
+                if not check_path(path, dataset_path):
+                    raise ValueError(
```
florian-hoenicke, line 95: you can provide a link to now.jina.ai

now/common/detect_schema.py
```diff
@@ -0,0 +1,107 @@
...
+    if os.path.isfile(dataset_path):
+        return []
+    elif os.path.isdir(dataset_path):
+        all_files = True
+        for file_or_directory in os.listdir(dataset_path):
+            if not os.path.isfile(os.path.join(dataset_path, file_or_directory)):
+                all_files = False
+        if all_files:
+            user_input.field_names = []
+        else:
```
florian-hoenicke, line 88: I think the whole logic here becomes too complicated. 
Do you see a way to simplify it?
For instance by breaking it down into smaller functions

now/data_loading/utils.py
```diff
@@ -55,3 +56,24 @@ def get_dataset_url(dataset: str, output_modality: Modalities) -> str:
         return f'{BASE_STORAGE_URL}/{data_folder}/{dataset}.{model_name}-{docarray_version}.bin'
     else:
         return f'{BASE_STORAGE_URL}/{data_folder}/{dataset}-{docarray_version}.bin'
+
+
+def _get_s3_bucket_and_folder_prefix(user_input: UserInput):
```
florian-hoenicke, line 61: I think we already have the code for working with S3 in our code base. I think you should find and use the existing functions.

tests/unit/test_detect_schema.py
```diff
@@ -0,0 +1,53 @@
+import os
+
+from now.common.detect_schema import _get_schema_local_folder, _get_schema_s3_bucket
+from now.now_dataclasses import UserInput
+
+
+def test_get_schema_local_folder(gif_resource_path):
```
florian-hoenicke, line 7: I think you should use parameterize to remove the redundancy in these tests

florian-hoenicke (changes requested):

now/data_loading/utils.py
azayz: yes I actually did this as a refactoring so now we call this function

now/common/detect_schema.py
```diff
@@ -0,0 +1,110 @@
...
+        json=json_data,
+    )
+    if response.json()['code'] == 200:
+        response = requests.get(response.json()['data']['download'])
+        ignored_fieldnames = ['embedding', 'id', 'mimeType', 'tags']
+        field_names = [el for el in response.json()[0] if el not in ignored_fieldnames]
+        field_names.extend(list(response.json()[0]['tags']['fields'].keys()))
+        user_input.field_names = field_names
+    else:
+        raise ValueError('DocumentArray doesnt exist or you dont have access to it')
```
violenil, line 31:
```suggestion
        raise ValueError('DocumentArray does not exist or you do not have access to it')
```

now/common/detect_schema.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+def _get_schema_s3_bucket(user_input: UserInput, **kwargs):
+    bucket, folder_prefix = _get_s3_bucket_and_folder_prefix(
+        user_input
+    )  # user has to provide the folder where folder structure begins
```
violenil, line 37: What do you mean by 'where folder structure begins'? Is it kind of like the base directory?

now/common/detect_schema.py
```diff
@@ -0,0 +1,110 @@
...
+
+    field_names = []
+
+    all_files = True
```
violenil, line 41: Would it not make sense to set `all_files` to `False` to begin with, and set it to `True` once your conditions are met? I.e. you have found all files?

now/common/detect_schema.py
```diff
@@ -0,0 +1,110 @@
...
+        response = requests.get(response.json()['data']['download'])
+        ignored_fieldnames = ['embedding', 'id', 'mimeType', 'tags']
+        field_names = [el for el in response.json()[0] if el not in ignored_fieldnames]
+        field_names.extend(list(response.json()[0]['tags']['fields'].keys()))
+        user_input.field_names = field_names
+    else:
+        raise ValueError('DocumentArray doesnt exist or you dont have access to it')
+
+
+def _get_schema_s3_bucket(user_input: UserInput, **kwargs):
```
violenil, line 34: Can you please add docstrings for these functions? Especially since they will be imported in other files (and remove leading underscore)

now/common/detect_schema.py
```diff
@@ -0,0 +1,110 @@
...
+        user_input.field_names = []
+    else:
+        for obj in list(bucket.objects.filter(Prefix=folder_prefix))[
+            1:
+        ]:  # first is the bucket path
+            if obj.key.endswith('/'):
+                continue
+            if len(obj.key.split('/')) - len(folder_prefix.split('/')) != 1:
+                raise ValueError(
+                    'File format different than expected, please check documentation.'
```
violenil, line 57: Link to documentation? Or anything that might be more helpful.

now/common/detect_schema.py
```diff
@@ -0,0 +1,110 @@
...
+        ]:
+            if field.key.endswith('.json'):
+                data = json.loads(field.get()['Body'].read())
+                field_names.extend(list(data.keys()))
+            else:
+                field_names.append(field.key.split('/')[-1])
+        user_input.field_names = field_names
+
+
+def check_path(path, root):
```
violenil, line 74: This function, for example, is only used here in `detect_schema.py` so this one can have a leading underscore, if you like.

now/common/detect_schema.py
```diff
@@ -0,0 +1,110 @@
...
+    path = os.path.abspath(path)
+    root = os.path.abspath(root)
+    return os.path.relpath(path, root).count(os.path.sep) == 1
+
+
+def _get_schema_local_folder(user_input: UserInput, **kwargs):
+    dataset_path = user_input.dataset_path.strip()
+    if os.path.isfile(dataset_path):
+        return []
```
violenil, line 83:
```suggestion
        user_input.field_names = []
```
This function should not return anything, right? The signature shows that it doesn't, and all calls of this function don't expect any return values.

now/common/options.py
```diff
@@ -107,35 +121,37 @@ def _create_app_from_output_modality(user_input: UserInput, **kwargs):
...
+    post_func=lambda user_input, **kwargs: _infer_app_type_from_demo_data(user_input),
 )
 
 
 def _get_demo_data_choices(user_input: UserInput):
-    if user_input.output_modality:
-        ds = user_input.app_instance.demo_datasets[user_input.output_modality]
-    else:
-        ds = user_input.app_instance.demo_datasets
+
```
violenil, line 129:
```suggestion
```

now/common/options.py
```diff
@@ -107,35 +121,37 @@ def _create_app_from_output_modality(user_input: UserInput, **kwargs):
...
+        for demo_data in all_demo_datasets
     ]
 
 
+def _infer_app_type_from_demo_data(user_input):
+    for modality, demos in AVAILABLE_DATASET.items():
+        for demo in demos:
+            if user_input.dataset_name == demo.name:
+                user_input.output_modality = modality
+    _create_app_from_output_modality(user_input)
```
violenil, line 144: 🥳

violenil (commented):
Looks good! I am so excited about this, big changes coming! I did leave some comments. Please can you write docstrings for the functions that are imported outside of the file they're defined in, and remove their leading underscore? I think it is better to have this leading underscore on internal functions, that are not used outside of a class.
Also, please could you write an informative description for the PR? It will make it easier to understand what was done here (and this is an important PR :) )
Good work!

now/common/options.py
```diff
@@ -38,7 +44,7 @@
~    elif user_input.output_modality == 'video':
~        app_name = Apps.TEXT_TO_VIDEO
~    else:
~        raise ValueError(f'Invalid output modality: {user_input.output_modality}')
~    user_input.app_instance = construct_app(app_name)
```
joschkabraun: This looks like fixing a symptom. In which cases is app_instance None? Also, can't we remove output modality question now given that we ask for search fields?

now/common/options.py
```diff
@@ -181,11 +198,12 @@
     depends_on=DATASET_TYPE,
     conditional_check=lambda user_input: user_input.dataset_type
     == DatasetTypes.S3_BUCKET,
+    post_func=lambda user_input, **kwargs: _get_schema_s3_bucket(user_input),
```
joschkabraun, line 201: we can remove the commented code (below)

now/common/options.py
```diff
@@ -199,9 +217,50 @@
...
+    choices=lambda user_input, **kwargs: _get_fields(user_input),
+    prompt_message='Please select the search fields:',
+    prompt_type='checkbox',
+    depends_on=DATASET_TYPE,
+    conditional_check=lambda user_input: user_input.field_names is not None
+    and len(user_input.field_names) > 0
+    and user_input.dataset_type != DatasetTypes.DEMO,
+    post_func=lambda user_input, **kwargs: _exclude_search_fields(user_input),
+)
+
```
joschkabraun, line 234: Dynamically modifying `user_input.field_names` looks dangerous to me as this isn't clear. I think it's better to let the function returning the choices, simply exclude the search fields.

joschkabraun (changes requested):
Nice refactoring in dialog! I think we can push it even further.

now/common/detect_schema.py
azayz: you are right, Ill change this

now/common/detect_schema.py
azayz: I think it's same logic, I believe it's easier to assume something is true and find one example that shows its false than the other way around

now/common/options.py
azayz: I tried using what we discussed during pairing but the code looked redundant for getting field names and removing the search fields can't be done in one line so I would rather use functions, however I changed the function naming to indicate more what we are doing

tests/unit/test_detect_schema.py
```diff
@@ -0,0 +1,110 @@
...
+def get_aws_info():
+    dataset_path = os.environ.get('S3_SCHEMA_FOLDER_PATH')
+    aws_access_key_id = os.environ.get('AWS_ACCESS_KEY_ID')
+    aws_secret_access_key = os.environ.get('AWS_SECRET_ACCESS_KEY')
+    region = 'eu-west-1'
+
+    return dataset_path, aws_access_key_id, aws_secret_access_key, region
+
+
+def test_set_field_names_from_local_folder(gif_resource_path):
```
florian-hoenicke, line 25: please use pytest fixtures for the tests to reduce the code redundancy

now/common/detect_schema.py
```diff
@@ -0,0 +1,224 @@
...
+
+    :param objects: list of objects in the bucket
+    """
+    for obj in objects[1:]:
+        if obj.key.endswith('/'):
+            return False
+    return True
+
+
+def _check_folder_structure_s3_bucket(objects, folder_prefix):
```
florian-hoenicke, line 80: well done creating separated function making it easy to follow

joschkabraun (changes requested):

now/common/detect_schema.py
```diff
@@ -0,0 +1,224 @@
...
+    response = requests.post(
+        'https://api.hubble.jina.ai/v2/rpc/docarray.getFirstDocuments',
+        cookies=cookies,
+        json=json_data,
+    )
+    if response.json()['code'] == 200:
+        response = requests.get(response.json()['data']['download'])
+        ignored_fieldnames = ['embedding', 'id', 'mimeType', 'tags']
+        field_names = [el for el in response.json()[0] if el not in ignored_fieldnames]
+        field_names.extend(list(response.json()[0]['tags']['fields'].keys()))
```
makram93, line 61: Add a check here for `tags`. if they are present and not empty

makram93 (changes requested):

now/now_dataclasses.py
```diff
@@ -103,7 +103,12 @@ class UserInput(BaseModel):
     aws_region_name: Optional[str] = None
 
     # Fields
+    field_names: Optional[List] = None
```
joschkabraun, line 106: when searching for this field, it's only written into to then read again from it, please remove it

now/common/options.py
```diff
@@ -181,25 +205,40 @@ def _get_demo_data_choices(user_input: UserInput):
...
+    choices=lambda user_input, **kwargs: [
+        {'name': field, 'value': field} for field in user_input.search_fields_candidates
+    ],
+    prompt_message='Please select the search fields:',
+    prompt_type='checkbox',
     depends_on=DATASET_TYPE,
-    is_terminal_command=True,
-    conditional_check=lambda user_input: user_input.dataset_type != DatasetTypes.DEMO,
-    post_func=lambda user_input, **kwargs: _parse_search_fields(user_input),
+    conditional_check=lambda user_input: user_input.field_names is not None
```
joschkabraun, line 222: this depends on `user_input.search_field_candidates`; it's also the question what happens if there are no search field candidates: is that an illegal state?

now/common/options.py
```diff
@@ -181,25 +205,40 @@ def _get_demo_data_choices(user_input: UserInput):
...
+    name='filter_fields',
+    choices=lambda user_input, **kwargs: [
+        {'name': field, 'value': field}
+        for field in user_input.filter_fields_candidates
+        if field not in user_input.search_fields
+    ],
+    prompt_message='Please select the filter fields',
+    prompt_type='checkbox',
+    depends_on=DATASET_TYPE,
+    conditional_check=lambda user_input: user_input.field_names is not None
```
joschkabraun, line 238: depends on `user_input.filter_fields_candidates`

now/common/options.py
azayz: yes that is an illegal state and an error is raised

tests/integration/test_end_to_end.py
```diff
@@ -411,11 +411,12 @@ def test_backend_custom_data(
         'flow_name': 'nowapi',
         'output_modality': 'image',
         'dataset_type': DatasetTypes.S3_BUCKET,
-        'dataset_path': os.environ.get('S3_IMAGE_TEST_DATA_PATH'),
+        'dataset_path': os.environ.get('S3_CUSTOM_DATA_PATH'),
         'aws_access_key_id': os.environ.get('AWS_ACCESS_KEY_ID'),
         'aws_secret_access_key': os.environ.get('AWS_SECRET_ACCESS_KEY'),
         'aws_region_name': 'eu-west-1',
-        'search_fields': 'x, y',
+        'search_fields': [],
```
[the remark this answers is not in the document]
azayz: done

now/now_dataclasses.py
```diff
@@ -103,7 +103,12 @@ class UserInput(BaseModel):
...
     search_fields: Optional[List] = None
+    search_fields_candidates: Optional[Dict] = None
+    search_fields_modalities: Optional[List] = None
```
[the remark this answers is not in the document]
azayz: the modalities dict isn't read in this PR but Kalim needs this search fields to modality mapping for his PR

now/common/options.py
```diff
@@ -181,25 +201,40 @@ def _get_demo_data_choices(user_input: UserInput):
...
+    choices=lambda user_input, **kwargs: [
+        {'name': field, 'value': field}
+        for field in user_input.filter_fields_candidates
+        if field not in user_input.search_fields
+    ],
+    prompt_message='Please select the filter fields',
+    prompt_type='checkbox',
+    depends_on=DATASET_TYPE,
+    conditional_check=lambda user_input: user_input.filter_fields_candidates is not None
+    and len(user_input.filter_fields_candidates) > 0
```
makram93, line 235:
```suggestion
    and len(set(user_input.filter_fields_candidates) - set(user_input.search_fields)) > 0
```

makram93 (approved):
Nice work on extracting the schema. Hopefully, with the introduction of multimodal DocArrays our task will become much simpler.
LGTM!

now/now_dataclasses.py
```diff
@@ -103,7 +103,11 @@ class UserInput(BaseModel):
     aws_region_name: Optional[str] = None
 
     # Fields
-    search_fields: Optional[List] = None
+    search_fields: Optional[List] = []
+    search_fields_candidates: Optional[List] = []
+    search_fields_modalities: Optional[Dict] = {}
+    filter_fields: Optional[List] = []
+    filter_fields_candidates: Optional[List] = []
```
joschkabraun, line 110: as discussed in pairing, we can remove `search_fields_candidates` as we have modalities; similarly, for `filter_fields_candidates`, introduce modalities here; then we can remove redundant code in `common/utils:_extract_tags_for_indexer`

joschkabraun (approved):
LGTM! Minor refactor suggestion

florian-hoenicke (approved):
LGTM

PR #725, opened 2022-12-02
chore: add a PR template

violenil:
closes #731

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -4,3 +4,5 @@
 ---
 
 - [ ] This PR references an open issue
+- [ ] Test added
+- [ ] Documentation adjusted
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,6 @@
+<!--- PR Description here --->
+
+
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,6 @@
+<!--- PR Description here --->
+
+
+---
+
+- [ ] This PR references an open issue
```
florian-hoenicke, line 6: I think it is also fine to have prs without issues. If I just want to clean up something, creating an issue would be an overhead

.github/pull_request_template.md
```diff
@@ -0,0 +1,6 @@
+<!--- PR Description here --->
+
```
florian-hoenicke, line 2: What about a checklist:
- [ ] test added
- [ ] documentation adjusted

.github/pull_request_template.md
violenil: But would it not allow us to more closely monitor the work that is being done, when each PR has an issue on the board, regardless of how small it is? I think it's a better way to keep track.

PR #728, opened 2022-12-05
fix: workaround core wolf

florian-hoenicke:
fixes the issue we currently have with the core/wolf when indexing larger datasets.
Instead of indexing the whole dataset at once, we break it into batches and introduced retries

### DIFF
--- a/now/run_backend.py
+++ b/now/run_backend.py
@@ -9,6 +9,7 @@ from typing import Dict, Optional
 import requests
 from docarray import DocumentArray
 from jina.clients import Client
+from tqdm import tqdm
 
 from now.admin.update_api_keys import update_api_keys
 from now.app.base.app import JinaNOWApp
@@ -120,7 +121,7 @@ def index_docs(user_input, dataset, client):
     """
     Index the data right away
     """
-    print(f"▶ indexing {len(dataset)} documents")
+    print(f"▶ indexing {len(dataset)} documents in batches")
     params = {
         'user_input': user_input.__dict__,
         'access_paths': ACCESS_PATHS,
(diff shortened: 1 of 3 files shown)


now/run_backend.py
```diff
@@ -165,15 +166,28 @@ def call_flow(
...
-        return_results=return_results,
-        continue_on_error=True,
-    )
+
+    # this is a hack for the current core/ wolf issue
+    # since we get errors while indexing, we retry
+    # TODO: remove this once the issue is fixed
+    l = list(dataset.batch(request_size * 100))
+    for j, batch in enumerate(tqdm(l)):
+        for i in range(5):
```
azayz, line 175: I would change the naming of these variables to give more context, for example l could be batches, j could be n_batch and i could be attempt

now/run_backend.py
```diff
@@ -165,15 +166,28 @@ def call_flow(
...
+                    inputs=batch,
+                    show_progress=True,
+                    parameters=parameters,
+                    return_results=return_results,
+                    continue_on_error=True,
+                )
+                break
+            except Exception as e:
+                print(f'batch {j}, try {i}', e)
+                sleep(5 * (i + 1))
```
azayz, line 189: don't you think because of this sleep, indexing will take so much more time

florian-hoenicke: but we just sleep between 5 and 30 seconds and only in case there is an error

florian-hoenicke: But I think we need an exit condition in case it is not working after 5 retries.
Okay and also I think in case of ci we should not have the retry since we want to see the error asap. What do you think?

azayz: I agree for CI, lets see how it also affects indexing time

azayz:
> Great that we can implement a workaround! I see that these changes are only for the text-to-video app, should we make similar changes to other apps or do they not suffer from this issue?

batching during index time is applied to all apps

PR #735, opened 2022-12-06
feat: n-to-n ElasticIndexer multi modal doc

violenil:
---

- [x] Test added
- [x] Documentation adjusted


now/executor/indexer/elastic/es_converter.py
```diff
@@ -0,0 +1,191 @@
...
+        da = DocumentArray()
+        for es_doc in result:
+            doc = DocumentArray.from_base64(es_doc['_source']['serialized_doc'])[0]
+            for k, v in es_doc['_source'].items():
+                if (
+                    k.startswith('embedding') or k.endswith('embedding')
+                ) and get_score_breakdown:
+                    if 'embeddings' not in doc.tags:
+                        doc.tags['embeddings'] = {}
+                    doc.tags['embeddings'][k] = v
```
maximilianwerk, line 100: Didn't we said, the embeddings should be put into the doc before serialization in this way? This ensures, least effort during search.

maximilianwerk (changes requested):
I did not review the tests. Good job!

The only important change for now is the was we handle `apply_default_bm25`, because it is an interface change. The other things can be addressed in a refactoring later.

now/executor/indexer/elastic/es_converter.py
violenil: This is true, we should not have embeddings in the **serialized** docs, which are removed [here](https://github.com/jina-ai/now/blob/feat-elastic-multi-modal-doc/now/executor/indexer/elastic/es_converter.py#L36) in the es_converter before serializing and indexing. In the conversion here, `convert_es_to_da` is loading the serialized doc (has no embeddings) and adds the embeddings it retrieved upon search from the index as tags, on the condition that we want to do a score breakdown. We would need those embeddings then.

Perhaps, we should also use `get_score_breakdown` to define whether the embeddings from ES should be returned upon search (restrict the result from ES). This is not something I have covered yet.

now/executor/indexer/elastic/es_query_builder.py
```diff
@@ -0,0 +1,246 @@
...
+                    'script_score': {
+                        'query': {
+                            'bool': query['bool'],
+                        },
+                        'script': {
+                            'source': sources[doc_id],
+                            'params': script_params[doc_id],
+                        },
+                    },
+                }
```
joschkabraun, line 171: code redundancy: `query_json` in `else` looks like `organic` in `if pinned_queries`

now/executor/indexer/elastic/es_query_builder.py
```diff
@@ -0,0 +1,246 @@
...
+            query['bool']['filter'] = es_search_filter
+
+        return query
+
+    @staticmethod
+    def get_pinned_query(
+        doc: Document, query_to_curated_ids: Dict[str, list] = {}
+    ) -> Dict:
+        pinned_query = {}
+        if getattr(doc, 'query_text', None):
```
joschkabraun, line 218: can we define a constant for the assumption that the query has attribute `query_text`? this will make future code changes easier

now/executor/indexer/qdrant/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM qdrant/qdrant:v0.7.0
+FROM qdrant/qdrant:v0.10.1
```
joschkabraun, line 1: Out of curiosity, was there a need to this change?

now/executor/indexer/elastic/elastic_indexer.py
```diff
@@ -1,92 +1,130 @@
...
         """
-        super().__init__(**kwargs)
-
         self.hosts = hosts
         self.metric = metric
         self.index_name = index_name
+        self.traversal_paths = traversal_paths
+        self.limit = limit
+        self.document_mappings = [FieldEmbedding(*dm) for dm in document_mappings]
+        self.default_semantic_scores = (
```
joschkabraun, line 75: I find the name `default_semantic_score` somewhat confusing. By its name I would think it means the just default values for all semantic scores (e.g. cos. sim. of 0 or so). But then I would also be confused why we need default values (presumably to fill some value in if an error occurs?).
If I correctly understand, `default_semantic_score` means something else, right? If so, can you change the name please.

now/executor/indexer/elastic/elastic_indexer.py
```diff
@@ -96,38 +134,31 @@ def index(
...
+        )
+        es_docs = ESConverter.convert_doc_map_to_es(
+            preprocessed_docs_map, self.index_name, self.encoder_to_fields
+        )
         try:
             success, _ = bulk(self.es, es_docs)
             self.es.indices.refresh(index=self.index_name)
         except Exception as e:
-            print(e)
             print(traceback.format_exc())
```
joschkabraun, line 149: should we change the `print` statements to use `self.logger`?

now/executor/indexer/elastic/elastic_indexer.py
```diff
@@ -1,92 +1,130 @@
...
+            document_mapping.encoder: document_mapping.fields
+            for document_mapping in self.document_mappings
+        }
         self.es_config = {'verify_certs': False} if not es_config else es_config
-        self.dims = dims if isinstance(dims, list) else [dims]
-        self.es_mapping = self._generate_es_mapping(dims)
+        self.es_mapping = (
+            self.generate_es_mapping(self.document_mappings, self.metric)
+            if not es_mapping
+            else es_mapping
```
joschkabraun, line 86: all the code of the form `var = other_val if not val else val` can be written as `var = val or other_val`

now/executor/indexer/elastic/elastic_indexer.py
```diff
@@ -145,38 +176,62 @@ def search(
...
+
+        # search_filter = parameters.get('filter', None)
+        limit = parameters.get('limit', self.limit)
+        get_score_breakdown = parameters.get('get_score_breakdown', False)
+        custom_bm25_query = parameters.get('custom_bm25_query', None)
+        apply_default_bm25 = parameters.get('apply_default_bm25', False)
+        semantic_scores = parameters.get('default_semantic_scores', None)
+        filter = parameters.get('filter', {})
+        if not self.default_semantic_scores:
+            self.default_semantic_scores = (
```
joschkabraun, line 199: here is also:  `var = other_val if not val else val` -> `var = val or other_val`

now/executor/indexer/elastic/elastic_indexer.py
```diff
@@ -197,246 +252,158 @@ def list(self, parameters: dict = {}, **kwargs):
...
-        Delete endpoint to delete document/documents from the index.
+        Endpoint to delete documents from an index. Either delete documents by filter condition
+        or by specifying a list of document IDs.
 
-        :param parameters: dictionary with filter conditions to select
+        :param parameters: dictionary with filter conditions or list of IDs to select
             documents for deletion.
         """
         search_filter = parameters.get('filter', None)
+        ids = parameters.get('ids', None)
```
joschkabraun, line 278: is that change reflected in docs or somewhere else on a higher level than this endpoint?

joschkabraun (changes requested):
Nice changes to the ElasticIndexer!
A few comments below.
Also, will tags from OCR detection be incorporated in this PR or in a follow-up?

now/executor/indexer/elastic/es_converter.py
```diff
@@ -0,0 +1,192 @@
...
+        :param index_name: name of the index to be used in Elasticsearch.
+        :param encoder_to_fields: dictionary mapping encoder to fields.
+        :return: a list of Elasticsearch documents as dictionaries ready to be indexed.
+        """
+        es_docs = {}
+        for executor_name, documents in docs_map.items():
+            for doc in documents:
+                if doc.id not in es_docs:
+                    es_docs[doc.id] = self.get_base_es_doc(doc, index_name)
+                    _doc = DocumentArray(Document(doc, copy=True))
```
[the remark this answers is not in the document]
violenil: Oh strange, I remember not being able to `to_base64` on a Document. Will try it. Thanks👍🏻

now/executor/indexer/elastic/es_preprocessor.py
```diff
@@ -0,0 +1,55 @@
...
+                        ] = field_doc[
+                            ..., 'embedding'
+                        ]  # will be a matrix, stacked
+                        # replace ChunkArray with Document
+                        setattr(doc, field_name, new_doc)
+
+        return docs_map
+
+    @staticmethod
+    def average_embeddings_of_subdocuments(field_doc):
```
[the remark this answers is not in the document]
violenil: I wanted to allow in future other methods to be used instead of simple 'averaging', that is why I made it a separate function.

now/executor/indexer/elastic/es_preprocessor.py
```diff
@@ -0,0 +1,55 @@
...
+                        # average the embeddings of the subdocuments
+                        # TODO: use a better way to aggregate embeddings (eg. k-means)
+                        embedding = self.average_embeddings_of_subdocuments(field_doc)
+                        new_doc = Document(
+                            text=' '.join(
+                                field_doc.texts
+                            ).strip(),  # concatenate texts of subdocuments
+                            embedding=embedding,
+                        )
+                        if field_doc[0].uri:
```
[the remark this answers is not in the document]
violenil: Yes, we make this assumption for now. Should support Videos (same uri for all frames). If there are multiple images for a document with different URI's, they will each need to be a separate field.

now/executor/indexer/elastic/es_converter.py
```diff
@@ -0,0 +1,192 @@
...
+                        es_doc['text'] = field_doc.text
+                    if hasattr(field_doc, 'uri') and field_doc.uri:
+                        es_doc['uri'] = field_doc.uri
+        return list(es_docs.values())
+
+    def get_embedding(self, field_doc: Union[ChunkArray, Document]) -> np.array:
+        if isinstance(field_doc, ChunkArray):
+            # average the embeddings of the chunks
+            # can be changed to first, last, etc. in future
+            embedding = self.average_embeddings_of_subdocuments(field_doc)
```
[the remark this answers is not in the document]
violenil: It does, it is a staticmethod (the one-liner 😉)

now/executor/indexer/elastic/es_query_builder.py
violenil: True. I just wanted to be as explicit as possible, since it is really easy to loose oversight in this query building process. I can however simplify this, you are right!

now/executor/indexer/elastic/es_query_builder.py
violenil: Yes, good idea. This is what I meant when talking about standardizing the format of our query Document from `@dataclass`. I will make this a constant.

now/executor/indexer/qdrant/Dockerfile
violenil: Yes, there is a need. In order to use multimodal document from docarray, we need docarray version `0.19.0`, which in turn needs a qdrant-client with version `0.10.1`, so these docker images needed to be updated.

now/executor/indexer/elastic/elastic_indexer.py
violenil: This `default_semantic_score` defines how Document scores (eg. cosine similarity between two fields) should be calculated. For example: 

```
default_semantic_score = [
        SemanticScore('query_text', 'title', 'clip', 1),
        SemanticScore('query_text', 'gif', 'clip', 1),
        SemanticScore('query_text', 'title', 'sbert', 1),
        SemanticScore('query_text', 'excerpt', 'sbert', 3),
        SemanticScore('query_text', 'my_bm25_query', 'bm25', 1),
    ]
```
Where a `SemanticScore` consists of 1. field from query, 2. field from document, 3. encoder and 4. weighting of this component.

If this isn't passed when searching, the `default_semantic_score` will be automatically generated as a cosine similarity between all possible fields encoded in the same vector space with a weighting of 1.

With this definition, I think the name `default_semantic_score` is suitable. Can you propose a better name? Perhaps  `default_semantic_score_combination`?

now/executor/indexer/elastic/elastic_indexer.py
violenil: Ah good point! I will change this. Thanks!

now/executor/indexer/elastic/elastic_indexer.py
violenil: Hmm I think this is not reflected, though it may be a new feature as far as I know. Will document it 👍🏻

violenil:
| Also, will tags from OCR detection be incorporated in this PR or in a follow-up?

I think it makes sense to incorporate this into a separate PR. In the executor tests for the NOWElasticIndexer I have also marked `test_search_chunk_using_sum_ranker` to skip, since ranking + OCR is not yet integrated. I will create a follow-up ticket for this.

now/executor/indexer/elastic/es_converter.py
```diff
@@ -0,0 +1,181 @@
...
+                if (
+                    k.startswith('embedding') or k.endswith('embedding')
+                ) and get_score_breakdown:
+                    if 'embeddings' not in doc.tags:
+                        doc.tags['embeddings'] = {}
+                    doc.tags['embeddings'][k] = v
+            da.append(doc)
+        return da
+
+    def convert_es_results_to_matches(
```
florian-hoenicke, line 94: Is this function used?

tests/executor/indexer/elastic/test_es_converter.py
```diff
@@ -0,0 +1,122 @@
...
+def test_convert_es_to_da(es_inputs):
+    """
+    This test should check whether retrieved elasticsearch documents are
+    correctly transformed back into Documents.
+    TODO: fill
+    """
+    pass
+
+
+def test_convert_es_results_to_matches(es_inputs):
```
florian-hoenicke, line 67: If the related function is not used and the test is not implement, we should remove it and rather create a followup ticket to implement this.

florian-hoenicke (changes requested):
What do you think about the following comments?

now/executor/indexer/elastic/es_converter.py
```diff
@@ -0,0 +1,181 @@
...
+
+        :param es_results: List of dictionaries containing results from Elasticsearch querying.
+        :param get_score_breakdown: whether to calculate the score breakdown for matches.
+        :return: `DocumentArray` that holds all matches in the form of `Document`s.
+        """
+        matches = DocumentArray()
+        for result in es_results:
+            d = self.convert_es_to_da(result, get_score_breakdown)[0]
+            d.scores[metric] = NamedScore(value=result['_score'])
+            if get_score_breakdown:
```
florian-hoenicke, line 113: Do we use the score breakdown?

now/executor/indexer/elastic/es_converter.py
```diff
@@ -0,0 +1,181 @@
...
+            encoder,
+            linear_weight,
+        ) in semantic_scores:
+            if encoder == 'bm25':
+                continue
+            q_emb = query_doc.tags['embeddings'][f'{query_field}-{encoder}']
+            d_emb = retrieved_doc.tags['embeddings'][
+                f'{document_field}-{encoder}.embedding'
+            ]
+            if metric == 'cosine':
```
florian-hoenicke, line 147: The score calculation could deserve a separate function

now/executor/indexer/elastic/es_preprocessor.py
```diff
@@ -0,0 +1,55 @@
...
+import numpy as np
+from docarray import Document, DocumentArray
+from docarray.array.chunk import ChunkArray
+
+
+class ESPreprocessor:
+    def __init__(self):
+        pass
+
+    def preprocess_docs_map(
```
florian-hoenicke, line 12: I think we should have a different name to preprocessing since we already have that term in use.
What about merge_docs_map?

now/executor/indexer/elastic/es_preprocessor.py
```diff
@@ -0,0 +1,55 @@
...
+        for executor_name, da in docs_map.items():
+            field_names = encoder_to_fields[executor_name]
+            for doc in da:
+                for field_name in field_names:
+                    field_doc = getattr(doc, field_name)
+                    if isinstance(field_doc, ChunkArray):
+                        # average the embeddings of the subdocuments
+                        # TODO: use a better way to aggregate embeddings (eg. k-means)
+                        embedding = self.average_embeddings_of_subdocuments(field_doc)
+                        new_doc = Document(
```
florian-hoenicke, line 34: merged_doc instead of new_doc?

now/executor/indexer/elastic/es_query_building.py
```diff
@@ -0,0 +1,228 @@
...
+def generate_semantic_scores(
+    docs_map: Dict[str, DocumentArray],
+    encoder_to_fields: Dict[str, Union[List[str], str]],
+) -> List[SemanticScore]:
+    """Generate semantic scores from document mappings."""
+    semantic_scores = []
+    # either take fields names from _metadata, if multimodal doc
+    # or take modality from tags of root doc
+    for executor_name, da in docs_map.items():
+        if executor_name == 'preprocessor':
```
florian-hoenicke, line 30: Since we don't do the skip connection anymore, this can be removed.

florian-hoenicke: Or what is your opinion?

florian-hoenicke (changes requested):
What do you think about the following comments?

now/executor/indexer/elastic/es_converter.py
```diff
@@ -0,0 +1,181 @@
...
+        :return: a DocumentArray containing all results.
+        """
+        if isinstance(result, Dict):
+            result = [result]
+        da = DocumentArray()
+        for es_doc in result:
+            doc = Document.from_base64(es_doc['_source']['serialized_doc'])
+            for k, v in es_doc['_source'].items():
+                if (
+                    k.startswith('embedding') or k.endswith('embedding')
```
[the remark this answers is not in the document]
violenil: Because in the elasticsearch index the embeddings for each field are stored under "{field_name}-{encoder}-embedding"

now/executor/indexer/elastic/es_converter.py
violenil: Yes.

tests/executor/indexer/elastic/test_es_converter.py
violenil: This function is used.

now/executor/indexer/elastic/es_converter.py
violenil: Being able to get a score breakdown is a feature that was requested as part of this refactoring.

now/executor/indexer/elastic/es_query_building.py
violenil: True!

.github/workflows/ci.yml
```diff
@@ -298,7 +298,7 @@ jobs:
           python -m pip install wheel
           pip install --no-cache-dir ".[full,test]"
       - name: Test executors separately
-        timeout-minutes: 10
+        timeout-minutes: 15
```
florian-hoenicke, line 301: I think, we should investigate why it takes now longer

now/executor/indexer/elastic/es_preprocessor.py
violenil: I would stick with the name 'preprocessing', since we don't actually merge the `docs_map` here. The function still returns a `docs_map`. We are just preparing the docs_map for indexing.

.github/workflows/ci.yml
violenil: As far as I can see, the test for executors now take 11:14 minutes. The tests I am adding to this section take a total of 32 seconds. I think this may have just pushed the session timeout over the edge. However, I don't know if there is a way to reduce the time it take for the Elasticsearch tests to run - it is already so quick.

now/executor/indexer/elastic/es_query_building.py
```diff
@@ -0,0 +1,225 @@
...
+        'linear_weight',
+    ],
+)
+metrics_mapping = {
+    'cosine': 'cosineSimilarity',
+    'l2_norm': 'l2norm',
+}
+
+
+def generate_semantic_scores(
```
jupyterjazz, line 21: I remember you showed me a library for query building. Decided not to use it?

now/executor/indexer/elastic/es_preprocessing.py
```diff
@@ -0,0 +1,50 @@
+from typing import Dict
+
+import numpy as np
+from docarray import Document, DocumentArray
+from docarray.array.chunk import ChunkArray
+
+
+def merge_subdocuments(docs_map, encoder_to_fields) -> Dict[str, DocumentArray]:
```
jupyterjazz, line 8: Just a small comment, I was wondering if it makes sense for this function to be part of the `ESConverter` because it seems like `merge_subdocuments` will only be used during converting, and the separate file might be a bit confusing?

jupyterjazz, line 8: Just a small comment, I was wondering if it makes sense for this function to be part of the `ESConverter` because it seems like `merge_subdocuments` will only be used during converting, and the separate file might be a bit confusing?

now/executor/indexer/elastic/es_query_building.py
```diff
@@ -0,0 +1,225 @@
...
+
+def get_scores(encoder, semantic_scores):
+    for (
+        query_field,
+        document_field,
+        _encoder,
+        linear_weight,
+    ) in semantic_scores:
+        if encoder == _encoder:
+            yield query_field, document_field, _encoder, linear_weight
```
jupyterjazz, line 225: another small comment: looks like we introduced this function for only one if statement. Is this true or am I missing something? If so, do we need it?

jupyterjazz (commented):
Looks great! :rocket:

added few comments

now/executor/indexer/elastic/es_preprocessing.py
violenil: @maximilianwerk  insisted on this separation of 'preprocessing' documents, and conversion into elasticsearch format. I think this is a good way to go for now.

now/executor/indexer/elastic/es_query_building.py
violenil: I anticipated that we might need to `get_scores` in other places in the code, whenever we need to see what semantic scores are relevant for a particular encoder. It may be overkill, but also makes the code more readable.

now/executor/indexer/elastic/es_converter.py
```diff
@@ -0,0 +1,185 @@
...
+        pass
+
+    def convert_doc_map_to_es(
+        self,
+        docs_map: Dict[str, DocumentArray],
+        index_name: str,
+        encoder_to_fields: dict,
+    ) -> List[Dict]:
+        """
+        Transform a dictionary (mapping encoder to DocumentArray) into a list of Elasticsearch documents.
```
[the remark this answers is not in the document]
violenil: Done.

florian-hoenicke (approved):
Besides the qdrant version update, it should not have any impact on the current deployments, I consider it save to merge.

joschkabraun (approved):
LGTM!

PR #741, opened 2022-12-07
feat: dynamic flow

makram93:
This PR addresses automatic flow generation. Points to note:
- Currently, there's no single search app, hence the logic for flow generation is added to the `base/app.py` to avoid duplication across apps (image-text or video app, etc.)
- In the second iteration, this will be refactored so that base/app.py only combines the executors stub provided by any app extending the base class.

---

- [ ] Test added
- [ ] Documentation adjusted


joschkabraun:
I think with this PR we should also remove all remaining apps into a single app as we generate the flow automatically (in regards to the PR description). We can later think about how to restructure this into a `BaseApp` and into a `SearchApp`.

makram93:
Git commits messed up. Hence deleting this and creating another PR

PR #747, opened 2022-12-09
test: revert quality to 224x224

florian-hoenicke:
- [x] deployment is 2x to 4x faster
- [ ] quality needs to be checked


tests/executor/preprocessor/test_unit.py
```diff
@@ -23,7 +23,7 @@ def download_mock(url, destfile):
         (
             'image/6785325056.jpg',
             1,
-            'Richard Branson TotallyLooksLike.com Zaphod Beeblebrox',
+            'RichavdE',  # with larger resolution it would be the following text: 'Richard Branson TotallyLooksLike.com Zaphod Beeblebrox',
```
makram93, line 26: But isn't this harmful? Or are we just concerned with finding any text and not full text?

joschkabraun:
Which queries did you try? Did you document the tested queries somewhere?

tests/executor/preprocessor/test_unit.py
florian-hoenicke: yes, just what text it is at the moment

florian-hoenicke:
> Which queries did you try? Did you document the tested queries somewhere?

Just a few queries we had before. You are right, it is not statistically significant. "girl on a motorbike" and "campfire".
I do a side by side comparison before.

PR #754, opened 2022-12-13
fix: reversed text and image modalities

IyadhKhalfallah:
Text and image modalities were showing in a reversed order in playground. Reversing the order for these was the solution.


azayz (changes requested):
please submit a screenshot on the effect of the change on the playground and also rename the branch to a suitable name with more details, since this is a fix it should start with fix-

jupyterjazz (approved):
Great work! :rocket:

Agree with Aziz's comment about the branch name, and also would be nice to move the screenshot you put in the issue to this PR

PR #761, opened 2022-12-13
feat: migrate demo ds

makram93:
- [x] The sample datasets loaded in playground has been  uploaded on s3 since to access Hubble playground beanstalk app will need token
- [x] datasets hosted on s3 - to be on the safe side (public URL)
- [x] optionally host on Docarray push pull if it never gets deleted
- [x] datasets should not be deleted at some point
- [x] datasets are on latest version
- [x] datasets are multi-modal docs
- [x] Adjust the dialog options for demo datasets 
- [ ] no datasets in GCP 

Playground:
image

image

---

- [x] Test added
- [ ] Documentation adjusted


now/common/detect_schema.py
```diff
@@ -100,15 +105,11 @@ def set_field_names_from_docarray(user_input: UserInput, **kwargs):
...
     )
     if response.json()['code'] == 200:
-        field_names = _extract_field_names_docarray(response)
+        search_modalities, filter_modalities = _extract_field_names_docarray(response)
     else:
         raise ValueError('DocumentArray does not exist or you do not have access to it')
-    (
-        search_fields_modalities,
-        filter_fields_modalities,
-    ) = _create_candidate_search_filter_fields(field_names)
```
joschkabraun, line 109: why don't we use `_create_candidate_search_filter_fields` to restrict which fields are selectable as filter fields?

now/common/options.py
```diff
@@ -102,7 +102,10 @@ def _create_app_from_output_modality(user_input: UserInput, **kwargs):
 
 
 def check_login_dataset(user_input: UserInput):
-    if user_input.dataset_type == DatasetTypes.DOCARRAY and user_input.jwt is None:
+    if (
+        user_input.dataset_type in [DatasetTypes.DEMO, DatasetTypes.DOCARRAY]
```
joschkabraun, line 106: Does that mean for even demo datasets we need now Jina login? I can imagine this being a bit annoying for a user wanting to play around with it.

makram93: If the datasets are hosted on Hubble then yes the login will be needed. Alternatively, we can move the datasets to S3 bucket to avoid this

now/common/detect_schema.py
makram93: Because the current `_create_candidate_search_filter_fields` written is from unimodal POV. In the case of multimodal, we will only check the dataclass attribute name and its type. We can use this type and also check the content to suggest if it could be candidate for filtering

makram93: Have updated the code to check the type and incase of custom type the code should check the content to determine if it is possible to add it as filter_field

now/common/options.py
```diff
@@ -210,18 +213,31 @@ def _infer_app_type_from_demo_data(user_input: UserInput):
...
 )
 
 
+def _update_filter_fields(user_input: UserInput):
+    """
+    Update the filter fields based on the search fields selected by the user.
+
+    :param user_input: UserInput object
+    """
+    user_input.filter_fields_modalities = {
```
joschkabraun, line 234: The dynamic changing of `filter_fields_modalities` looks bad to me. We avoided doing it in the detect schema PR as these are hidden changes which are difficult to understand later on.

now/common/options.py
```diff
@@ -233,14 +249,20 @@ def _infer_app_type_from_demo_data(user_input: UserInput):
...
 )
 
 
+def _reset_filter_modalities(user_input: UserInput):
+    filter_modalities = {}
+    if user_input.filter_fields:
+        for field, field_type in user_input.filter_fields_modalities.items():
+            if field in user_input.filter_fields:
+                filter_modalities[field] = field_type
+    user_input.filter_fields_modalities = filter_modalities
```
joschkabraun, line 263: Similarly, here

now/common/detect_schema.py
joschkabraun: The current logic has those issues
- it would allow any kind of types be allowed as search fields (e.g. music, or arbitrary videos)
- as far as I understand, it only allows for usage of `text` fields for filtering. But how are e.g. floats handled?

now/deployment/flow.py
```diff
@@ -152,7 +152,10 @@ def deploy_flow(
             gateway_port_internal = None  # Since host contains protocol
 
         elif deployment_type == 'remote':
-            flow = deploy_wolf(path=flow_yaml)
+            try:
+                flow = deploy_wolf(path=flow_yaml)
+            except BaseException as e:
```
joschkabraun, line 157: This `try ... except` block is obsolete

joschkabraun (changes requested):

now/common/detect_schema.py
makram93: For the docarray/demo, this PR only allows:
-  Those dataclass fields which are text-based, their content has text and not blob or tensors hence, fields with any non-text content will be filtered out
- All items in the tags because tags can only hold string/bool/float and this matches with the allowed column's type for qdrant, i.e., `str, float, int, bool`

deployment/playground/src/constants.py
```diff
@@ -17,9 +17,7 @@ class Parameters:
     {"iceServers": [{"urls": ["stun:stun.l.google.com:19302"]}]}
 )
 
-root_data_dir = (
-    'https://storage.googleapis.com/jina-fashion-data/data/one-line/datasets/'
-)
+root_data_dir = 'https://jina-now-demo.s3.eu-central-1.amazonaws.com/'
```
jupyterjazz, line 20: Should we make this uppercase?

now/common/options.py
```diff
@@ -40,24 +40,16 @@
...
-        _search_modality = user_input.search_fields_modalities[
-            user_input.search_fields[0]
-        ]
+    if len(user_input.search_fields) != 1:
+        raise ValueError(
+            'Currently only one search field is supported. Please choose one field.'
+        )
+    if user_input.search_fields[0] not in user_input.search_fields_modalities.keys():
+        raise ValueError(
+            f'Search field specified is not among the search candidates field. Kindly '
```
jupyterjazz, line 49:
```suggestion
            f'Search field specified is not among the search candidate fields. Kindly '
```

now/common/options.py
```diff
@@ -202,11 +197,10 @@ def _get_demo_data_choices(user_input: UserInput, **kwargs):
     ],
     prompt_message='Please select the search fields:',
     prompt_type='checkbox',
-    depends_on=DATASET_TYPE,
+    depends_on=True,
```
jupyterjazz, line 200: What's the purpose of this?

now/common/utils.py
```diff
@@ -207,11 +207,10 @@ def get_indexer_config(
...
 def _extract_tags_for_indexer(user_input: UserInput):
-    final_tags = [
-        [tag, value]
-        for tag, value in user_input.filter_fields_modalities.items()
-        if tag in user_input.filter_fields
-    ]
+    final_tags = []
+    for tag, value in user_input.filter_fields_modalities.items():
+        if tag in user_input.filter_fields:
+            final_tags.append([tag, value])
```
jupyterjazz, line 213: Did you change anything here?

jupyterjazz (commented):
Nice!

left few comments and questions

deployment/playground/src/constants.py
makram93: Yep, okay

now/common/options.py
makram93: It builds the hierarchy, typically what depends on what but in this case you are right, we can omit it and have no `depends_on` since it should be shown in all scenarios

now/common/utils.py
makram93: Yes.. filter_field_modalities holds all candidates fields and only those selected in `filter_fields` are added as final tags to be set as column in the table

jupyterjazz (approved):
LGTM!

violenil (approved):
LGTM!🚀

now/common/detect_schema.py
```diff
@@ -63,21 +63,58 @@ def _create_candidate_search_filter_fields(field_name_to_value):
...
+            )
+        field_pos = value['structValue']['fields']['position']['numberValue']
+        try:
+            modality = da[0].chunks[field_pos].modality.lower()
+            if modality not in AVAILABLE_MODALITIES_FOR_SEARCH:
+                raise ValueError(
+                    f'The modality {modality} is not supported for search. Please use one of the following modalities: '
+                    f'{AVAILABLE_MODALITIES_FOR_SEARCH}'
+                )
+        except Exception as e:
```
joschkabraun, line 96: I might miss this, but doesn't this `except` also catches above `ValueError` and overwrites its hint to the available modalities for search

now/common/detect_schema.py
```diff
@@ -63,21 +63,58 @@ def _create_candidate_search_filter_fields(field_name_to_value):
...
+            raise ValueError(
+                f'No modality found for {field_name}. Please follow the steps in the documentation'
+                f' to add modalities to your documents https://docarray.jina.ai/datatypes/multimodal/'
+            )
+        # only the text fields are added to the filter modalities
+        if modality == 'text':
```
joschkabraun, line 102: Can you use the constant used to define which fields are allowed for filtering (in this case which aren't). This makes it easier to unify handling of modalities

now/common/options.py
```diff
@@ -40,24 +40,16 @@
...
-            user_input.search_fields[0]
-        ]
+    if len(user_input.search_fields) != 1:
+        raise ValueError(
+            'Currently only one search field is supported. Please choose one field.'
+        )
+    if user_input.search_fields[0] not in user_input.search_fields_modalities.keys():
+        raise ValueError(
+            f'Search field specified is not among the search candidate fields. Kindly '
+            'choose one of the following: {user_input.search_fields_modalities.keys()}'
```
joschkabraun, line 50: This line also, needs `f'...'`

now/common/detect_schema.py
joschkabraun: Okay!

joschkabraun (approved):
LGTM! A few minor comments

tests/unit/test_dialog.py
```diff
@@ -52,105 +43,71 @@ def __call__(self, question: Dict):
...
-            'search_fields': ['uri'],
-            'filter_fields': ['text'],
-            'cluster': 'new',
-            'deployment_type': 'local',
-            'jwt': {'token': os.environ['WOLF_TOKEN']},
-            'admin_emails': ['team-now@jina.ai'],
-        },
-        {},
-    ),
+    # (
```
florian-hoenicke, line 56: needs to be uncommented

florian-hoenicke (changes requested):
LGTM once the issues are addressed mentioned by me and the others.
Great work!

now/common/detect_schema.py
makram93: Good catch! 👍

tests/unit/test_dialog.py
```diff
@@ -52,105 +43,71 @@ def __call__(self, question: Dict):
...
+            'filter_fields': [],
+            'filter_fields_modalities': {'label': 'text'},
             'cluster': 'new',
             'deployment_type': 'local',
         },
         {},
     ),
-    (
-        {
-            'app': Apps.IMAGE_TEXT_RETRIEVAL,
```
joschkabraun, line 63: Is there a reason to delete all those tests?

makram93: These tests were specifically for `subset_laion` dataset which is not multi-modal, hence these tests are not applicable here.

PR #763, opened 2022-12-14
refactor: remove output modality

joschkabraun:
This PR

- removes `output_modality` from `UserInput` and uses `search_fields` instead
- removes output modality question from CLI 
- removes unused app `TextAndImage`, `MusicToMusic`
- removes unused modality `TextAndImage`, `MusicToMusic`
- fixes identification of modalities in the case we have no subfolder structure in local path or s3 bucket
- defines constants to manage which modalities are supported search & filter fields
- removes `app` attribute from `NOWPreprocessor` as it's unused
- updates BFF and playground to reflect removal of apps
- refactor `detect_schema` to use the same code for s3 and local files to understand the structure of the data

CLI: 
Screenshot 2022-12-15 at 20 03 13

BFF still works (only showed `text-to-video`)
Screenshot 2022-12-15 at 20 06 23

---

- [x] Tests changed
- [ ] Documentation adjusted
- [x] checked changes are back compatible for relevant flows


now/data_loading/data_loading.py
```diff
@@ -215,24 +208,22 @@ def fetch_da_from_url(
     return da
 
 
-def get_dataset_url(dataset: str, output_modality: Modalities) -> str:
+def get_dataset_url(dataset: str) -> str:
+    output_modality = None
```
violenil, line 212: Perhaps we can keep the naming consistent here with that of above, using `search_modality` instead of `output_modality`. What do you think?

joschkabraun: Yes, that is a good point!

florian-hoenicke (approved):
Nice cleanup

PR #764, opened 2022-12-14
refactor(utils): use hubble to get user token

npitsillos:
This PR refactor's this [function](https://github.com/jina-ai/now/blob/main/now/utils.py#L127) to use `hubble.utils.auth.Auth.get_user_token()` for obtaining the `auth_token` without requiring the file to be created in the GHA workflow.

---

- [ ] Test added
- [ ] Documentation adjusted


azayz (approved):
LGTM, wait for CI to work

.github/workflows/ci.yml
```diff
@@ -260,10 +265,10 @@ jobs:
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           S3_CUSTOM_DATA_PATH: ${{ secrets.S3_CUSTOM_DATA_PATH }}
-#      - name: Setup tmate session for debugging
-#        if: ${{ failure() || (github.event_name == 'workflow_dispatch' && github.event.inputs.debug_enabled) }}
-#        uses: mxschmitt/action-tmate@v3
-#        timeout-minutes: 60
+      - name: Setup tmate session for debugging
```
makram93, line 268: Let's comment out this part. One can enable it for their respective PR if needed but having this, in general, is not a wanted behavior

now/utils.py
```diff
@@ -125,9 +125,7 @@ def jina_auth_login():
 
 
 def get_info_hubble(user_input):
-    with open(user('~/.jina/config.json')) as fp:
-        config_val = json.load(fp)
-        user_token = config_val['auth_token']
+    user_token = hubble.utils.auth.Auth.get_auth_token()
```
nomagick, line 128:
```suggestion
```

now/utils.py
```diff
@@ -125,9 +125,7 @@ def jina_auth_login():
...
     client = hubble.Client(token=user_token, max_retries=None, jsonify=True)
```
nomagick, line 129:
```suggestion
    client = hubble.Client(max_retries=None, jsonify=True)
```
Just omit the token parameter will be enough.

nomagick (commented):
You don't have to set the token manually.
The SDK is capable of finding the token by itself.

makram93 (approved):
LGTM! Just a minor change - disable tmux

now/utils.py
npitsillos: Thanks for pointing that out! I didn't know that, I will make this change.

nomagick: To get the token `Auth.get_auth_token()`.
Or `client._token`.

PR #767, opened 2022-12-14
fix: enable local testing

makram93:
- [x] Enable local testing
- [x] Update the executor id in ci.yml
- [x] fix the test_dialog by updating the admin_emails


florian-hoenicke (approved):
great that you spotted the issues

PR #768, opened 2022-12-15
feat: use data class in data loading

azayz:
Use data class in data loading
---

- [ ] Test added
- [ ] Documentation adjusted


now/constants.py
```diff
@@ -23,6 +25,15 @@ class Modalities(BetterEnum):
     TEXT_AND_IMAGE = 'text-and-image'
 
 
+def modalities_mapping():
```
florian-hoenicke, line 28: I think it does not have to be a function but can just be a dictionary.

now/data_loading/data_loading.py
```diff
@@ -23,12 +21,12 @@
 from now.utils import download, sigmap
 
 
-def load_data(app: JinaNOWApp, user_input: UserInput) -> DocumentArray:
+def load_data(user_input: UserInput, data_class) -> DocumentArray:
```
florian-hoenicke, line 24: if possible, you can define a type for data_class

now/data_loading/data_loading.py
```diff
@@ -141,27 +84,25 @@ def _extract_es_data(user_input: UserInput) -> DocumentArray:
     return extracted_docs
 
 
-def _load_from_disk(app: JinaNOWApp, user_input: UserInput) -> DocumentArray:
+def _load_from_disk(user_input: UserInput, dataclass) -> DocumentArray:
```
florian-hoenicke, line 87: type hint for dataclass.

now/data_loading/data_loading.py
```diff
@@ -170,28 +111,133 @@ def _load_from_disk(app: JinaNOWApp, user_input: UserInput) -> DocumentArray:
         )
 
 
-def _list_files_from_s3_bucket(app: JinaNOWApp, user_input: UserInput) -> DocumentArray:
+def from_files(
```
florian-hoenicke, line 114: maybe better name: from_local files?

now/data_loading/data_loading.py
```diff
@@ -170,28 +111,133 @@ def _load_from_disk(app: JinaNOWApp, user_input: UserInput) -> DocumentArray:
...
 
 
+def create_docs_from_subdirectories_s3(
+    subdirectories: List, path: str, fields: List[str], data_class, bucket
+) -> List[Document]:
+    docs = []
+    kwargs = {}
+    for subdirectory in subdirectories:
+        for obj in list(bucket.objects.filter(Prefix=subdirectory))[1:]:
+            file = obj.key.split('/')[-1]
```
florian-hoenicke, line 214: Some lines in this function are duplicated. We should have this logic in a separate function that is called by both functions

florian-hoenicke (changes requested):

now/common/detect_schema.py
```diff
@@ -208,10 +220,15 @@ def _check_contains_files_only_local_folder(dataset_path):
...
     """
+    file_name = {}
+    first_file = ''
     for file_or_directory in os.listdir(dataset_path):
         if not os.path.isfile(os.path.join(dataset_path, file_or_directory)):
-            return False
-    return True
+            return file_name
+        if not first_file and not file_or_directory.startswith('.'):
+            first_file = file_or_directory
```
joschkabraun, line 229: issue is with that solution: by chance the first file could be a not supported file type but the rest of the files are all jpegs.

now/common/detect_schema.py
```diff
@@ -266,7 +285,14 @@ def set_field_names_from_local_folder(user_input: UserInput, **kwargs):
         raise ValueError(
             'The path provided is not a folder, please check documentation https://now.jina.ai'
         )
-    if _check_contains_files_only_local_folder(dataset_path):
+    file_names = _check_contains_files_only_local_folder(dataset_path)
```
joschkabraun, line 288: in the following is a lot of duplicate code. Similarly for s3 code

joschkabraun (changes requested):

florian-hoenicke (changes requested):

now/run_backend.py
```diff
@@ -205,3 +206,161 @@ def estimate_request_size(index, max_request_size):
...
+    :param dict2: dict to be used for updating
+    """
+    for key, value in dict2.items():
+        if key not in dict1:
+            dict1[key] = value
+
+
+def create_dataclass(user_input: UserInput):
+    """
+    Create a dataclass from the user input using the selected search anf filter fields
```
jupyterjazz, line 225:
```suggestion
    Create a dataclass from the user input using the selected search and filter fields
```

now/run_backend.py
```diff
@@ -205,3 +206,161 @@ def estimate_request_size(index, max_request_size):
...
+    and their corresponding modalities
+
+    :param user_input: user input
+    """
+    all_annotations = {}
+    all_class_attributes = {}
+    all_modalities = {}
+    all_modalities.update(user_input.search_fields_modalities)
```
jupyterjazz, line 233:
```suggestion
    all_modalities = user_input.search_fields_modalities
```

now/run_backend.py
```diff
@@ -205,3 +206,161 @@ def estimate_request_size(index, max_request_size):
...
+    update_dict_with_no_overwrite(all_modalities, user_input.filter_fields_modalities)
```
jupyterjazz, line 234: I was wondering if instead of ignoring it, we should raise an error in case there's an overlap between search and filter fields..

tests/unit/test_dataclass.py
```diff
@@ -0,0 +1,116 @@
...
+    expected_dataclass,
+):
+    user_input = UserInput()
+    user_input.dataset_type = dataset_type
+    user_input.search_fields = search_fields
+    user_input.search_fields_modalities = search_fields_modalities
+    user_input.filter_fields = filter_fields
+    user_input.filter_fields_modalities = filter_fields_modalities
+    mm_doc = create_dataclass(user_input)
+    assert str(mm_doc) == "<class 'now.run_backend.MMDoc'>"
```
jupyterjazz, line 113: maybe better to check `mm_doc.__class__.__name__`?

now/run_backend.py
```diff
@@ -205,3 +206,161 @@ def estimate_request_size(index, max_request_size):
...
+
+    file_mapping_to_dataclass_fields = create_dataclass_fields_file_mappings(
+        user_input.search_fields + user_input.filter_fields, all_modalities
+    )
+    user_input.files_to_dataclass_fields = file_mapping_to_dataclass_fields
+    (
+        search_fields_annotations,
+        search_fields_class_attributes,
+    ) = create_annotations_and_class_attributes(
```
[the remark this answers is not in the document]
jupyterjazz: Agree with Florian.
Would be nice to somehow combine `create_annotations_and_class_attributes` calls for both search and filter fields.

tests/resources/subdirectories_structure_folder/folder2/test.txt
```diff
@@ -0,0 +1 @@
+hello
```
jupyterjazz, line 1: `subdirectories_structure_folder` seems a bit verbose, maybe just `subdirectories`?

jupyterjazz (commented):
Great work!

left some minor suggestions

now/run_backend.py
azayz: it is expected to be an overlap, in fact these dictionaries contain candidates for search and filter fields so some fields could be used for search or for filtering. We create these dicts before the user chooses therefore there's an overlap

now/constants.py
azayz: only partial migration I believe is possible now, full migration will add complexity to the PR and it's out of scope I would rather do it in another refactoring PR

now/data_loading/data_loading.py
azayz: done

now/run_backend.py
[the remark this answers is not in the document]
azayz: adressed

now/common/options.py
```diff
@@ -50,9 +51,16 @@ def _create_app_from_user_input(user_input: UserInput, **kwargs):
             f'choose one of the following: {user_input.search_fields_modalities.keys()}'
         )
     _search_modality = user_input.search_fields_modalities[user_input.search_fields[0]]
-    if _search_modality in ['image', 'text']:
+
+    if user_input.dataset_type in [DatasetTypes.DEMO, DatasetTypes.DOCARRAY]:
```
joschkabraun, line 55: Why not apply the same mapping you did for the other `DatasetTypes` to them as well to standardize it?
I think this different handling of the modalities adds unnecessary cognitive load on the developer. We should standardize it.

now/data_loading/data_loading.py
```diff
@@ -151,29 +99,239 @@ def _load_from_disk(app: JinaNOWApp, user_input: UserInput) -> DocumentArray:
...
-    bucket, folder_prefix = get_s3_bucket_and_folder_prefix(user_input)
+    :param path: The path to the directory
+    :param fields: The fields to search for in the directory
+    :param files_to_dataclass_fields: The mapping of the files to the dataclass fields
+    :param data_class: The dataclass to use for the document
+
+    :return: A DocumentArray with the documents
+    """
 
+    def get_subdirectories_local_path(directory):
```
joschkabraun, line 118: no need to create this function

now/data_loading/data_loading.py
```diff
@@ -151,29 +99,239 @@ def _load_from_disk(app: JinaNOWApp, user_input: UserInput) -> DocumentArray:
...
+    :param fields: The fields to search for in the directory
+    :param files_to_dataclass_fields: The mapping of the files to the dataclass fields
+    :param data_class: The dataclass to use for the document
+
+    :return: The list of documents
+    """
+
+    docs = []
+    kwargs = {}
+    for subdirectory in subdirectories:
```
joschkabraun, line 158: this logic fails if all the files are structured in this format e.g.
2022/uuid/file.jpeg
2022/uuid/text.txt
2022/uuid2/file.jpeg
2022/uuid2/text.txt
2021/uuid3/file.jpeg
2021/uuid3/text.txt

joschkabraun (changes requested):
Only halfway reviewed it, but there are is a lot of redundancy in the `data_loading` for s3 & local files. I would recommend looking at how this is handled in `detect_schema` and follow a similar a schema.

tests/unit/test_dataclass.py
azayz: mm_doc's class name is actually python type, so when you do that you get type 
'type'

now/common/options.py
azayz: correct, applied

now/data_loading/data_loading.py
azayz: done

now/common/detect_schema.py
```diff
@@ -41,20 +40,20 @@ def _create_candidate_search_filter_fields(field_name_to_value):
...
-                search_fields_modalities[field_name] = modality
+                search_fields_modalities[field_name] = MODALITIES_MAPPING[modality]
                 break
             elif field_name == 'uri' and field_value.split('.')[-1] in file_types:
-                search_fields_modalities[field_name] = modality
+                search_fields_modalities[field_name] = MODALITIES_MAPPING[modality]
                 break
             elif field_name == 'text' and field_value:
-                search_fields_modalities[field_name] = Modalities.TEXT
+                search_fields_modalities[field_name] = MODALITIES_MAPPING['text']
```
joschkabraun, line 49: `modality = 'text'`, isn't it?

now/common/options.py
```diff
@@ -50,7 +51,11 @@ def _create_app_from_user_input(user_input: UserInput, **kwargs):
             f'choose one of the following: {user_input.search_fields_modalities.keys()}'
         )
     _search_modality = user_input.search_fields_modalities[user_input.search_fields[0]]
-    if _search_modality in ['image', 'text', 'video']:
+
+    if _search_modality in [Modalities.TEXT, Modalities.VIDEO, Modalities.IMAGE]:
```
joschkabraun, line 55: I think we can skip this `if` condition, or is there a need for it?

now/data_loading/data_loading.py
```diff
@@ -122,27 +70,32 @@ def _extract_es_data(user_input: UserInput) -> DocumentArray:
...
+    Loads the data from disk into multimodal documents.
+
+    :param user_input: The user input object.
+    :param dataclass: The dataclass to use for the DocumentArray.
+    """
     dataset_path = user_input.dataset_path.strip()
     dataset_path = os.path.expanduser(dataset_path)
     if os.path.isfile(dataset_path):
         try:
             return DocumentArray.load_binary(dataset_path)
```
joschkabraun, line 84: We assume that the docarray is in multi-modal format, right? Then I would suggest that we check that it has this format.

now/data_loading/data_loading.py
```diff
@@ -151,45 +104,181 @@ def _load_from_disk(app: JinaNOWApp, user_input: UserInput) -> DocumentArray:
...
-    s3_uri = user_input.dataset_path
-    if not s3_uri.startswith('s3://'):
-        raise ValueError(
-            f"Can't process S3 URI {s3_uri} as it assumes it starts with: 's3://'"
+    :param file_paths: The list of file paths
+    :param fields: The fields to search for in the directory
+    :param files_to_dataclass_fields: The mapping of the files to the dataclass fields
+    :param data_class: The dataclass to use for the document
+    :param path: The path to the directory
+    :param s3_dataset: Whether the dataset is stored on s3
```
joschkabraun, line 156: for convention, I would suggest something indicating it is a boolean, e.g. `is_s3_dataset`

now/data_loading/data_loading.py
```diff
@@ -151,45 +104,181 @@ def _load_from_disk(app: JinaNOWApp, user_input: UserInput) -> DocumentArray:
...
 
-def deep_copy_da(da: DocumentArray) -> DocumentArray:
-    new_da = DocumentArray()
-    for i, d in enumerate(da):
-        new_doc = deepcopy(d)
-        new_doc.id = str(uuid.uuid4())
-        new_da.append(new_doc)
-    return new_da
+
+def _extract_file_and_full_file_path(obj, path=None, s3_dataset=False):
```
joschkabraun, line 274: I would rename `obj` to something else to help understand that this isn't only code for s3 but in general to interact with a file system.

joschkabraun, line 274: Also, I would rename `s3_dataset` here as well.

now/data_loading/data_loading.py
```diff
@@ -151,45 +104,181 @@ def _load_from_disk(app: JinaNOWApp, user_input: UserInput) -> DocumentArray:
...
+            if file in fields:
+                kwargs[files_to_dataclass_fields[file]] = file_full_path
+                continue
+            if file.endswith('.json'):
+                if s3_dataset:
+                    for field in data_class.__annotations__.keys():
+                        if field not in kwargs.keys():
+                            kwargs[field] = file_full_path
+                else:
+                    json_f = open(file_full_path)
```
joschkabraun, line 185: for reading files, it's recommended to use a context manager

now/run_backend.py
```diff
@@ -192,3 +200,133 @@ def estimate_request_size(index, max_request_size):
...
+
+    :param dict1: dict to be updated
+    :param dict2: dict to be used for updating
+    """
+    for key, value in dict2.items():
+        if key not in dict1:
+            dict1[key] = value
+
+
+def create_dataclass(user_input: UserInput):
```
joschkabraun, line 217: Maybe move this into its own file under `data_loading`?

now/run_backend.py
```diff
@@ -192,3 +200,133 @@ def estimate_request_size(index, max_request_size):
...
+    """
+
+    modalities_count = {}
+    for modality_name in MODALITIES_MAPPING.keys():
+        modalities_count[modality_name] = 0
+
+    dataclass_fields_to_file_mapping = {}
+    filter_count = 0
+    for f in fields:
+        if not isinstance(f, typing.Hashable):
```
joschkabraun, line 317: Not sure what the reason for this is, but it looks a bit like "solving the symptom and not the culprit". What causes the selection of fields which aren't hashable? This looks like we didn't do the right filtering earlier.

now/run_backend.py
```diff
@@ -192,3 +200,133 @@ def estimate_request_size(index, max_request_size):
...
+    """
+    Create a mapping between the dataclass fields and the file fields
+
+    :param fields: list of fields
+    :param fields_modalities: dict of fields and their modalities
+    """
+
+    modalities_count = {}
+    for modality_name in MODALITIES_MAPPING.keys():
+        modalities_count[modality_name] = 0
```
joschkabraun, line 312: can be abbreviated to `modalities_count = defaultdict(int)`

joschkabraun, line 312: can be simplified to `modalities_count = defaultdict(int)`

joschkabraun (approved):
Really great work!! :rocket:
Some minor comments. PR can be merged after addressing those, green CI and testing with customer flow.

now/common/detect_schema.py
azayz: Text which is coming from docarray.typing

now/common/options.py
azayz: I think we keep it to make sure that search modality is among those modalities else ye'll trigger the value error, also it makes it easier to debug with the value error

now/common/detect_schema.py
joschkabraun: Isn't `Modalities.Text = 'text'` as it's defined this way in `constants.py`?

now/common/options.py
joschkabraun: We determine candidates for search modality in `detect_schema`, so this shouldn't be an error which can happen. Let's remove as otherwise, we duplicate the checks for which modalities are allowed for searching.

now/common/detect_schema.py
azayz: yes it is

now/data_loading/data_loading.py
```diff
@@ -46,68 +48,21 @@ def load_data(app: JinaNOWApp, user_input: UserInput) -> DocumentArray:
...
 def _pull_docarray(dataset_name: str):
     try:
-        return DocumentArray.pull(name=dataset_name, show_progress=True)
+        docs = DocumentArray.pull(name=dataset_name, show_progress=True)
+        if is_multimodal(docs[0]):
+            return docs
+        else:
+            raise ValueError(
+                f'The dataset {dataset_name} does not contain a multimodal DocumentArray.'
+            )
```
jupyterjazz, line 59: would be nice to add a link here guiding developers on how to create proper multimodal data

now/data_loading/data_loading.py
```diff
@@ -27,11 +31,9 @@ def load_data(app: JinaNOWApp, user_input: UserInput) -> DocumentArray:
         da = _pull_docarray(user_input.dataset_name)
     elif user_input.dataset_type == DatasetTypes.PATH:
         print('💿  Loading files from disk')
-        da = _load_from_disk(app, user_input)
-        da = _load_tags_from_json_if_needed(da, user_input)
+        da = _load_from_disk(user_input=user_input, dataclass=data_class)
```
jupyterjazz, line 34: to stay consistent, we should call this `data_class` too

now/data_loading/data_loading.py
```diff
@@ -122,27 +77,38 @@ def _extract_es_data(user_input: UserInput) -> DocumentArray:
...
         try:
-            return DocumentArray.load_binary(dataset_path)
-        except Exception as e:
+            docs = DocumentArray.load_binary(dataset_path)
+            if is_multimodal(docs[0]):
+                return docs
+            else:
+                raise ValueError(
+                    f'The file {dataset_path} does not contain a multimodal DocumentArray.'
+                )
```
jupyterjazz, line 97: same here about the instructions

jupyterjazz (approved):
LGTM!

left some minor comments

now/common/options.py
azayz: correct removed

now/common/detect_schema.py
```diff
@@ -46,7 +47,9 @@ def _create_candidate_search_filter_fields(field_name_to_value):
                 search_fields_modalities[field_name] = MODALITIES_MAPPING[modality]
                 break
             elif field_name == 'text' and field_value:
-                search_fields_modalities[field_name] = MODALITIES_MAPPING['text']
+                search_fields_modalities[field_name] = MODALITIES_MAPPING[
+                    Modalities.TEXT
```
joschkabraun, line 51: why not use the variable `modality`?

azayz: because we loop over all the modalities, so it could be that this field name is coming from docarray and the field name is text so we know this field's modality is text but the modality variable could be image or anything

joschkabraun: Ah, this makes sense. Now I understand the different handling. I would suggest to move this check then out of the for loop to avoid confusion

jupyterjazz (approved):
:rocket:

violenil (approved):
Looks great!🚀

now/data_loading/create_dataclass.py
```diff
@@ -0,0 +1,136 @@
...
+
+    :param dict1: dict to be updated
+    :param dict2: dict to be used for updating
+    """
+    for key, value in dict2.items():
+        if key not in dict1:
+            dict1[key] = value
+
+
+def create_dataclass(user_input: UserInput):
```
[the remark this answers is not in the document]
azayz: yes and added in doctoring

now/data_loading/create_dataclass.py
```diff
@@ -0,0 +1,136 @@
...
+    """
+    Create a dataclass from the user input using the selected search and filter fields
+    and their corresponding modalities
```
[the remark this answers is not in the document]
azayz: added in doctoring

PR #770, opened 2022-12-16
fix: replicas for local deployment

IyadhKhalfallah:
- Added condition to add replicas for high-performance flow only in remote deployment.

- Fixed typo for timeout sending in test mode.


joschkabraun (changes requested):
Good findings! I also really like how you simplified verbose `if if .. else else if ... else` structure 🚀 
A few things:
- we have a PR template which is required to be used
- did you have a look at the full ticket? It was also suggest to replace the readiness probe as done here: https://docs.jina.ai/concepts/flow/readiness/

now/app/text_to_video/app.py
```diff
@@ -55,27 +55,20 @@ def required_docker_memory_in_gb(self) -> int:
...
-        if finetuning:
-            if dataset_len > 200_000 and is_jina_email:
-                print(f"🚀🚀🚀 You are using high performance flow")
-                self.flow_yaml = os.path.join(
-                    flow_dir, 'ft-flow-video-clip-high-performance.yml'
-                )
-            else:
-                self.flow_yaml = os.path.join(flow_dir, 'ft-flow-video-clip.yml')
+        flow_prefix = 'ft-' if finetuning else ''
+        if dataset_len > 200_000 and is_jina_email and deployment_type == 'remote':
```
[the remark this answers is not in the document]
florian-hoenicke: That is true. On the other hand apart from debugging, there would be not much benefit of allowing multiple replicas locally. I guess it could give a performance boost, but since local deployment is not for production, the performance boost does not add much value.
Therefore, I think it is fine to not support replicas on local deployments.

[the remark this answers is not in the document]
florian-hoenicke: But regarding the readiness of the client I agree. This should also be handled here. In that case, we can altogether remove the waiting logic and use `client.is_flow_ready`

Then, you can also remove your additional condition. I know, I suggested it to you in our pairing but turns out there is a better way.
1. remove the condition on local deployment you just introduced
2. remove the waiting logic for the flow to start https://github.com/jina-ai/now/blob/2ff6c826fe87dc50632bd7057b5a4ff6c18b0c10/now/deployment/flow.py#L99
3. remove the waiting logic where it tries to make empty requests to the flow (right before indexing) until it is successful https://github.com/jina-ai/now/blob/f643197c886fb4db0ca545bda9dabbb26d42d5e2/now/run_backend.py#L158
4. introduce:
```
client.is_flow_ready()
```
to make sure the flow is ready before indexing (sleep can be removed) https://github.com/jina-ai/now/blob/2ff6c826fe87dc50632bd7057b5a4ff6c18b0c10/now/deployment/flow.py#L102

florian-hoenicke:
Regarding the readiness of the client I agree. This should also be handled here. In that case, we can altogether remove the waiting logic and use `client.is_flow_ready`

Then, you can also remove your additional condition. I know, I suggested it to you in our pairing but turns out there is a better way.
1. remove the condition on local deployment you just introduced
2. remove the waiting logic for the flow to start https://github.com/jina-ai/now/blob/2ff6c826fe87dc50632bd7057b5a4ff6c18b0c10/now/deployment/flow.py#L99
3. remove the waiting logic where it tries to make empty requests to the flow (right before indexing) until it is successful https://github.com/jina-ai/now/blob/f643197c886fb4db0ca545bda9dabbb26d42d5e2/now/run_backend.py#L158
4. introduce:
```
client.is_flow_ready()
```
to make sure the flow is ready before indexing (sleep can be removed) https://github.com/jina-ai/now/blob/2ff6c826fe87dc50632bd7057b5a4ff6c18b0c10/now/deployment/flow.py#L102

now/app/text_to_video/app.py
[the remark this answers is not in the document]
IyadhKhalfallah: @joschkabraun this was implemented during a discussion since the local deployment doesn't really require 20 replicas. But you are right, the ticket should cover more.

florian-hoenicke:
Do we get a lot of error logs even in the normal case where everything is working?
This could pollute the logs

IyadhKhalfallah:
> Do we get a lot of error logs even in the normal case where everything is working? This could pollute the logs

@florian-hoenicke if you are referring to `check_pods_health()` it is only logging when there are errors. In normal cases, it shouldn't log anything other than the usual CLI scenario.

joschkabraun (approved):
LGTM, great work!

now/deployment/flow.py
```diff
@@ -45,22 +46,24 @@ def wait_for_lb(lb_name, ns):
...
-            or not len(pod.status.container_statuses) == 1
-            or not pod.status.container_statuses[0].ready
-        ]
-        if len(not_ready) == 0 and f.num_deployments == len(pods):
-            return
-        sleep(1)
+    pods = v1.list_namespaced_pod(ns).items
+    for pod in pods:
+        try:
+            if pod.status and pod.status.container_statuses:
```
florian-hoenicke, line 55: I think it should make the code more readable if the if statements are combined. What do you think?

florian-hoenicke (changes requested):

now/deployment/flow.py
IyadhKhalfallah: @florian-hoenicke Could you please have a look at the new commit and let me know if that responds to the comment?

now/deployment/flow.py
```diff
@@ -45,22 +46,20 @@ def wait_for_lb(lb_name, ns):
...
-        if len(not_ready) == 0 and f.num_deployments == len(pods):
-            return
-        sleep(1)
+    pods = v1.list_namespaced_pod(ns).items
+
+    for pod in pods:
+        try:
+            container_statuses = pod.status.container_statuses
+            if 'Error' in container_statuses[0].state.waiting.message:
+                raise Exception(
```
florian-hoenicke, line 58: The exception will have no effect. because it is caught by the try

florian-hoenicke (dismissed):

now/deployment/flow.py
IyadhKhalfallah: @florian-hoenicke actually, the try is for `container_statuses[0].state.waiting.message`. That can raise an error because one of `state.waiting.reason` might not exist. And in that case, we don't want to have an error in the CLI, we just want to catch it and pass, right?

PR #771, opened 2022-12-16
fix: reversed text and image modalities

IyadhKhalfallah:
Text and image modalities were showing in a reversed order in playground. Reversing the order for these was the solution.
Screenshot 2022-12-16 at 10 16 52


azayz (approved):
LGTM

jupyterjazz (approved):
LGTM! 🚀

Issue #757, opened 2022-12-13
feat: multi-modal bff


florian-hoenicke:
Make the models decoupled from the application type.

UAC
- [ ] only one single request model for each endpoint
- [ ] only one single response model for each endpoint

PR #772, opened 2022-12-16
Feat multimodal bff

joschkabraun:
New BFF:

Screenshot 2022-12-20 at 19 16 56

---

- [x] This PR references an open issue: #757 
- [x] Test added
- [x] Customer flows will break


deployment/bff/app/v1/routers/helper.py
```diff
@@ -6,6 +6,39 @@
...
+    if len(field_dict) != 1:
+        raise ValueError(
+            f"Multi-modal document isn't supported yet. "
+            f"Can only set one value but have {list(field_dict.keys())}"
+        )
+
+    try:
+        for field_name, field_value in field_dict.items():
+            if field_value.text:
+                doc = Document(text=field_value.text)
```
violenil, line 24: Would it be possible to formulate this doc from `@dataclass` instead of a regular `Document`, which can have one of fields `[text, uri, blob]` or perhaps `[query_text, query_uri, query_blob]` ? We also here iterate over `field_name` which could be used to set this value.

Perhaps this is not yet part of this PR, but for integration with NOWElasticIndexer, the query doc needs to be created with a `multi_modal_schema`.

joschkabraun: This will be part of a follow-up, once the multi-class data class is used in the repo. This PR is currently to change the interface through BFF.

deployment/bff/app/v1/models/search.py
```diff
@@ -0,0 +1,62 @@
...
+    _ProtoValueType, List[_ProtoValueType], Dict[str, _ProtoValueType]
+]
+
+
+class _NamedScore(BaseModel):
+    value: Optional[float] = None
+
+
+class IndexRequestModel(BaseRequestModel):
+    data: List[Tuple[Dict[str, ModalityModel], Dict[str, Any]]] = Field(
```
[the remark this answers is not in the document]
joschkabraun: Re 1st question: Yes, I would argue there is a need. The user differentiates between the data (blob, uri, etc.) and then the tags associated with them. Also, this makes the code significantly easier. Also, `ModalityModel` is basically a dictionary (mapping e.g. `blob` to its value), so the data are more nested than the tags.
Re 2nd question: We don't change the data here as this is the BFF.

violenil (approved):
LGTM!🚀

florian-hoenicke (approved):
LGTM

PR #774, opened 2022-12-19
refactor: no executor tests

florian-hoenicke:
Both jobs had the same precondition and therefore can be joined.
The executor tests are benefiting from all the runners we have since they are now distributed.


.github/workflows/ci.yml
```diff
@@ -275,46 +275,6 @@ jobs:
...
-      #        timeout-minutes: 60
-      - name: Upload coverage
-        uses: actions/upload-artifact@v2
-        with:
-          name: coverage_executor
-          path: .coverage
-
   # just for blocking the merge until all parallel core-test are successful
   success-all-test-coverage:
     runs-on: ubuntu-latest
```
violenil, line 280: I think this 'needs' also should be updated to remove `executor-test`

.github/workflows/cd.yml
```diff
@@ -53,36 +53,6 @@ jobs:
...
-          pytest --suppress-no-test-exit-code -v -s tests/executor/indexer/elastic
-          pytest --suppress-no-test-exit-code -v -s tests/executor/preprocessor
-#      - name: Setup tmate session for debugging
-#        if: ${{ failure() || (github.event_name == 'workflow_dispatch' && github.event.inputs.debug_enabled) }}
-#        uses: mxschmitt/action-tmate@v3
-#        timeout-minutes: 60
-
 
   prerelease:
     runs-on: ubuntu-latest
```
violenil, line 58: same here

.github/workflows/ci.yml
florian-hoenicke: thanks - done

.github/workflows/cd.yml
florian-hoenicke: thanks - done

florian-hoenicke:
test if it fixed by the core

PR #785, opened 2022-12-20
fix: filter types

makram93:
One of the field has modality `text`, however the qdrant accepts str. Hence a conversion is needed

---

- [x] Test added
- [ ] Documentation adjusted


now/common/detect_schema.py
```diff
@@ -71,28 +70,32 @@ def _extract_field_candidates_docarray(response):
     """
     search_modalities = {}
     filter_modalities = {}
-    response = requests.get(response.json()['data']['download'])
-    da = DocumentArray.from_dict(response.json())
-    if not da[0]._metadata:
+    da = requests.get(response.json()['data']['download']).json()[
```
azayz, line 73: the naming da kinda implies document array, I would suggest you name it otherwise since it's now only the first doc, something like doc

now/executor/abstract/base_indexer/base_indexer.py
```diff
@@ -417,7 +417,7 @@ def parse_columns(columns):
             corrected_list = []
             for i in range(0, len(columns), 2):
                 # This conversion is needed for MMDocs
-                if columns[i + 1] in ['Text', 'stringValue']:
```
jupyterjazz, line 420: So the problem was that "text" was capitalized instead of being lowercase?

makram93: Yes.. So, now we just send lower `text`

PR #804, opened 2022-12-23
feat: elastic cluster in executor

florian-hoenicke:
Since the external cluster is not up and running at the moment. I suggest that we create the cluster inside the executor for now and see how this performs. I think this is better than pushing an in-memory indexer.
Let's see if we get it performant. I guess adding more CPU could then also help to speed it up


violenil (approved):
Great idea! I think this is a good way to go for now.

jupyterjazz (approved):

PR #820, opened 2022-12-27
fix: split executor tests

IyadhKhalfallah:
---

- [ ] Created two executor tests instead of one


.github/workflows/ci.yml
```diff
@@ -299,25 +299,52 @@ jobs:
       - name: Test executors separately
         timeout-minutes: 15
```
violenil, line 300: You should be able to reduce the timeout for the `executor-test` and `executor-test_indexers` now that they are split. Can you check how long they tend to take and add a few minutes buffer? Hopefully should be something more like 10 minutes or so

.github/workflows/ci.yml
```diff
@@ -299,25 +299,52 @@ jobs:
...
         run: |
-          pytest --suppress-no-test-exit-code --cov=now --cov-append -v -s tests/executor/auth
           pytest --suppress-no-test-exit-code --cov=now --cov-append -v -s tests/executor/indexer/base
           pytest --suppress-no-test-exit-code --cov=now --cov-append -v -s tests/executor/indexer/elastic
+      - name: Upload coverage
```
azayz, line 304: if I understand this correctly, we have two coverage reports one for indexers and one for the rest of the executors, if so is it possible to only have one coverage report

makram93 (approved):
LGTM!

.github/workflows/cd.yml
```diff
@@ -74,10 +74,37 @@ jobs:
       - name: Test executors separately
         timeout-minutes: 15
         run: |
-          pytest --suppress-no-test-exit-code -v -s tests/executor/auth
           pytest --suppress-no-test-exit-code -v -s tests/executor/indexer/base
           pytest --suppress-no-test-exit-code -v -s tests/executor/indexer/elastic
+#      - name: Setup tmate session for debugging
+#        if: ${{ failure() || (github.event_name == 'workflow_dispatch' && github.event.inputs.debug_enabled) }}
+#        uses: mxschmitt/action-tmate@v3
+#        timeout-minutes: 60
```
jupyterjazz, line 82: should we remove this?

jupyterjazz (approved):
LGTM!

.github/workflows/ci.yml
```diff
@@ -284,25 +284,51 @@ jobs:
...
+      ]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python 3.7
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
+      - name: Prepare environment
+        run: |
+          sudo apt update && sudo apt install ffmpeg
```
florian-hoenicke, line 312: Good that you think about how to optimize the ci.
I think this is the wrong way since it introduces code duplication.
A better way is using a matrix run.
However, the best solution will be to run all the tests together and use pytest split to get the perfect split of all executors

florian-hoenicke (changes requested):

.github/workflows/ci.yml
IyadhKhalfallah: @florian-hoenicke I think you are right about this, although I'm not sure I get your last sentence. Do you think we should keep the executor tests apart and apply the matrix run on them (like I did in the PR I just created [here](https://github.com/jina-ai/now/pull/885) or should we mix the core tests and executor tests together?

Issue #793, opened 2022-12-22
refactor: create multimodal document in bff


florian-hoenicke:
We need to create mmdocs in the bff because it will become the only type of document that is accepted by the flow.

UAC
- [ ] bff uses mmdoc for requests sent to the flow

Issue #760, opened 2022-12-13
feat: dummy multi-modal queries


florian-hoenicke:
UAC
- [ ] multimodal queries (one field per query). This means we run dummy multi-modal queries. It is a multi-modal document with only one modality

PR #822, opened 2022-12-27
Feat multimodal docs bff

joschkabraun:
Only internal changes in BFF, except for the updated `SearchRequestModel` see below:
Screenshot 2022-12-28 at 14 55 33

---

- [x] This PR references an open issue: #793 and #760
- [x] Test added
- [x] customer flow checked


.github/workflows/cd.yml
```diff
@@ -113,6 +113,10 @@ jobs:
           filters: |
             has_deployment_changed:
             - 'deployment/**'
+      - name: replace now commit hash in Dockerfile
```
azayz, line 116: what is the point of this ?

deployment/bff/app/v1/routers/cloud_temp_link.py
```diff
@@ -24,11 +24,11 @@ def temp_link(data: CloudTempLinkRequestModel):
         docs.append(Document(id=id, uri=uri))
 
     docs = jina_client_post(
-        data=data,
+        request_model=data,
         inputs=DocumentArray(docs),
         endpoint='/temp_link_cloud_bucket',
         parameters={},
-        target_executor=r'\Asecurity_check\Z|\Apreprocessor\Z',
```
azayz, line 31: why is this removed ?

deployment/bff/app/v1/routers/search.py
```diff
@@ -10,7 +11,12 @@
     SearchResponseModel,
     SuggestionRequestModel,
 )
-from deployment.bff.app.v1.routers.helper import field_dict_to_doc, jina_client_post
+from deployment.bff.app.v1.routers.helper import (
+    fetch_user_input,
+    field_dict_to_mm_doc,
+    jina_client_post,
+)
+from now.run_backend import create_dataclass
```
azayz, line 19: this is no longer in run backend, it's in it's own file now

.github/workflows/cd.yml
joschkabraun: It's current solution to the issue that one can't check in which step of a Dockerfile an environment variable is used and by that invalidating only the necessary caches. Original solution here: https://github.com/jina-ai/now/blob/main/.github/actions/push-executor/action.yml#L70

deployment/bff/app/v1/routers/cloud_temp_link.py
joschkabraun: We don't have an executor called `security_check` anymore.

.github/workflows/cd.yml
joschkabraun: We need the NOW package now in the BFF to create the right dataclass.

deployment/playground/src/search.py
```diff
@@ -87,7 +87,7 @@ def search(
     if endpoint == 'suggestion':
         data[attribute_name] = attribute_value
     elif endpoint == 'search':
-        data['query'] = {'search_field': {attribute_name: attribute_value}}
+        data['query'] = {f'query_{attribute_name}': {attribute_name: attribute_value}}
```
violenil, line 90: Is it possible to remove some nesting here? For example:
```suggestion
        data['query'] = {attribute_name: attribute_value}
```

now/executor/autocomplete/executor.py
```diff
@@ -48,7 +48,7 @@ def update_save_words(self, word):
     def search_update(
         self, docs: Optional[DocumentArray] = None, parameters: dict = {}, **kwargs
     ):
-        for doc in docs:
+        for doc in docs['@r,c']:
```
violenil, line 51: good catch

violenil (approved):
LGTM! 🚀
Just added some small comments/suggestions.

deployment/playground/src/search.py
joschkabraun: No, we unfortunately need this nesting as we now introduce the API to have multiple query fields.
So in the future, it might look like
```python
{'query': 
    {'title': {'text': 'a title'}},
    {'banner': {'uri': 'www.website/banner.jpg'}}
}
```

deployment/bff/app/v1/routers/helper.py
```diff
@@ -1,90 +1,62 @@
...
+    with TemporaryDirectory() as tmp_dir:
+        try:
+            if field_names_to_dataclass_fields:
+                field_dict_orig = deepcopy(field_dict)
+                field_dict = {}
+                for (
+                    file_name,
+                    field_name_data_class,
+                ) in field_names_to_dataclass_fields.items():
+                    field_dict[field_name_data_class] = field_dict_orig[file_name]
```
jupyterjazz, line 39:
```suggestion
                field_dict = {
                    field_name_data_class: field_dict_orig[file_name]
                    for file_name, field_name_data_class in field_names_to_dataclass_fields.items()
                }
```

deployment/bff/app/v1/routers/helper.py
```diff
@@ -1,90 +1,62 @@
...
+            data_class_kwargs = {}
+            for field_name_data_class, field_value in field_dict.items():
+                # save blob into a temporary file such that it can be loaded by the multimodal class
+                if field_value.blob:
+                    file_path = os.path.join(tmp_dir, field_name_data_class)
+                    with open(file_path, 'wb') as f:
+                        f.write(base64.b64decode(field_value.blob.encode('utf-8')))
+                    field_value.blob = None
+                    field_value.uri = file_path
```
jupyterjazz, line 48: wondering if we could ignore this and directly pass the blob to the dataclass, but it would complicate things I guess

jupyterjazz (approved):
LGTM! :rocket:

deployment/bff/app/v1/routers/helper.py
joschkabraun: No, `Image` expects either `tensor` or `uri`. We could convert blob to image tensor, but then we would have to create different conversions for `Video` and any other modalities in the future will allow blob as possible value (e.g. `Audio`, `3D`, etc.).

deployment/bff/app/v1/routers/helper.py
```diff
@@ -1,90 +1,62 @@
...
-def field_dict_to_doc(field_dict: dict) -> Document:
+
+def field_dict_to_mm_doc(
+    field_dict: dict, data_class, field_names_to_dataclass_fields={}
+) -> Document:
     """Converts a dictionary of field names to their values to a document.
 
     :param field_dict: key-value pairs of field names and their values
-    :return: document
+    :param data_class: data class which shall be used to create the document
```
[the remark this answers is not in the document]
joschkabraun: Let me know if the update is helpful.

deployment/bff/app/v1/routers/helper.py
jupyterjazz: I was more thinking about using type `Blob` instead of `Image` when we have a blob.

deployment/bff/app/v1/routers/helper.py
joschkabraun: Good catch, thanks!

PR #823, opened 2022-12-27
fix: removed output modality from playground

IyadhKhalfallah:
**Removed output modality from playground**

---

- [x] Removed dependencies on output modality in the playground

Screenshot 2022-12-27 at 15 51 38


deployment/playground/playground.py
```diff
@@ -100,13 +100,8 @@ def deploy_streamlit():
...
         nav_to(redirect_to)
     else:
-        da_img, da_txt = load_example_queries(params.data, params.output_modality)
+        da_img, da_txt = load_example_queries(params.data)
 
-        if params.output_modality == 'text':
-            # censor words in text incl. in custom data
-            from better_profanity import profanity
-
-            profanity.load_censor_words()
```
jupyterjazz, line 109: Do we remove profanity? It's still used in the codebase so if we're not using it we should remove it from other files too (playground.py, requirements.py)

deployment/playground/playground.py
```diff
@@ -402,21 +396,11 @@ def render_matches(OUTPUT_MODALITY):
             all_cs = [c1, c2, c3, c4, c5, c6, c7, c8, c9]
 
             for c, match in zip(all_cs, list_matches[st.session_state.page_number]):
-                match.mime_type = OUTPUT_MODALITY
-
-                if OUTPUT_MODALITY == 'text':
-                    render_text_result(match, c)
-
-                elif OUTPUT_MODALITY == 'video':
+                match.mime_type = 'text-or-image-or-video'
```
jupyterjazz, line 399: is mime_type used somewhere? maybe we should remove this too?

deployment/playground/playground.py
```diff
@@ -402,21 +396,11 @@ def render_matches(OUTPUT_MODALITY):
...
-
-                elif OUTPUT_MODALITY == 'text-or-image-or-video':
-                    try:
-                        render_graphic_result(match, c)
-                    except:
-                        render_text_result(match, c)
-                else:
-                    raise ValueError(f'{OUTPUT_MODALITY} not handled')
+                except:
+                    render_text_result(match, c)
```
jupyterjazz, line 403: can you help me understand this try-except logic? i.e. why are we calling render_text_result in case of an exception

jupyterjazz (commented):
Looks good!
Left some comments.
Can you also post a playground pic?

deployment/playground/playground.py
IyadhKhalfallah: Actually, that was the second import of profanity in the file. So removing this will clean the code

deployment/playground/playground.py
IyadhKhalfallah: Yes it is used [here](https://github.com/jina-ai/now/blob/main/deployment/playground/playground.py#L463)

deployment/playground/playground.py
IyadhKhalfallah: I kept the same logic as before, this was inside the OUTPUT_MODALITY == 'text-or-image-or-video' condition. I'm not sure that it should be changed, but please let me know if you see any inconsistency!

deployment/playground/playground.py
jupyterjazz: Makes sense about the import but you also removed this line `profanity.load_censor_words()`?

deployment/playground/playground.py
jupyterjazz: Yes, but before we had default mime_types e.g. text, image. Now, there's only one custom type (text-or-image-or-video) which I don't think docarray will understand, and thus use it.

deployment/playground/playground.py
jupyterjazz: I misread it, thanks for explaining

deployment/playground/playground.py
IyadhKhalfallah: Makes sense, I just added that back on top of the file! Nice catch though

joschkabraun (approved):
LGTM!
We need to also adjust the playground URL returned from the NOW package in `run_all_k8s`.

deployment/playground/playground.py
IyadhKhalfallah: @jupyterjazz Could you please give me more context about this? I don't understand where the problem can be. Also, I tested this in the UI and it seems to be working as expected. In which case will this not work?

jupyterjazz (approved):
LGTM!

PR #835, opened 2022-12-28
refactor: schema detection takes up to a minute

IyadhKhalfallah:
---

- [x] Test added
- [x] Benchmark: from 65s to less than 1s locally


now/common/detect_schema.py
```diff
@@ -148,10 +148,37 @@ def set_field_names_from_docarray(user_input: UserInput, **kwargs):
         raise ValueError('DocumentArray does not exist or you do not have access to it')
 
 
-def identify_folder_structure(file_paths: List[str], separator: str) -> str:
+def identify_s3_folder_structure(bucket, folder_prefix) -> str:
```
azayz, line 151: use only one function `identify_folder_structure` but add argument `is_s3_dataset` to know which code to use
 also add type hints

now/common/detect_schema.py
```diff
@@ -148,10 +148,37 @@ def set_field_names_from_docarray(user_input: UserInput, **kwargs):
...
     """This function identifies the folder structure.
     It works with a local file structure or a remote file structure.
 
+    :param bucket: s3 bucket
+    :param folder_prefix: prefix of the s3 folder
+    :return: if all the files are in the same folder then returns 'single_folder' else returns 'sub_folders'
+    :raises Exception: if a file exists in sub_folder structure
+    """
+
+    structure_identifier = list(bucket.objects.filter(Prefix=folder_prefix).limit(2))[
```
azayz, line 161: add comments to explain what are you trying to get here, it's hard to understand just by looking at the code

now/common/detect_schema.py
```diff
@@ -225,21 +252,24 @@ def set_field_names_from_s3_bucket(user_input: UserInput, **kwargs):
...
-    )  # user has to provide the folder where folder structure begins
-
-    objects = list(bucket.objects.filter(Prefix=folder_prefix))
-    file_paths = [
-        obj.key
-        for obj in objects
-        if not obj.key.endswith('/') and not obj.key.split('/')[-1].startswith('.')
-    ]
-    folder_structure = identify_folder_structure(file_paths, '/')
+    bucket, folder_prefix = get_s3_bucket_and_folder_prefix(user_input)
```
azayz, line 255: type hints

now/common/detect_schema.py
```diff
@@ -225,21 +252,24 @@ def set_field_names_from_s3_bucket(user_input: UserInput, **kwargs):
...
+    # user has to provide the folder where folder structure begins
+    try:
+        first_file = list(bucket.objects.filter(Prefix=folder_prefix).limit(2))[1]
```
azayz, line 258: same comment as above

azayz (changes requested):

azayz (approved):
LGTM after CI green, would love to see a small benchmark of the speed up

now/common/detect_schema.py
```diff
@@ -44,10 +44,10 @@
~                search_fields_modalities[field_name] = MODALITIES_MAPPING[modality]
~                break
~            elif field_name == 'uri' and field_value.split('.')[-1] in file_types:
~                search_fields_modalities[field_name] = MODALITIES_MAPPING[modality]
~                break
```
joschkabraun: Let's remove this function if it's only used for s3 and is so short

now/common/detect_schema.py
```diff
@@ -145,31 +145,18 @@
~            user_input.filter_fields_modalities,
~        ) = _extract_field_candidates_docarray(response)
~    else:
         raise ValueError('DocumentArray does not exist or you do not have access to it')
 
 
-def identify_folder_structure(file_paths: List[str], separator: str) -> str:
+def identify_folder_structure(first_file: str, folder_prefix: str) -> str:
```
joschkabraun, line 151: I would add the dataset path in quotation marks; that helps with understanding the error.

now/common/detect_schema.py
```diff
@@ -145,31 +145,18 @@
~            user_input.filter_fields_modalities,
~        ) = _extract_field_candidates_docarray(response)
```
joschkabraun: what is `i` used for? I think we need to increment `i` below?

now/common/detect_schema.py
```diff
@@ -145,31 +145,18 @@
...
~    else:
         raise ValueError('DocumentArray does not exist or you do not have access to it')
 
 
-def identify_folder_structure(file_paths: List[str], separator: str) -> str:
+def identify_folder_structure(first_file: str, folder_prefix: str) -> str:
     """This function identifies the folder structure.
-    It works with a local file structure or a remote file structure.
 
-    :param file_paths: list of file paths
```
joschkabraun, line 155: we can limit this to the first few hundred fails maybe?

now/data_loading/data_loading.py
```diff
@@ -253,15 +255,24 @@ def _list_files_from_s3_bucket(
...
+    try:
+        # gets the first file in an s3 bucket, index 0 is reserved for the root folder name
+        first_file = list(bucket.objects.filter(Prefix=folder_prefix).limit(2))[1].key
+        i = 2
+        while first_file.split('/')[-1].startswith('.'):
+            first_file = list(bucket.objects.filter(Prefix=folder_prefix).limit(i + 1))[
+                i
+            ].key
+    except Exception as e:
+        raise Exception(f'Empty folder, data is missing.')
```
joschkabraun, line 267: this code looks familiar, let's wrap it in a function

joschkabraun (changes requested):
Nice speed-up in detection!

now/common/detect_schema.py
```diff
@@ -225,21 +212,30 @@ def set_field_names_from_s3_bucket(user_input: UserInput, **kwargs):
...
+    try:
+        # gets the first file in an s3 bucket, index 0 is reserved for the root folder name
+        first_file = list(bucket.objects.filter(Prefix=folder_prefix).limit(2))[1].key
+        i = 2
+        while first_file.split('/')[-1].startswith('.'):
+            first_file = list(bucket.objects.filter(Prefix=folder_prefix).limit(i + 1))[
+                i
+            ].key
+    except Exception as e:
+        raise Exception(f'Empty folder, data is missing.')
```
joschkabraun, line 226: I would add the dataset path in quotation marks; that helps with understanding the error.

now/common/detect_schema.py
```diff
@@ -225,21 +212,30 @@ def set_field_names_from_s3_bucket(user_input: UserInput, **kwargs):
...
+
+    folder_structure = identify_folder_structure(first_file, folder_prefix)
     if folder_structure == 'single_folder':
+        objects = list(bucket.objects.filter(Prefix=folder_prefix))
```
joschkabraun, line 230: we can limit this to the first few hundred fails maybe?

now/common/detect_schema.py
```diff
@@ -225,21 +212,30 @@ def set_field_names_from_s3_bucket(user_input: UserInput, **kwargs):
...
-        for obj in objects
-        if not obj.key.endswith('/') and not obj.key.split('/')[-1].startswith('.')
-    ]
-    folder_structure = identify_folder_structure(file_paths, '/')
+    bucket, folder_prefix = get_s3_bucket_and_folder_prefix(user_input)
+    # user has to provide the folder where folder structure begins
+    try:
+        # gets the first file in an s3 bucket, index 0 is reserved for the root folder name
+        first_file = list(bucket.objects.filter(Prefix=folder_prefix).limit(2))[1].key
+        i = 2
```
joschkabraun, line 220: I think we need to increment `i`, don't we?

now/common/detect_schema.py
```diff
@@ -63,6 +63,21 @@ def _create_candidate_index_filter_fields(field_name_to_value):
     return index_fields_modalities, filter_field_modalities
 
 
+def get_first_file_in_folder_structure(bucket, folder_prefix, dataset_path):
```
joschkabraun, line 66: Let's call it `get_first_file_in_folder_structure_s3`

joschkabraun (approved):
LGTM!
Nice, speed-up.

PR #851, opened 2023-01-01
feat: integrate elastic

florian-hoenicke:
- script to start elastic cluster within the executor
- use elastic search user role since elastic can not be started as root
- use elastic instead of qdrant
- remove qdrant
- adjust bff to work with root level responses
- workaround for parameters, core flattens params when using k8s 
- elastic executor retries to connect to the cluster
- unification executor names since this is crucial for docs_map (thanks @violenil) 
- deactivate threshold functionality of playground


now/common/flow/flow-clip.yml
```diff
@@ -25,7 +25,7 @@ executors:
         memory: 1G
     env:
       JINA_LOG_LEVEL: DEBUG
-  - name: encoder_clip
+  - name: encoderclip
```
violenil, line 28: Can we add this name to our constants to avoid this kind of issue in the future? Then `common_setup` can use this to set an env variable.

violenil:
Please also show a screenshot of a working customer deployment :)

florian-hoenicke:
local run
image

now/common/flow/flow-clip.yml
florian-hoenicke: This is tackled in @joschkabraun's pr that rewrites the flow generation

PR #860, opened 2023-01-02
refactor: remove code

florian-hoenicke:
removes unused code from the preprocessor


florian-hoenicke:
I thought this would be easy to remove but looks like it needs further investigation

PR #872, opened 2023-01-05
Test include bff in offline flow

florian-hoenicke:
search endpoint of bff in offline flow
multi-modal return values in bff


jupyterjazz (approved):
LGTM!

I also just stumbled upon the bff issue, great that you fixed it!

Issue #852, opened 2023-01-01
bug: loading from local folder failed


florian-hoenicke:
When loading gifs from a local folder, we get the following error:

UAC
- [ ] end to end test checking that it works

IyadhKhalfallah:
@florian-hoenicke Could you please upload the error? Also, is it possible to send me some data to try with?

PR #878, opened 2023-01-05
fix: recreate default semantic scores for every query

joschkabraun:
Symptom: Elastic indexer expected `query_text` in search document after sending a search request with text; searching with `blob` (in `query_image`) didn't work anymore afterwards.

Issue: the elastic indexer uses `self.default_semantic_scores` when searching. Attribute `self.default_semantic_scores` is set after one search query and not updated anymore. So `search` doesn't work anymore with new queries with a different search field.

Solution:
- remove `self.default_semantic_scores` as it’s only used on `/search`
- recreate `semantic_scores` for every request

Search with text:
Screenshot 2023-01-05 at 16 26 27

Search with image (note changed order of results due to similar visual features)
Screenshot 2023-01-05 at 16 26 52

Can also confirm that #852 isn't an issue (anymore) as data were from local folder.

---

- [ ] Test added
- [ ] Documentation adjusted


azayz (approved):
Nice catch

makram93 (approved):
LGTM! 🚀

PR #880, opened 2023-01-05
fix: filter out index fields in docarray

IyadhKhalfallah:
---

- [ ] Test added
- [ ] Documentation adjusted


now/app/search_app/app.py
```diff
@@ -156,7 +156,7 @@ def indexer_stub(user_input: UserInput, encoder2dim: Dict[str, int]) -> Dict:
                     [
                         encoder_name,
                         dim,
-                        list(user_input.field_names_to_dataclass_fields.values()),
+                        user_input.index_fields,
```
joschkabraun, line 159: Why this change? I don't think it works.

now/data_loading/data_loading.py
```diff
@@ -39,6 +157,7 @@ def load_data(user_input: UserInput, data_class=None) -> DocumentArray:
     elif user_input.dataset_type == DatasetTypes.DEMO:
         print('⬇  Download DocumentArray dataset')
         da = DocumentArray.pull(name=user_input.dataset_name, show_progress=True)
+        da = get_da_with_index_fields(da, user_input)
     da = set_modality_da(da)
```
joschkabraun, line 161: let's handle demo data with the same function as we do with docarray usually

joschkabraun (changes requested):

now/data_loading/data_loading.py
```diff
@@ -1,26 +1,141 @@
...
+        modality_value = document.text
+    elif document.tensor:
+        modality_value = document.tensor
+    else:
+        document.summary()
+        raise Exception(f'Document {document} cannot be transformed.')
+    return {field: modality_value}
+
+
+def _field_dict_to_mm_doc(
```
joschkabraun, line 50: Please refactor this function and the one defined in the BFF to avoid duplicate code.

now/data_loading/data_loading.py
```diff
@@ -1,26 +1,141 @@
...
+    create_dataclass_fields_file_mappings,
+)
 from now.data_loading.elasticsearch import ElasticsearchExtractor
 from now.log import yaspin_extended
 from now.now_dataclasses import UserInput
-from now.utils import sigmap
+from now.utils import docarray_typing_to_modality_string, sigmap
+
+
+def _construct_tags(document: Document, field) -> Document:
```
joschkabraun, line 32: Just so we don't forget it: As in the other PR, this can be written as a one-liner.

now/data_loading/data_loading.py
```diff
@@ -1,26 +1,141 @@
...
+                    data_class_kwargs[field_name_data_class] = field_value.uri
+                elif field_value.text:
+                    data_class_kwargs[field_name_data_class] = field_value.text
+                elif field_value.tensor:
+                    data_class_kwargs[field_name_data_class] = field_value.tensor
+
+            doc = Document(data_class(**data_class_kwargs))
+        except Exception as e:
+            raise Exception(
+                f'Not a correctly encoded request. Please see the error stack for more information. \n{e}'
```
joschkabraun, line 104: This exception doesn't make senes in this context.

now/data_loading/data_loading.py
```diff
@@ -1,26 +1,141 @@
...
+            )
+
+    return doc
+
+
+def get_da_with_index_fields(da: DocumentArray, user_input: UserInput):
+    dataclass = create_dataclass(user_input)
```
joschkabraun, line 111: Instead of creating `dataclass` here. We should create at the top-level of `data_loading` as it's now used for all `DatasetTypes`. Then we can also remove the code for creating it in `run_backend.run`

now/data_loading/data_loading.py
```diff
@@ -1,26 +1,141 @@
...
+            doc = Document(data_class(**data_class_kwargs))
+        except Exception as e:
+            raise Exception(
+                f'Not a correctly encoded request. Please see the error stack for more information. \n{e}'
+            )
+
+    return doc
+
+
+def get_da_with_index_fields(da: DocumentArray, user_input: UserInput):
```
joschkabraun, line 110: stylistic suggestion: let's have `load_data` as first function in this file.

joschkabraun (changes requested):
Nice work! 🚀 
Some refactor suggestions

now/data_loading/data_loading.py
```diff
@@ -59,20 +81,24 @@ def add_metadata_to_da(da, user_input):
...
+#         for field in non_index_fields:
+#             non_index_field_doc = getattr(d, field, None)
+#             d.tags.update(
+#                 {
+#                     field: non_index_field_doc.content
+#                     if non_index_field_doc.content
+#                     else non_index_field_doc.uri
+#                 }
+#             )
+#     return da
```
IyadhKhalfallah, line 101: This is temporarily removed, but it should be updated after solving the failing CI tests.

now/utils.py
```diff
@@ -140,6 +152,86 @@ def get_info_hubble(user_input):
...
+                    file_ending = get_blob_format(base64_decoded)
+                    file_path = os.path.join(
+                        tmp_dir, field_name_data_class + '.' + file_ending
+                    )
+                    with open(file_path, 'wb') as f:
+                        f.write(base64_decoded)
+                    field_value.blob = None
+                    field_value.uri = file_path
+                if field_value.content is not None:
+                    data_class_kwargs[field_name_data_class] = field_value.content
```
joschkabraun, line 209: these `elif` shouldn't be necessary by the construction of `field_value.content`

now/utils.py
```diff
@@ -140,6 +152,86 @@ def get_info_hubble(user_input):
     return response['data'], client.token
 
 
+def get_blob_format(blob_value):
```
joschkabraun, line 155: rename to something like `extract_supported_file_type_from_blob` or so?

now/app/search_app/app.py
```diff
@@ -156,7 +156,11 @@ def indexer_stub(user_input: UserInput, encoder2dim: Dict[str, int]) -> Dict:
                     [
                         encoder_name,
                         dim,
-                        list(user_input.field_names_to_dataclass_fields.values()),
+                        [
+                            v
+                            for k, v in user_input.field_names_to_dataclass_fields.items()
+                            if k in user_input.index_fields
```
joschkabraun, line 162: I believe by construction of `field_names_to_dataclass_fields` the `if` condition shouldn't be necessary

now/run_backend.py
```diff
@@ -46,9 +46,7 @@ def run(
         user_input.field_names_to_dataclass_fields = {
```
joschkabraun, line 46: this attribute will be set by `create_dataclass`

now/run_backend.py
```diff
@@ -46,9 +46,7 @@ def run(
...
             field: field for field in user_input.index_fields
         }
-        data_class = None
-    else:
-        data_class = create_dataclass(user_input)
+    data_class = create_dataclass(user_input)
```
joschkabraun, line 49: let's move `create_dataclass` into `load_data` as `data_class` isn't actally used in `app.setup`

tests/unit/data_loading/test_data_loading.py
```diff
@@ -106,19 +105,6 @@ def test_da_local_path_image_folder(image_resource_path: str):
         assert doc.chunks[0].content is not None
 
 
-def test_da_custom_ds(da: DocumentArray):
```
joschkabraun, line 109: why was this test removed?

now/utils.py
```diff
@@ -140,6 +152,86 @@ def get_info_hubble(user_input):
...
+        )
+    return file_ending
+
+
+def field_dict_to_mm_doc(
+    field_dict: dict,
+    data_class: type,
+    field_names_to_dataclass_fields={},
+    bff_use=False,
+) -> Document:
```
joschkabraun, line 176: let's add the documentation back of this function

now/utils.py
joschkabraun: The dictionary `field_dict` is now of type `Dict[str,Union[Document,ModalityModel]]` we should make this more clear. In this case `field_value.content or field_value.uri` should suffice.

joschkabraun (changes requested):

now/utils.py
```diff
@@ -14,14 +15,22 @@
...
 from pyfiglet import Figlet
 
 from now.thirdparty.PyInquirer.prompt import prompt
 
+SUPPORTED_FILE_TYPES = {
+    Text: ['txt', 'md'],
+    Image: ['jpg', 'jpeg', 'png', 'bmp', 'tiff', 'tif'],
+    Video: ['gif'],
+}
+
```
joschkabraun, line 33: this should be in `constants` / I believe is already present

requirements.txt
```diff
@@ -15,4 +15,6 @@ python-dotenv
 boto3==1.26.43
 elasticsearch==8.4.1
 pydantic==1.9.2
-httpx==0.23.2
\ No newline at end of file
+httpx==0.23.2
+av==9.1.1
+filetype
```
joschkabraun, line 20: let's fix `filetype` as well

joschkabraun (changes requested):
Great solution! Some comments

now/app/search_app/app.py
```diff
@@ -51,7 +51,7 @@
```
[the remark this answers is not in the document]
joschkabraun: So it will be fixed, with the other PR?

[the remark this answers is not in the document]
IyadhKhalfallah: The other PR is aiming to remove the dependency of filter fields from `create_dataclass`. I think it should be fixed by then (this PR #930 )

now/data_loading/data_loading.py
```diff
@@ -48,6 +49,21 @@
...
+        temp_da = []
+        for field in user_input.index_fields:
+            _index_field_doc = getattr(d, field)
+            temp_da.append(_index_field_doc)
+        d.chunks = DocumentArray(temp_da)
+        d._metadata['multi_modal_schema'] = {
+            k: d._metadata['multi_modal_schema'][k] for k in user_input.index_fields
+        }
+        for i, k in enumerate(user_input.index_fields):
+            d._metadata['multi_modal_schema'][k]['position'] = float(i)
```
joschkabraun, line 63: Isn't the type `int`?

now/deployment/deployment.py
```diff
@@ -14,18 +13,18 @@
~
~
~def status_wolf(flow_id):
~    loop = get_or_create_eventloop()
~    return loop.run_until_complete(CloudFlow(flow_id=flow_id).status)
~
~
```
joschkabraun: Isn't `gateway` actually accurate?

now/deployment/deployment.py
```diff
@@ -34,7 +33,10 @@ def list_all_wolf(status='Serving', namespace='nowapi'):
     jflows = loop.run_until_complete(CloudFlow().list_all(phase=status))['flows']
     # Transform the JCloud flow response to a much simpler list of dicts
     for flow in jflows:
-        flows.append({'id': flow['id'], 'name': flow['status']['endpoints']['gateway']})
```
joschkabraun, line 37: Isn't `gateway` actually accurate?

now/data_loading/data_loading.py
```diff
@@ -48,6 +49,21 @@ def load_data(user_input: UserInput, data_class=None) -> DocumentArray:
...
```
jupyterjazz, line 63: I think we can do both in one iteration.
Also, instead of `k` can you name it something else? `field` for example

now/data_loading/data_loading.py
```diff
@@ -48,6 +49,21 @@ def load_data(user_input: UserInput, data_class=None) -> DocumentArray:
     return da
 
 
+def get_da_with_index_fields(da: DocumentArray, user_input: UserInput):
+    for d in da:
+        temp_da = []
+        for field in user_input.index_fields:
+            _index_field_doc = getattr(d, field)
+            temp_da.append(_index_field_doc)
+        d.chunks = DocumentArray(temp_da)
```
jupyterjazz, line 58: I think this should also work?
```suggestion
        d.chunks = [getattr(d, field) for field in user_input.index_fields]
```

jupyterjazz (commented):
Nice! 
left some comments
can you also add a unit test?

now/data_loading/data_loading.py
IyadhKhalfallah: @joschkabraun This is actually of type float according to the existing format

now/deployment/deployment.py
IyadhKhalfallah: This should actually depend on the executor name. It was failing and I checked it with @makram93 , this was the fix. @makram93 maybe you have more to add here?

now/data_loading/data_loading.py
IyadhKhalfallah: I tried to combine these but I was more focused on making this more readable. The logic in it can be confusing, what do you think?

now/deployment/deployment.py
joschkabraun: Okay!

now/data_loading/data_loading.py
joschkabraun: Where does it say that? Because executing this snippet, gives me that the type is `int`
```python
from docarray import dataclass, Document

from docarray.typing import Text


@dataclass
class MMDoc:
    title: Text
    desc: Text


doc0 = Document(MMDoc(title='title0', desc='desc0'))
print(type(doc0._metadata['multi_modal_schema']['title']['position']))
```

IyadhKhalfallah: @joschkabraun that's a bit confusing! The screenshot below shows the output while debugging. Your approach of checking this seems more correct though, but I'm not sure what's the cause of this
Screenshot 2023-01-24 at 09 33 27

joschkabraun (approved):
Great work! :rocket:

jupyterjazz (approved):
LGTM!

PR #889, opened 2023-01-06
fix: bucket url json

florian-hoenicke:
You can select index fields that are within a json file on s3.
The problem is, that the content can not be loaded on client side.
Moreover, the field name is not known anymore within the preprocessor.
How would the preprocessor then be able to select the right field.
The solution I went for is to put the field names in the tags of the chunk.
I decided against the other option to pass the field names separately since this is against the encapsulation principle in object-oriented design.


florian-hoenicke:
reviewed by @joschkabraun

PR #896, opened 2023-01-09
fix: bug docarray pull

azayz:
---

- [ ] Test added
- [ ] Documentation adjusted


now/utils.py
```diff
@@ -136,6 +136,7 @@ def get_info_hubble(user_input):
             'Your hubble account is not verified. Please verify your account to deploy your flow as admin.'
         )
     user_input.jwt = {'token': client.token}
+    user_input.user_name = response['data']['name']
```
florian-hoenicke, line 139: we could change it to admin_name

florian-hoenicke (changes requested):

jupyterjazz (approved):
LGTM!

let's have 'team-now' as a constant since it's used in so many places

IyadhKhalfallah (approved):
LGTM

now/common/detect_schema.py
```diff
@@ -65,7 +66,7 @@
~            filter_field_candidates_to_modalities[field_name] = field_value.__class__
~
~    if len(index_field_candidates_to_modalities.keys()) == 0:
~        raise ValueError(
```
joschkabraun: Is it possible to use the previously used environment variables?

now/now_dataclasses.py
```diff
@@ -45,6 +45,7 @@ class UserInput(BaseModel):
     deployment_type: Optional[str] = None
     secured: Optional[StrictBool] = None
     jwt: Optional[Dict[str, str]] = None
+    admin_name: Optional[str] = None
```
joschkabraun, line 48: Let's remove this attribute as it can be fetched from hubble SDK.

joschkabraun (changes requested):

now/data_loading/data_loading.py
```diff
@@ -38,7 +38,20 @@ def load_data(user_input: UserInput, data_class=None) -> DocumentArray:
         da = _extract_es_data(user_input=user_input, data_class=data_class)
     elif user_input.dataset_type == DatasetTypes.DEMO:
         print('⬇  Download DocumentArray dataset')
-        da = DocumentArray.pull(name=user_input.dataset_name, show_progress=True)
+        if (
+            'LOCAL_TESTING' in os.environ
```
joschkabraun, line 42: this environment variable

now/now_dataclasses.py
azayz: I think we should still keep this, because this is the same behaviour as the JWT token, both info come from Hubble SDK.

PR #898, opened 2023-01-09
fix: filter construction in bff

### LINKED ISSUE #891: bug: elasticsearch filtering doesn't work
ES indexer doesn't filter tags correctly

Reason is that indexer expects each tag to have an individual field in the es doc where we only store them inside `tags`

Current:
es_doc = {... , tags={color: red, status: finished}, ...}

Expected:
es_doc = {..., tags={color: red}, color=red, status=finished, ...}

We still want to have `tags` inside the doc because it's used by the tags endpoint

jupyterjazz:
Closes #891 

- Filter construction fixed
- Import adjusted
- Unused dependency removed

---

- [x] Test added
- [ ] Documentation adjusted


joschkabraun (approved):
Nice catch! LGTM

PR #906, opened 2023-01-10
fix: add tags to docarrays

IyadhKhalfallah:
---

- [ ] Test added
- [ ] Documentation adjusted

### DIFF
--- a/now/data_loading/data_loading.py
+++ b/now/data_loading/data_loading.py
@@ -17,23 +17,6 @@ from now.now_dataclasses import UserInput
 from now.utils import sigmap
 
 
-def _construct_tags_field(document: Document, field) -> Document:
-    """
-    Create a multimodal docarray structure from a unimodal `Document`.
-    """
-    if document.blob:
-        modality_value = document.blob
-    elif document.uri:
-        modality_value = document.uri
-    elif document.text:
-        modality_value = document.text
-    elif document.tensor:
-        modality_value = document.tensor
-
-        raise Exception(f'Document {document} cannot be transformed.')
-    return {field: modality_value}
-
-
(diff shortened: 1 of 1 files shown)


jupyterjazz (approved):
LGTM!

now/data_loading/data_loading.py
```diff
@@ -17,6 +17,36 @@
...
+    if document.blob:
+        modality_value = document.blob
+    elif document.uri:
+        modality_value = document.uri
+    elif document.text:
+        modality_value = document.text
+    elif document.tensor:
+        modality_value = document.tensor
+    else:
+        document.summary()
```
azayz, line 33: remove document summary

now/data_loading/data_loading.py
```diff
@@ -17,6 +17,36 @@
 from now.utils import sigmap
 
 
+def _construct_tags_field(document: Document, field) -> Document:
```
joschkabraun, line 20: This can be abbreviated to `{field: document.content or document.uri}`, so let's use this one-liner instead of introducing this function.

joschkabraun (changes requested):

now/data_loading/data_loading.py
IyadhKhalfallah: Thanks @azayz will do!

now/data_loading/data_loading.py
IyadhKhalfallah: @joschkabraun thanks for mentioning this, didn't know that document.content actually contains the rest. I'll adjust the changes

joschkabraun (approved):
Nice catch!

PR #915, opened 2023-01-11
refactor: add filter fields to es mapping

jupyterjazz:
---

- [ ] Test added
- [ ] Documentation adjusted


jupyterjazz:
> Just curious, do we have test cases added for this? In end_to_end test we do specify the filter_field but I am not sure if we perform any filtering check over there.

We have an integration test with bff for this but that's a good point, would be nice to check filters in e2e tests too.
Let's merge this now and I'll add that this week.

PR #928, opened 2023-01-12
fix: list endpoint when indexer is empty

makram93:
When there are no documents in the indexer, the '/list' endpoint throws an error.

image

---

- [ ] Test added
- [ ] Documentation adjusted


jupyterjazz (approved):
good catch!

PR #930, opened 2023-01-13
fix: support tags for folder structure

IyadhKhalfallah:
The tags logic for folder structure should be handled while creating the document in subfolders

---

- [ ] Test added
- [ ] Documentation adjusted


now/data_loading/create_dataclass.py
```diff
@@ -154,3 +155,15 @@ def create_dataclass_fields_file_mappings(fields: List, fields_modalities: Dict)
...
+def update_dict_filter_field_types(dict_fields: Dict, filter_fields: List):
+    """
+    Update dict types to python types if it is a filter field
+
+    :param dict_fields: dict to be updated
+    :param filter_fields: provided filter fields
+    """
+    for key, value in dict_fields.items():
+        if key in filter_fields:
+            dict_fields[key] = str
```
makram93, line 169: Why are we setting all the filter field types to `str`? Can they not be float/int types and are these not supported by the indexer?

IyadhKhalfallah: @makram93 The only available modality for filtering in `Text`. So, reverting it to python types would be `str`, right?

jupyterjazz: We can filter integers/floats too

makram93: @IyadhKhalfallah, That is only applicable when checking dataclass field, i.e. , the dc field that are of type `Text` should be eligible for filter, however any fields in the tags are also eligible, same goes for the content in the json files (in case of s3 source)

now/data_loading/create_dataclass.py
```diff
@@ -47,6 +47,7 @@ class DataClass:
     """
     all_modalities = {}
     all_modalities.update(user_input.index_field_candidates_to_modalities)
+    update_dict_filter_field_types(all_modalities, user_input.filter_fields)
```
joschkabraun, line 50: Instead of patching the filter field types here, why setting them all to string when we create them? (So, solving the issue on a more top-level)

joschkabraun (changes requested):

now/data_loading/create_dataclass.py
IyadhKhalfallah: @makram93 @jupyterjazz I was discussing this with @florian-hoenicke on Friday and we agreed that we can only accept text (string) in filters. Could we discuss this again if you don't agree?

now/data_loading/data_loading.py
```diff
@@ -236,9 +237,11 @@ def create_docs_from_subdirectories(
...
                         data = json.load(f)
                     for el, value in data.items():
-                        if el in field_names_to_dataclass_fields.keys():
-                            kwargs[field_names_to_dataclass_fields[el]] = value
-        docs.append(Document(data_class(**kwargs)))
+                        dict_tags[el] = value
+        doc = Document(data_class(**kwargs))
+        if len(dict_tags) > 0:
+            doc.tags.update(dict_tags)
+        docs.append(doc)
```
jupyterjazz, line 244: I think we need to flatten these tags, otherwise `value` might be a dict. That's what we do when we present the tags in the cli

Related to this ticket
https://github.com/jina-ai/now/issues/945

I thought we'd fix it after the refactoring but we might as well do it here, wdyt?

now/data_loading/create_dataclass.py
```diff
@@ -149,21 +149,18 @@ def create_dataclass_fields_file_mappings(fields: List, fields_modalities: Dict)
...
-            ] = f
+            file_mapping_to_dataclass_fields[
+                f
+            ] = f'{docarray_typing_to_modality_string(field_modality)}_{modalities_count[field_modality]}'
             modalities_count[fields_modalities[f]] += 1
         else:
-            dataclass_fields_to_file_mapping[f'filter_{filter_count}'] = f
+            file_mapping_to_dataclass_fields[f] = f'filter_{filter_count}'
             filter_count += 1
-    file_mapping_to_dataclass_fields = {
```
joschkabraun, line 166: nice!

now/data_loading/create_dataclass.py
```diff
@@ -149,21 +149,18 @@ def create_dataclass_fields_file_mappings(fields: List, fields_modalities: Dict)
...
-            dataclass_fields_to_file_mapping[
-                f'{docarray_typing_to_modality_string(field_modality)}_{modalities_count[field_modality]}'
-            ] = f
+            file_mapping_to_dataclass_fields[
+                f
+            ] = f'{docarray_typing_to_modality_string(field_modality)}_{modalities_count[field_modality]}'
             modalities_count[fields_modalities[f]] += 1
         else:
-            dataclass_fields_to_file_mapping[f'filter_{filter_count}'] = f
+            file_mapping_to_dataclass_fields[f] = f'filter_{filter_count}'
```
joschkabraun, line 164: I think we don't want to have `filter_X` in dataclass anymore, do we?

joschkabraun:
Can we also make sure that loading tags for customer works fine?

now/data_loading/create_dataclass.py
IyadhKhalfallah: You're definitely right. I'm pairing with @jupyterjazz about this later to align, since he already worked on removing the `filter_X` logic

IyadhKhalfallah:
> Can we also make sure that loading tags for customer works fine? That is part of the UAC.

I'll test this with lottiefiles and add a screenshot here after making the changes

violenil:
Please also add a short description to this PR :)

joschkabraun:
Let's also make sure that changes from https://github.com/jina-ai/now/pull/880 are reversed.
Let's also check in the end-to-end test that the returned documents contain the right tags from s3.

now/data_loading/create_dataclass.py
IyadhKhalfallah: For more context, this is to be changed in #974

.github/workflows/cd.yml
```diff
@@ -48,6 +48,7 @@ jobs:
           S3_SCHEMA_FOLDER_PATH: ${{ secrets.S3_SCHEMA_FOLDER_PATH }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           S3_CUSTOM_DATA_PATH: ${{ secrets.S3_CUSTOM_DATA_PATH }}
+          S3_CUSTOM_MM_DATA_PATH: ${{ secrets.S3_CUSTOM_MM_DATA_PATH }}
```
joschkabraun, line 51: what's the difference to `S3_CUSTOM_DATA_PATH`?

now/app/base/app.py
```diff
@@ -188,7 +188,11 @@ def preprocess(
     ) -> DocumentArray:
         """Loads and preprocesses every document such that it is ready for indexing."""
         for doc in docs:
+            dict_tags = dict()
             for chunk in doc.chunks:
+                if chunk.tags:
```
joschkabraun, line 193: Why is it needed to update `tags` in the flow/preprocessor from the chunks?

tests/integration/remote/assertions.py
```diff
@@ -155,12 +155,3 @@ def assert_search_custom_s3(host, create_temp_link=False):
 
     response_json = response.json()
     assert len(response_json) == 2
-    for doc in response_json:
```
joschkabraun, line 158: Why do we remove this testing?

joschkabraun (changes requested):

.github/workflows/cd.yml
IyadhKhalfallah: `S3_CUSTOM_DATA_PATH` is a link to a folder structure that is not nested (doesn't test MM structure). The one I added is a subfolder structure

now/app/base/app.py
IyadhKhalfallah: In `maybe_download_from_s3()`, we are adding URIs to the tags in chunks but we are not updating the tags of the document itself. That is why tags are missing in S3 actually. The change here is to get the tags from the chunks and set them back to the document

now/data_loading/create_dataclass.py
```diff
@@ -149,21 +149,18 @@ def create_dataclass_fields_file_mappings(fields: List, fields_modalities: Dict)
...
             filter_count += 1
-    file_mapping_to_dataclass_fields = {
-        v: k for k, v in dataclass_fields_to_file_mapping.items()
-    }
```
jupyterjazz, line 168: since filter field creation happens outside dataclass, it should be removed from here

tests/integration/remote/test_e2e_s3_bucket.py
```diff
@@ -47,3 +47,40 @@ def test_backend_custom_data(
 
     assert_search_custom_s3(host=response['host'], create_temp_link=False)
     assert_search_custom_s3(host=response['host'], create_temp_link=True)
+
+
+@pytest.mark.remote
+@pytest.mark.parametrize('dataset', ['custom_s3_bucket'])
+@pytest.mark.parametrize('query_fields', ['image'])
+def test_backend_custom_mm_data(
```
jupyterjazz, line 55: These two tests are very similar so let's parametrize it

tests/integration/local/test_end_to_end.py
```diff
@@ -4,7 +4,6 @@
 from tests.integration.local.conftest import SEARCH_URL, get_flow, get_request_body
 
 from now.constants import ACCESS_PATHS, Models
-from tests.integration.local.data import artworks_data, elastic_data, pop_lyrics_data
```
jupyterjazz, line 7: this needs to be imported, that's why ci is failing now

jupyterjazz (changes requested):

.pre-commit-config.yaml
```diff
@@ -15,9 +15,4 @@ repos:
     hooks:
       - id: isort
         exclude: ^now/thirdparty/
-  - repo: https://github.com/hadialqattan/pycln
-    rev: "v1.2.5"
-    hooks:
-      - id: pycln
-        args: [ --config=pyproject.toml ]
-        exclude: ^now/thirdparty/
```
IyadhKhalfallah, line 23: Importing the data does not pass with this pre-commit. I'll add that back when required

tests/integration/local/test_end_to_end.py
IyadhKhalfallah: This was removed automatically in the pre-commit, thanks 🎉

tests/integration/remote/test_e2e_s3_bucket.py
```diff
@@ -14,10 +14,26 @@
...
+        ),
+    ],
+)
 @pytest.mark.parametrize('query_fields', ['image'])
 def test_backend_custom_data(
     start_bff,
     dataset: str,
+    dataset_path: str,
+    index_fields: list,
+    model_key: str,
```
jupyterjazz, line 35: we can drop `model_key` argument and assign add it to kwargs like `field + '_model' for field in index_fields`

now/data_loading/create_dataclass.py
jupyterjazz, line 164: there's no need to have filter fields in the `file_mapping_to_dataclass_fields` anymore because we don't create them in the dataclass

now/app/base/app.py
```diff
@@ -188,7 +188,11 @@ def preprocess(
...
+                    for key, value in chunk.tags.items():
+                        dict_tags.update({key: value})
```
jupyterjazz, line 195:
```suggestion
                    dict_tags.update(chunk.tags)
```

now/app/base/app.py
jupyterjazz: can't we do this inside `maybe_download_from_s3()`? i.e. only when we actually use that function

jupyterjazz (changes requested):
thanks for making the changes!
left some comments

now/data_loading/create_dataclass.py
IyadhKhalfallah: @jupyterjazz I think this is a bigger change and will be addressed in #974

IyadhKhalfallah: Removing this will propagate some errors that need to be solved

jupyterjazz: Oh ok, I thought that was included in the scope of this ticket. 
Currently I'm working on a bug which is related to removing filter fields from dataclass and I'll do the removal then if that's ok for you

IyadhKhalfallah: @jupyterjazz Sure, thanks 🙏

now/utils.py
```diff
@@ -246,6 +246,15 @@ def get_bucket(uri, aws_access_key_id, aws_secret_access_key, region_name):
     return bucket
 
 
+def update_tags(d):
+    dict_tags = dict()
+    for chunk in d.chunks:
+        if chunk.tags:
+            dict_tags.update(chunk.tags)
+    d.tags.update(dict_tags)
```
jupyterjazz, line 254: Thanks for addressing the comments!
One last thing from my side:

I think we can do this
```suggestion
    for chunk in d.chunks:
        d.tags.update(d.chunks.tags)
```
but now that I think about it, what will happen if there is more than one chunk with uris? One of them will be lost, right? We would update the dict first with `{uri: first_uri}`, and then `{uri: second_uri}`, do I understand it correctly?

Also, maybe it's simpler to do this update outside the threads? We could just iterate over the docs after line 285 in `maybe_download_from_s3` and update `d.tags`

florian-hoenicke: Why does the chunk have tags?

IyadhKhalfallah: @florian-hoenicke I am not sure how we are using that, I kept it just in case. But I think we can remove it from being there

jupyterjazz: > Why does the chunk have tags?

We keep the original S3 uri like that. Not sure if it's needed though.
https://github.com/jina-ai/now/blob/0b05a264ca629f9e107e4adba9f3b744000a3b2d/now/utils.py#L224

> what will happen if there is more than one chunk with uris?

Also, looking at the code, if we have more than one chunk with S3 uri one of them will get lost but I don't think that's a big deal rn. 

I think we can just iterate over the `docs` in the end of `maybe_download_from_s3` and update tags:
```python
for d in docs:
    for c in d.chunks:
        d.tags.update(c.tags)
```

jupyterjazz: Or maybe we can even do it on this iteration
https://github.com/jina-ai/now/blob/0b05a264ca629f9e107e4adba9f3b744000a3b2d/now/utils.py#L262
so that we don't have to iterate over the docs twice

now/data_loading/data_loading.py
```diff
@@ -256,11 +257,12 @@ def create_docs_from_subdirectories(
                             kwargs[field] = file_full_path
                 else:
                     with open(file_full_path) as f:
-                        data = json.load(f)
+                        data = flatten_dict(json.load(f))
                     for el, value in data.items():
-                        if el in field_names_to_dataclass_fields.keys():
-                            kwargs[field_names_to_dataclass_fields[el]] = value
-        docs.append(Document(data_class(**kwargs)))
+                        dict_tags[el] = value
```
jupyterjazz, line 262: I think we can replace this part by

```python
with open(file_full_path) as f:
    dict_tags.update(flatten_dict(json.load(f))
```

so we don't need the last for loop

joschkabraun (approved):
Nice! 🚀

joschkabraun:
Does customer deployment work with latest changes?

jupyterjazz (approved):
Great work! 🚀

now/executor/preprocessor/executor.py
```diff
@@ -83,7 +83,7 @@ def preprocess(self, docs: DocumentArray, *args, **kwargs) -> DocumentArray:
 
             docs = self.app.preprocess(docs)
 
-            # as _maybe_download_from_s3 moves S3 URI to tags['uri'], need to move it back for post-processor & accurate
+            # As _maybe_download_from_s3 moves S3 URI to tags['uri'], need to move it back for post-processor & accurate
```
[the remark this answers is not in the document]
IyadhKhalfallah: Thankfully your PR solves this @jupyterjazz 🤣

IyadhKhalfallah:
> Does customer deployment work with latest changes?

@joschkabraun I just tested and it works as expected 🚀

PR #937, opened 2023-01-16
fix: allow all as an options in filter

makram93:
Allow `All of the above` as an option in the filter fields

When there's only one filter field
image

When multiple filter fields
image

---

- [ ] Test added
- [ ] Documentation adjusted


jupyterjazz (commented):
What will happen if the field name is 'all'?

Also, can you please put a screenshot?

makram93:
>What will happen if the field name is 'all'?
Then it will be treated as the `All of the above` options. I will change it to maybe `__all__` so that it can be unique to some extent

PR #938, opened 2023-01-16
refactor: minor change

jupyterjazz:
---

- [ ] Test added
- [ ] Documentation adjusted

### DIFF
--- a/now/run_backend.py
+++ b/now/run_backend.py
@@ -41,7 +41,6 @@ def run(
     :param kwargs: Additional arguments
     :return:
     """
-
     if user_input.dataset_type in [DatasetTypes.DEMO, DatasetTypes.DOCARRAY]:
         user_input.field_names_to_dataclass_fields = {
             field: field for field in user_input.index_fields


florian-hoenicke (approved):
Great that you found the problematic test.

tests/unit/cli/test_cli.py
```diff
@@ -1,20 +0,0 @@
-import os
-import subprocess
-
-import pytest
-
-
-def test_start_with_parameters():
```
joschkabraun, line 7: Isn't this test essential to actually test if the CLI works?

joschkabraun (changes requested):
Nice finding, but not sure if we should actually remove this test.

tests/unit/cli/test_cli.py
jupyterjazz: Hmm, but e2e tests also use cli, right?
The only reason I removed this is because I couldn't come up with a way to delete the flow. 
Any suggestions on how we can do it?

florian-hoenicke: We could remove it and make a ticket to write a small test just for the cli without any other flow deployment happening

joschkabraun: We can use local deployment option instead of remote.

tests/integration/cli/test_cli.py
```diff
@@ -1,18 +1,30 @@
...
+def test_start_with_parameters(cleanup, deployment_type, dataset):
     """Checks if now can be started from cli"""
     dataset_path = os.path.join(
         os.path.dirname(__file__), '..', '..', 'resources', 'image'
     )
     python_file = os.path.join(os.path.dirname(__file__), '..', '..', '..', 'main.py')
     bash_command = (
-        f"python {python_file} start --app search_app --index_fields .jpg --dataset_path "
-        f"{dataset_path} --dataset_type path --deployment_type remote --flow_name test --secured false"
+        f"python {python_file} start --app {Apps.SEARCH_APP} --index_fields .jpg --dataset_path "
```
joschkabraun, line 25: Do we still need `--app` argument?

joschkabraun: If we do, we need to fix options.

tests/integration/conftest.py
```diff
@@ -34,3 +42,45 @@ def with_hubble_login_patch(mocker: MockerFixture) -> None:
...
+                if 'host' not in flow_details:
+                    print('nothing to clean up')
+                    return
+                host = flow_details['host']
+                flow_id = get_flow_id(host)
+                terminate_wolf(flow_id)
+            else:
+                print('\nDeleting local cluster')
+                kwargs = {
+                    'app': Apps.SEARCH_APP,
```
joschkabraun, line 66: is `app` still needed?

joschkabraun (approved):
Nice work!  🚀

PR #942, opened 2023-01-17
feat: expose semantic scores

violenil:
This PR exposes the `semantic_scores` parameter in the BFF and playground. This parameter is already supported in the indexer, but only receives default values currently. The aim is to allow the user to customize this semantic scoring by adding their own field combinations and weights. 

After changes, playground looks like this:
Screenshot 2023-01-20 at 08 26 43

---

- [x] Test added
- [ ] Documentation adjusted


deployment/playground/playground.py
```diff
@@ -102,22 +102,46 @@ def deploy_streamlit():
...
+                except Exception as e:
+                    print("Filters couldn't be loaded from the endpoint properly.", e)
+                    st.session_state.filters = 'notags'
+        if not st.session_state.index_fields_dict:
+            # get index fields from user input
             try:
                 if params.secured.lower() == 'true':
                     response = client.post(
-                        on='/tags',
+                        on='/get_index_fields',
```
azayz, line 128: same code for tags and get_index_fields endpoints can we refactor into one function ?

deployment/playground/playground.py
```diff
@@ -287,9 +311,73 @@ class UI:
...
+        field_mod == 'text'
+        for field_mod in list(st.session_state.index_fields_dict.values())
+    ):
+        bm25.button('Add bm25 score', key='bm25', on_click=toggle_bm25_slider)
+        if st.session_state.show_bm25_slider:
+            bm25_weight = st.slider(
+                label='Adjust bm25 weight',
+                min_value=0.0,
+                max_value=1.0,
+                value=0.5,
```
azayz, line 346: default value is 0.5 ? shouldn't it be 1 ?

deployment/playground/playground.py
```diff
@@ -287,9 +311,73 @@ class UI:
...
+            key='index_field_' + str(i),
+        )
+        enc = encoder.selectbox(
+            label='encoder', options=['clip'], key='encoder_' + str(i)
+        )
+        w = weight.slider(
+            label='weight',
+            min_value=0.0,
+            max_value=1.0,
+            value=0.5,
```
azayz, line 371: same here

now/constants.py
```diff
@@ -48,6 +48,7 @@ class DatasetTypes(BetterEnum):
 
 SURVEY_LINK = 'https://10sw1tcpld4.typeform.com/to/VTAyYRpR?utm_source=cli'
 
+TAG_OCR_DETECTOR_TEXT_IN_DOC = '_ocr_detector_text_in_doc'
```
azayz, line 51: this was removed. merge issues ?

azayz (changes requested):

deployment/playground/playground.py
violenil: Yes! Good idea.

deployment/playground/playground.py
violenil: I was pairing with someone (can't remember who) and they suggested 0.5 as default value. Up for discussions though :)

now/constants.py
violenil: ah yes, let me remove. Bad rebase on my part

deployment/playground/playground.py
jupyterjazz: 🚀  I'm the someone 🚀 
Because from the user standpoint I should also be able to increase importance of a certain field by actually increasing it and not by decreasing others, it makes more sense to me

deployment/playground/playground.py
```diff
@@ -615,6 +704,21 @@ def setup_session_state():
...
         st.session_state.filters_set = False
 
+    if 'len_semantic_scores' not in st.session_state:
+        st.session_state['len_semantic_scores'] = 0
+
+    if 'index_fields_dict' not in st.session_state:
+        st.session_state.index_fields_dict = {}
+
+    if 'encoder' not in st.session_state:
+        st.session_state.encoder = 'clip'
```
jupyterjazz, line 714: are we still keeping the encoder option here even though:
1. we should in the end choose the encoder that was used for the index field?
2. we only have 1 encoder now?

deployment/playground/playground.py
violenil: Then I leave it the way it is :)

now/executor/indexer/elastic/es_query_building.py
```diff
@@ -21,7 +12,7 @@
 def generate_semantic_scores(
     docs_map: Dict[str, DocumentArray],
     encoder_to_fields: Dict[str, Union[List[str], str]],
-) -> List[SemanticScore]:
+) -> List[Tuple]:
```
jupyterjazz, line 15: would be nice to have an example in the docstring

jupyterjazz (approved):
LGTM!

deployment/playground/playground.py
violenil: Good question! I would like to leave the encoder in the selection box, as a 'placeholder', as we have the model selection ticket already in progress by @azayz .

now/executor/indexer/elastic/es_query_building.py
violenil: Yes, will amend. I removed `SemanticScore` and simplified to `Tuple` because we can't send a named tuple in the flow. Will provide an example from the tests in the docstring.

deployment/playground/playground.py
jupyterjazz: Hmm, still feels like an unnecessary complexity to me. If there's more than one encoder for an index field, why not encode the query with all of them.. I'm not sure I see the benefit users can get with choosing the specific encoder. But it's ok, we can go with this and maybe revisit the issue later 👍

now/executor/indexer/elastic/es_query_building.py
violenil: done.

deployment/playground/playground.py
```diff
@@ -295,12 +317,88 @@ class UI:
...
+            st.session_state.semantic_scores['bm25'] = [
+                'query_text',
+                'bm25_text',
+                'bm25',
+                bm25_weight,
+            ]
+
+    for i in range(st.session_state['len_semantic_scores']):
+        query_field, index_field, encoder, weight = st.columns([0.25, 0.25, 0.25, 0.25])
+        q_field = query_field.selectbox(
```
florian-hoenicke, line 373: a function can be used for the select box since it is duplicate code

florian-hoenicke (changes requested):

deployment/playground/playground.py
violenil: Could you please make a suggestion? I am just calling the selectbox method multiple times with, with different parameters.

violenil:
Deployment with customer (lf):

Screenshot 2023-01-23 at 08 07 57

azayz (approved):

Issue #923, opened 2023-01-12
refactor: add tags outside the data class


jupyterjazz:
In case of local folder, S3 and elasticsearch datasources, filters fields are added directly to the dataclass with default names such as `filter_0`, `filter_1`, etc.

This is a problem because we lose the name of the field.

Alternatively, we can not use the exact name in the dataclass because of the code injection issues.

Thus, we have to adjust tags outside the dataclass.
Preferably here
https://github.com/jina-ai/now/blob/011a5bcb32c066eab179d673de2d490807167b39/now/data_loading/data_loading.py#L51

jupyterjazz:
Decided to go with another solution. Iyadh will work on that refactoring.

PR #944, opened 2023-01-17
refactor: adjust filter names

jupyterjazz:
Closes https://github.com/jina-ai/now/issues/923 and https://github.com/jina-ai/now/issues/945

---

- [ ] Test added
- [ ] Documentation adjusted

### DIFF
--- a/now/common/detect_schema.py
+++ b/now/common/detect_schema.py
@@ -198,6 +198,16 @@ def _extract_field_names_single_folder(
     return {file_ending: file_ending for file_ending in file_endings}
 
 
+def read_tags_from_json(data):
+    fields_dict = {}
+    for el, value in data.items():
+        if isinstance(value, dict):
+            fields_dict.update(flatten_dict({el: value}))
+        else:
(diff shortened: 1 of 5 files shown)


now/data_loading/data_loading.py
```diff
@@ -235,7 +242,8 @@ def create_docs_from_subdirectories(
                 else:
                     with open(file_full_path) as f:
                         data = json.load(f)
-                    for el, value in data.items():
+                    tags = read_tags_from_json(data)
```
joschkabraun, line 245: we should add all tags from the json to the document

now/common/detect_schema.py
```diff
@@ -198,6 +198,16 @@ def _extract_field_names_single_folder(
     return {file_ending: file_ending for file_ending in file_endings}
 
 
+def read_tags_from_json(data):
```
joschkabraun, line 201: Is this a fix for `flatten_dict`? 
Then let's move it into `flatten_dict`

jupyterjazz:
changed the direction

PR #965, opened 2023-01-20
fix: enable callbacks

makram93:
Enables the callback. This will help the cloud apps to pass the callback function to get the real-time indexing steps. It also uses `status_callbacks` to print the messages.

---

- [ ] Test added
- [ ] Documentation adjusted


now/run_backend.py
```diff
@@ -159,7 +164,17 @@ def call_flow(
                     parameters=parameters,
                     return_results=return_results,
                     continue_on_error=True,
+                    on_done=kwargs.get('on_done', None),
```
azayz, line 167: what does this do ?

now/data_loading/data_loading.py
```diff
@@ -17,26 +17,31 @@
...
 
     :param user_input: The configured user object. Result from the Jina Now cli dialog.
     :param data_class: The dataclass that should be used for the DocumentArray.
+    :param status_callback: The callback function that should be used to print the status.
     :return: The loaded DocumentArray.
     """
     da = None
     if user_input.dataset_type in [DatasetTypes.DOCARRAY, DatasetTypes.DEMO]:
-        print('⬇  Pull DocumentArray dataset')
+        status_callback('⬇  Pull DocumentArray dataset')
```
azayz, line 33: what could status callback be besides print ?

azayz (commented):
LGTM after comments, great job

now/run_backend.py
makram93: if my kwargs contain something like this: `kwargs['on_done'] = some_func_to_be_called_when_done` where `some_func_to_be_called_when_done` is function, then the Jina Client will invoke that function when the request is completed. Though I don't need it for the cloud apps, I just added it in case it is needed in future

now/data_loading/data_loading.py
makram93: Think of this from the cloud apps perspective where I just call `cli(kwargs)`, now if I want the Jina NOW package to return something back to the caller, then I can implement some function in the now-core and pass this in kwargs to NOW package, i.e., 
`def somefunc(*args, **kwargs):
  ....
  
  kwargs['status_callback']=some_func
  cli(kwargs)`
  
  This way, instead of printing on the terminal, everytime this callback will be invoke and I will get the message

azayz (approved):
LGTM

now/data_loading/data_loading.py
```diff
@@ -17,26 +17,31 @@
 from now.utils import sigmap
 
 
-def load_data(user_input: UserInput, data_class=None) -> DocumentArray:
+def load_data(
+    user_input: UserInput, data_class=None, status_callback=print
```
joschkabraun, line 21: let's rename this to a name which encapsulates its nature as `print` and as a `callback`

joschkabraun (approved):
LGTM!
Only suggestion is to change naming to something more broad. 
Also, should we incorporate this into `yaspin`?

makram93:
>Also, should we incorporate this into yaspin?

I believe we could, but this would be noise or redundant information. Alos, I could only find two places where yaspin is used and it is not much useful.

PR #968, opened 2023-01-23
chore: update pr template

violenil:
This PR just adds another checkbox to the PR template. For some time now, we have made it a requirement that all PR's should make a successful customer deployment and add a screenshot to the PR. So that this is not forgotten, we can add a check to the PR template.


makram93 (approved):

joschkabraun (approved):
Good catch!

jupyterjazz (approved):
Nice 🚀

PR #962, opened 2023-01-20
feat: model choice dialog

azayz:
closes https://github.com/jina-ai/now/issues/759
---

- [x] This PR references an open issue
- [x] Test added
- [x] Documentation adjusted

PR #975, opened 2023-01-25
docs: update api usage

violenil:
This PR updates the documentation with latest usage of the API. It also may include smaller refactoring changes to the documentation (eg. removing FAQ, updating the README).

---

- [ ] Test added (not applicable)
- [x] Documentation adjusted


makram93:
Please also get feedback from @fogx or @saahil once this PR is ready for review, to align with their expectations.

fogx:
I'm not clear on what we want to use this documentation for going forward, but from the content it LGTM :+1:

fogx:
not sure what's going on here, but the previous button is looking wierd
image

violenil:
> not sure what's going on here, but the previous button is looking wierd image

I dont have this issue.

Screenshot 2023-01-26 at 13 09 53

saahil:
Do we need to keep this instruction for local deployment? https://ft-docs-api-update--jina-docs.netlify.app/user-guides/cli_api/#local-execution
The users don't actually have a way to deploy NOW apps locally.

saahil:
For the third element in `List` of  `semantic_scores`, what are the allowed values? In the example, there's `encoderclip` - are there any more types of scores that can be used? 

Other than that, Swagger documentation LGTM 👍

saahil:
Commets about the user documentation
- "_Bootstrap your image search case in minutes._" - This is in the title. However, the search use-cases go beyond just images. Can we update to something like "_Bootstrap your multimodal search app in minutes._" ?
- There are several links on the front page that are referencing static URL's, e.g. "_Jina NOW is production ready_" takes you to `now.jina.ai` domain, when it should be taking to `user_guides/production`. Same for "_guide to loading your data_".

azayz (commented):
Great job but when I see netlify I don't see the images you added
Screenshot 2023-01-26 at 17 30 42

violenil:
> For the third element in `List` of `semantic_scores`, what are the allowed values? In the example, there's `encoderclip` - are there any more types of scores that can be used?

When deploying their application, the user will choose which models they want to use to encode their documents with. This choice currently consists of `encoderclip` and `encodersbert`. I will provide another example using `encodersbert` to demonstrate this option.

violenil:
> * There are several links on the front page that are referencing static URL's, e.g. "_Jina NOW is production ready_" takes you to `now.jina.ai` domain, when it should be taking to `user_guides/production`. Same for "_guide to loading your data_".

@saahil I could not replicate this issue. Could you try again and confirm that this is still the case?

azayz (approved):
LGTM

fogx:
Some further feedback:

The documentation explains how a user can use the CLI tool to deploy search apps. 
Where/how do we promote that feature and how does a user get to this documentation page?

Why would a user [use the Jina client](https://ft-docs-api-update--jina-docs.netlify.app/user-guides/cli_api/#search-requests) instead of the API to perform a search request? This should be answered inside the docs imo.

> When deploying their application, the user will choose which models they want to use to encode their documents with.

how can a user find out which models they used to encode after they deployed? I might not remember or have used the default selection.

I guess this is less of a documentation issue but: if the encoding model that is being used is already set, why does the user have to specify it again? For example, if i want to increase the weight of the title compared to the description, why does the model used to encode that data matter?

The `Guide to loading your data` should be the first user guide and should be linked to much higher in the `Quick start` . For example in the first sentence when we mention that you can use your own custom dataset. 

Is it possible to generate an API reference like in [Jina and docarray](https://docs.jina.ai/api-rst/)?

saahil:
> @saahil When deploying their application, the user will choose which models they want to use to encode their documents with.

@violenil Is this change also part of the current PR? I've used Jina NOW in the last week, and this choice wasn't given to me when deploying through the CLI. I was only given the choice to pick which fields to index, and which ones to use for filtering 
> This choice currently consists of `encoderclip` and `encodersbert`. I will provide another example using `encodersbert` to demonstrate this option.
That would be great @violenil 

> @saahil I could not replicate this issue. Could you try again and confirm that this is still the case?
This is still the case 

image
When I click on this link, I'm taken to https://now.jina.ai/user-guides/production/ 
image

I believe the link should take me to https://ft-docs-api-update--jina-docs.netlify.app/user-guides/production.

violenil:
> @violenil Is this change also part of the current PR? I've used Jina NOW in the last week, and this choice wasn't given to me when deploying through the CLI. I was only given the choice to pick which fields to index, and which ones to use for filtering

@saahil this feature was implemented this week with #962 so it's new (and exciting) :) you can give it a try!

saahil:
>once the PR is merged, we need the viewer to be redirected to the documentation deployed from main (now.jina.ai/user-guides/production), so this is correct imo!

When the PR is merged, that behavior would make sense to me too. But as long as it's in review, if there are changes in the _linked_ page, one cannot view them in this staging env. And that's generally problematic? 
In addition to the general precaution that if we ever move top or second-level domains (e.g. we don't know if Jina NOW will continue to be called the same, even after we decide to call the cloud product Jina Search, and hence might get a different second-level domain), then the hard coded links to now.jina.ai will become dead. If we use relative URLs everywhere, that can be avoided :) 

Last point - these two link targets are currently not the same (they should be)
image

fogx:
```
{
  "host": "grpcs://now-example-best-artworks.dev.jina.ai",
  "port": 443,
  "limit": 10,
  "query": [
    {
      "name": "query_text_0",
      "modality": "text",
      "value": "cute cats"
    }
  ],
  "create_temp_link": false
}
```
The working example is missing ""'s on the false statement

It also fails with the following error:
> raise Exception('Result without content', doc.id, doc.tags)\nException: ('Result without content', 'baa05c8458bfe33246a9f7f39096c087', {'embeddings': {}})\n"

For the dummy example we could point them to the `schema` tab?

The example for semantic scores is still empty
image

Why does the example have `OrderedMap` in it? I think this is missleading, because it's not part of the actual example syntax.
image


The examples are very helpful though 👍

fogx:
When running the query using the client, i get the following error:
```
@dataclass
class DataPoint:
    main_text: Text
    image: Image
    description: Text

query_doc = Document(DataPoint(main_text=query, description="abc", image="../data/cat1.jpg"))response = client2.search(query_doc, parameters={"limit": 9, "filters": {}}, )
```

```
   stacks: "elasticsearch.BadRequestError: BadRequestError(400, \'search_phase_execution_exception\', \'runtime error\')\n"
    executor: "NOWElasticIndexer"
  }
}
exec_endpoint: "/search"
target_executor: ""
python-BaseException
```

fogx:
Another topic that is missing from the documentation is how users can update the index and add/update/remove data from an existing search app.

violenil:
> Another topic that is missing from the documentation is how users can update the index and add/update/remove data from an existing search app.

This is currently not possible, as we don't expose any other endpoints beside `/suggestion` and `/search`. This would first need to be implemented, then we can write docs for this.

fogx:
> This is currently not possible, as we don't expose any other endpoints beside /suggestion and /search. This would first need to be implemented, then we can write docs for this.

Then the information that it is not possible and what users should do instead (redeploy?) should be inside the docs. Also, AFAIK the S3 bucket will get synced after some time?

docs/user-guides/cli_api.md
```diff
@@ -2,27 +2,120 @@
...
+query_doc = Document(Query(text='cat', image='<example-image-uri>.png'))
+
+client = Client(
+    host='grpcs://now-example-bird-species.dev.jina.ai&data=bird-species' # add your own gateway here
+)
 response = client.search(
-        Document(text=search_text), # or in case you send an image: Document(uri=image_path),
-        parameters={"limit": 9, "filter": {}},
+        query_doc,
+        parameters={"limit": 9, "filters": {}},
```
jupyterjazz, line 56:
```suggestion
        parameters={'limit': 9, 'filters': {}},
```

docs/user-guides/cli_api.md
```diff
@@ -2,27 +2,120 @@
...
+  'query': [
+              {'name': 'query_text', 'value': 'black dress', 'modality': 'text'},
+              {'name': 'query_image', 'value': '<insert_image_uri>', 'modality': 'image'},
+           ],
+  'limit': '5',
+  'filters': {}
+}"
+```
+
+Make sure to grab the `host` from the CLI deploment print as shown in the previous section.
```
jupyterjazz, line 96:
```suggestion
Make sure to grab the `host` from the CLI deployment print as shown in the previous section.
```

docs/user-guides/cli_api.md
```diff
@@ -2,27 +2,120 @@
...
+}
+```
+
+Optional parameters:
+
+  - **limit**: set the number of results in the response
+    - example: `5`
+  - **filters**: a dictionary of filters, with filter name as key and target as value
+    - example: `{'tags__color': {'$eq': 'blue'}, 'tags__price': {'$gt': 100}}`
+  - **semantic_scores**: list of semantic scores defining how fields should be compaired and weighted. Each semantic score
```
jupyterjazz, line 114:
```suggestion
  - **semantic_scores**: list of semantic scores defining how fields should be compared and weighted. Each semantic score
```

jupyterjazz (approved):
Great work! 🚀 

left some minor comments

PR #980, opened 2023-01-26
refactor: improve listing files from s3

IyadhKhalfallah:
---

- [ ] Added a test

### DIFF
--- a/now/utils.py
+++ b/now/utils.py
@@ -332,6 +332,29 @@ def get_email():
         return ''
 
 
+def list_s3_objects(bucket_name, prefix, user_input, max_keys=1000):
+    session = boto3.Session(
+        aws_access_key_id=user_input.aws_access_key_id,
+        aws_secret_access_key=user_input.aws_secret_access_key,
+    )
+
+    client = session.client('s3', region_name='eu-west-1')
+    response = client.list_objects_v2(
+        Bucket=bucket_name, Prefix=prefix, MaxKeys=max_keys
+    )
+    while response:
+        for content in response.get('Contents', []):
+            yield content['Key']
+        if not response.get('IsTruncated'):
+            break
+        response = client.list_objects_v2(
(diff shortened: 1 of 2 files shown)


IyadhKhalfallah:
Benchmark: 
- **list_objects_v2 (new method):** 47.67947483062744
- **filter (current method):** 53.21679997444153

now/utils.py
```diff
@@ -332,6 +332,29 @@ def get_email():
         return ''
 
 
+def list_s3_objects(bucket_name, prefix, user_input, max_keys=1000):
+    session = boto3.Session(
+        aws_access_key_id=user_input.aws_access_key_id,
+        aws_secret_access_key=user_input.aws_secret_access_key,
+    )
+
+    client = session.client('s3', region_name='eu-west-1')
```
azayz, line 341: take the region name from user_input

azayz (changes requested):

joschkabraun:
This blogpost details listing 67 billion objects in 169 hours -> so listing 67k objects (lottie) sholud take 0.6 seconds
https://blog.purestorage.com/purely-technical/listing-67-billion-objects-in-1-bucket/

IyadhKhalfallah:
> This blogpost details listing 67 billion objects in 169 hours -> so listing 67k objects sholud take 0.6 seconds https://blog.purestorage.com/purely-technical/listing-67-billion-objects-in-1-bucket/

@joschkabraun I tried that one, and it has actually 2 seconds more latency than the one in the PR. Something that might make this better, as mentioned in the blogpost, is using asyncio and ProcessPoolExecutor to concurrently list each keys. That will actually interfere with the existing event loop that we have for the prompt. Do you think we should give this a shot? If yes, should this be part of this ticket?

azayz (approved):
LGTM after Joschka's comment

PR #989, opened 2023-01-27
fix: semantic scores

violenil:
Playground is failing when trying to search using custom semantic scores.
When model selection was introduced in the dialog in #962 this feature was not added to the playground. Therefore, this PR also adds the model selection for custom semantic scores in the playground.

---

- [ ] Added a test (we don't have any tests for playground, but bff has test for semantic scores feature)

Working customer deployment:

Screenshot 2023-01-31 at 14 30 30

Follow-up ticket for exposing index fields & encoders in BFF: #1008


deployment/playground/playground.py
```diff
@@ -370,11 +398,15 @@ def customize_semantic_scores():
...
+            options=list(st.session_state.field_names_to_dataclass_fields.keys()),
             key='index_field_' + str(i),
         )
+        id_field = st.session_state.field_names_to_dataclass_fields[id_field]
+        encoder_options = get_encoder_options(q_field, id_field)
         enc = encoder.selectbox(
-            label='encoder', options=['clip'], key='encoder_' + str(i)
+            label='encoder',
+            options=encoder_options,
+            key='encoder_' + str(i),
```
azayz, line 409: what's i here ? can we rename to something better ?

now/executor/indexer/elastic/elastic_indexer.py
```diff
@@ -453,6 +453,49 @@ def update_tags(self):
         except Exception:
             self.logger.info(traceback.format_exc())
 
+    @secure_request(on='/get_encoder_to_fields', level=SecurityLevel.USER)
```
azayz, line 456: this is the previous endpoint in preprocessor right ? get_index_fields but with encoder aswell, my question is where should this endpoint be in elastic indexer or preprocessor ?

azayz (commented):
Great Job

deployment/bff/app/v1/models/search.py
```diff
@@ -42,10 +42,10 @@ class SearchRequestModel(BaseRequestModel):
         description='If true, a temporary link to the file is created. '
         'This is useful if the file is stored in a cloud bucket.',
     )
-    semantic_scores: List[Tuple] = Field(
+    semantic_scores: List[List] = Field(
         default=None,
-        description='List of tuples where each tuple contains a query_field, index_field, encoder_name and weight. '
-        'This defines how scores should be calculated for documents.',
+        description='List of lists where each nested list contains a query_field, index_field, encoder_name and weight.'
```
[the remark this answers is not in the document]
violenil: we cannot pass tuples in a flow

deployment/playground/playground.py
violenil: `i` is really just iterating over the number of semantic scores the user has added using the `Add semantic score` button

now/executor/indexer/elastic/elastic_indexer.py
violenil: I moved it to the indexer because the indexer already holds the `encoder_to_fields` (and `document_mappings`) which this endpoint takes advantage of. I don't think it matters too much which executor gets this endpoint, but it was more convenient for me this way. Let me know if you disagree.

deployment/bff/app/v1/models/search.py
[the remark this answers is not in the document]
joschkabraun: But this is the search request model so what we put into the bff

deployment/playground/playground.py
```diff
@@ -323,10 +332,21 @@ def toggle_bm25_slider():
         st.session_state.show_bm25_slider = True
 
 
+def get_encoder_options(q_field: str, id_field: str) -> List[str]:
+    encoders_options = [
+        encoder
+        for encoder in st.session_state.index_fields_dict.keys()
+        if id_field in st.session_state.index_fields_dict[encoder].keys()
+    ]
+    if q_field.startswith('image'):
```
joschkabraun, line 341: Can we use the `MODALITY_TO_MODELS` constant for that?

joschkabraun (commented):
Small comment

deployment/playground/playground.py
```diff
@@ -114,10 +115,19 @@ def deploy_streamlit():
                 index_fields_dict = get_info_from_endpoint(
                     client,
                     params,
-                    endpoint='/get_index_fields',
+                    endpoint='/get_encoder_to_fields',
```
makram93, line 118: Can we expose this endpoint in the bff as well so that the FE can also consume it for cloud apps? They do not use Jina Client, hence they won't be able to get this `encoder_to_fields` for semantic scores.

makram93 (changes requested):

deployment/playground/playground.py
violenil: after some more discussion we decided to move this to another PR, exposing the `/get_index_fields` endpoint in the BFF. Similarly, the `/tags` endpoint should also be exposed in the BFF, so the FE can also consume it. I will create a ticket for this.

deployment/playground/playground.py
violenil: Its not possible to import from `now.constants` currently, since we don't have `now` as dependency for playground/bff

deployment/bff/app/v1/models/search.py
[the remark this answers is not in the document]
violenil: Done. Semantic scores are now tuples :)

deployment/playground/playground.py
violenil: Created ticket #1008

azayz (approved):
LGTM

azayz:
it would be great to see if modifying the semantic scores shows effect on results

deployment/bff/app/v1/routers/search.py
```diff
@@ -48,14 +48,21 @@ def search(data: SearchRequestModel):
         key = 'tags__' + key if not key.startswith('tags__') else key
         query_filter[key] = {'$eq': value}
 
+    if data.semantic_scores:
```
joschkabraun, line 51: To simplify this, let's set `default=[]` in the `SearchRequestModel` definition

deployment/playground/playground.py
joschkabraun: It is dependency for BFF, we can simply add it to the playground.

deployment/playground/playground.py
```diff
@@ -338,14 +358,22 @@ def customize_semantic_scores():
...
+            st.session_state.index_fields_dict[encoder][field]
+            for encoder in st.session_state.index_fields_dict.keys()
+            for field in st.session_state.index_fields_dict[encoder].keys()
+        ]
     ):
         bm25.button('Add bm25 score', key='bm25', on_click=toggle_bm25_slider)
         if st.session_state.show_bm25_slider:
             query_field_selectbox, bm25_slider = st.columns([0.5, 0.5])
             q_field = query_field_selectbox.selectbox(
                 'Select query field for bm25 scoring',
```
joschkabraun, line 371: Can we apply `bm25` to multiple fields independently?

violenil: Yes, you can add bm25 on different query fields (you need to select which query field should be considered for the bm25 calculation). As for the documents, we currently just concatenate all index fields of modality `Text` and use this for bm25 score calculation, ie. you cannot select the index field on the document which should be used for bm25.

now/executor/indexer/elastic/elastic_indexer.py
```diff
@@ -453,6 +453,62 @@ def update_tags(self):
...
+                'title': 'text_0',
+                'picture': 'image_0',
+            }
+        """
+        print(
+            'field_names_to_dataclass_fields: ',
+            self.user_input.field_names_to_dataclass_fields,
+        )
+        print('index_fields: ', self.user_input.index_fields)
+        print(
```
jupyterjazz, line 481: you have some prints left in this fn

deployment/playground/playground.py
```diff
@@ -114,10 +115,19 @@ def deploy_streamlit():
...
-                    endpoint='/get_index_fields',
+                    endpoint='/get_encoder_to_fields',
                     info='index_fields_dict',
                 )
+                field_names_to_dataclass_fields = get_info_from_endpoint(
+                    client,
+                    params,
+                    endpoint='/get_encoder_to_fields',
+                    info='field_names_to_dataclass_fields',
+                )
```
jupyterjazz, line 126: maybe we can avoid fetching the data twice and change `get_info_from_endpoint` so that if `info` is None it returns the whole tags

jupyterjazz (approved):
LGTM!

deployment/playground/playground.py
violenil: done

now/executor/indexer/elastic/elastic_indexer.py
violenil: thanks!

violenil:
> it would be great to see if modifying the semantic scores shows effect on results

@azayz for your request, here's the birds demo data deployment, where I search with two different semantic score definitions:

1) searching with the text query 'cat' only comparing on the `label` index field using SBert encoder.
Screenshot 2023-02-06 at 13 25 02

2) searching with text query 'cat' comparing on both `label` (using SBert) and `image` (using CLIP) fields.
Screenshot 2023-02-06 at 13 25 25

There are definitely differences. Only indexed 50 documents, and perhaps my query is not very suitable, but at least you can see there are changes when semantic scores are customized and adjusted.

makram93 (approved):

PR #992, opened 2023-01-30
fix: elastic persistent workspace

florian-hoenicke:
When a pod is restarted the data is currently lost. In this pr, we persist the data at the wolf workspace assigned to the pod


now/executor/indexer/elastic/elastic_indexer.py
```diff
@@ -95,15 +96,29 @@ def __init__(
             self.es.indices.create(index=self.index_name, mappings=self.es_mapping)
 
     def setup_elastic_server(self):
-        # volume is not persisted at the moment.
         try:
-            subprocess.Popen(['./start-elastic-search-cluster.sh'])
-            self.logger.info('elastic server started')
+            if "K8S_NAMESPACE_NAME" in os.environ:
```
makram93, line 100: Where is this set in the environment? Is it done by the wolf when deploying the flow?

violenil (approved):
LGTM!

now/executor/indexer/elastic/elastic_indexer.py
florian-hoenicke: Yes, they set it in the k8s yaml

makram93 (approved):

Issue #990, opened 2023-01-30
bug: content in the CC level when selecting all index fields with all models available


makram93:
When selecting multiple index fields with multiple models for embedding for each index field (similar to the configuration given in the screenshots, the results return upon search contain values in the `cc` level instead of `c` [as expected in the bff](https://github.com/jina-ai/now/blob/main/deployment/bff/app/v1/routers/search.py#L85-L96) level and this leads to the playground throwing error. I am not sure if this is a bug in the backend or if this is expected. 
![image](https://user-images.githubusercontent.com/6537525/215442275-de689ae4-8776-4a3d-a9da-61a764d68a44.png)

This needs to investigate about the structure app. For any other configuration such as multiple index fields but selecting only one model for each field does not cause this issue, i.e., the content of the matches returns are at the `c` level. 

![Image](https://user-images.githubusercontent.com/6537525/215523118-51a07940-2663-49cd-b3cd-d7d768c9f2c0.png)

In the above screenshot I made a deployment and when searching, I get some docs with both the indexed fields, however some docs returned only have text, this does not mean that the image is not there, it is just at `cc` level instead of at `c` level.

PR #995, opened 2023-01-31
fix: results with no content

azayz:
#990 

---

- [ ] Added a test

### DIFF
--- a/deployment/bff/app/v1/routers/search.py
+++ b/deployment/bff/app/v1/routers/search.py
@@ -65,7 +65,6 @@ def search(data: SearchRequestModel):
         scores = {}
         for score_name, named_score in doc.scores.items():
             scores[score_name] = named_score.to_dict()
-        # since multimodal doc is not supported, we take the first chunk
         if doc.chunks:
             # temporary fix for filtering out tags
             field_names = [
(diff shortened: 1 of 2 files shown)


deployment/bff/app/v1/routers/search.py
```diff
@@ -74,9 +73,14 @@ def search(data: SearchRequestModel):
                 if doc._metadata['multi_modal_schema'][field]['attribute_type']
                 != 'primitive'
             ]
-            field_names_and_chunks = [
-                [field_name, getattr(doc, field_name)] for field_name in field_names
-            ]
+            field_names_and_chunks = []
+            for field_name in field_names:
+                chunk = getattr(doc, field_name)
+                if chunk.chunks:
```
jupyterjazz, line 79: this should always be true, right? can you explain what was the problem?

azayz: no we have a method in elastic indexer called aggregate_embedding that changes results to chunk level instead of chunk chunk level

jupyterjazz: I don't see why we should change the structure. Let's aggregate them on cc level, this just brings confusion to me

azayz: I agree with you

joschkabraun: So back to in progress, right?

PR #997, opened 2023-01-31
refactor: create preprocessor utils

### LINKED ISSUE #998: fix: related file changes don't update preprocessor version
changing preprocessor code doesn't trigger an update because some of the code is outside preprocessor package

jupyterjazz:
Closes #998 

`maybe_download_from_s3` is only used inside preprocessor but we store it in general utils
as a result, change to the function doesn't trigger an update

---

- [ ] Added a test


now/executor/preprocessor/utils.py
```diff
@@ -0,0 +1,104 @@
+from __future__ import annotations, print_function, unicode_literals
```
florian-hoenicke, line 1: Can we rename utils to something else?

florian-hoenicke (changes requested):

PR #1002, opened 2023-01-31
fix: results with no content

### LINKED ISSUE #990: bug: content in the CC level when selecting all index fields with all models available

azayz:
closes #990 

---

- [ ] Added a test


azayz:
Screenshot 2023-02-01 at 13 18 17

jupyterjazz (approved):
LGTM!

Issue #971, opened 2023-01-24
bug: can't select animationID as filter field for customer deployment


joschkabraun:
When selecting `animationID` as filter fields, an error is thrown in the preprocessor.

![Image](https://user-images.githubusercontent.com/47435119/214260748-ece341bc-420f-44e7-91b4-d3cb2a037a14.png)

PR #1007, opened 2023-02-01
refactor: cleanup dataclass creation

### LINKED ISSUE #974: refactor: remove filter fields from dataclass creation
Filter fields shouldn't be a field in the dataclass we create but only the index fields. The filter fields should only appear as tags.

Hint:
- one relevant location in code of `create_dataclass_fields_file_mappings`: https://github.com/jina-ai/now/blob/main/now/data_loading/create_dataclass.py#L164

jupyterjazz:
Closes #974 

- removed filter fields logic from dataclass creation and refactored other parts of the project accordingly
- added tag creation in es extractor
- combined `_add_tags_to_da` and `_get_da_with_index_fields`

Also, fixes https://github.com/jina-ai/now/issues/971

screenshot of a deployment
Screenshot 2023-02-03 at 11 58 34

---

- [x] Added a test

### DIFF
--- a/now/data_loading/create_dataclass.py
+++ b/now/data_loading/create_dataclass.py
@@ -28,15 +28,13 @@ def create_dataclass(
     user_input: UserInput = None,
 ):
     """
-    Create a dataclass from the selected index and filter fields
+    Create a dataclass from the selected index fields
     and their corresponding modalities or directly from the user input which should
     contain that information. If both are provided, the user input will be used.
 
     for example:
     the index fields modalities can be:
     {'test.txt': Text , 'image.png': Image}
-    the filter fields modalities can be:
-    {'price': float, 'description': str}
 
     the dataclass will be:
 
@@ -56,13 +54,9 @@ def create_dataclass(
     """
 
     if user_input:
-        fields_modalities = {}
-        fields_modalities.update(user_input.index_field_candidates_to_modalities)
-        update_dict_with_no_overwrite(
-            fields_modalities, user_input.filter_field_candidates_to_modalities
-        )
-        fields = user_input.index_fields + user_input.filter_fields
+        fields_modalities = user_input.index_field_candidates_to_modalities
         dataset_type = user_input.dataset_type
+        fields = user_input.index_fields
 
     file_mapping_to_dataclass_fields = create_dataclass_fields_file_mappings(
         fields,
@@ -150,7 +144,6 @@ def create_dataclass_fields_file_mappings(fields: List, fields_modalities: Dict)
     modalities_count = defaultdict(int)
 
     file_mapping_to_dataclass_fields = {}
-    filter_count = 0
     for f in fields:
         if not isinstance(f, typing.Hashable):
             continue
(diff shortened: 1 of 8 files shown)


now/data_loading/data_loading.py
```diff
@@ -79,24 +65,36 @@ def add_metadata_to_da(da, user_input):
...
+    for doc in da:
+        filtered_chunks = []
+        for field in doc._metadata['multi_modal_schema'].keys():
+            field_doc = getattr(doc, field)
+            if field not in user_input.index_fields:
+                if field_doc.blob or field_doc.tensor is not None:
+                    continue
+                doc.tags.update(
+                    {
+                        field: field_doc.content
```
violenil, line 81: so we keep all the original fields that are not index fields as tags?

now/data_loading/data_loading.py
```diff
@@ -79,24 +65,36 @@ def add_metadata_to_da(da, user_input):
...
+                        if isinstance(field_doc.content, str)
```
violenil, line 82: can we have tags that are neither `str` nor `uri` but have type `float` or `int` like a `price` field for example?

violenil (approved):
Looks good! Left some questions for my own understanding.

now/data_loading/data_loading.py
jupyterjazz: yes, we don't load them though, just strings and other primitive types

now/data_loading/data_loading.py
jupyterjazz: Yes, in general we can, but this is the case where we remove a non-index chunk which won't contain int or float

now/executor/preprocessor/s3_download.py
```diff
@@ -24,13 +24,13 @@
~    :param max_workers: number of threads to create in the threadpool executor to make execution faster
~    """
~
~    flat_docs = docs['@c']
~    filtered_docs = [c for c in flat_docs if c.uri.startswith('s3://')]
```
joschkabraun: if you look at custom gateway PR, `d._metadata['s3_tags']` should already contain the right path

now/data_loading/create_dataclass.py
```diff
@@ -134,7 +126,7 @@
~        return Document(uri=value)
~
~    def my_getter(doc: 'Document'):
~        return doc.uri
~
```
joschkabraun: nice! :rocket:

now/data_loading/data_loading.py
```diff
@@ -246,24 +243,33 @@
...
+        kwargs, dict_tags = {}, {}
+        file_info = [
+            _extract_file_and_full_file_path(file, path, is_s3_dataset)
+            for file in files
+        ]
+        # first store index fields given as files
+        for file, file_full_path in file_info:
             if file in fields:
                 kwargs[field_names_to_dataclass_fields[file]] = file_full_path
-                continue
```
joschkabraun, line 257: Does that mean we put all the files which weren't indexed as tags into the docarray? Do we want that?

joschkabraun (commented):
Nice refactoring! :rocket:

now/data_loading/data_loading.py
jupyterjazz: That's correct, but only paths, not the actual content. Yeah, probably there's no need to do that.

now/executor/preprocessor/s3_download.py
jupyterjazz: You mean `d._metadata['_s3_uri_for_tags']` right? there's no `d._metadata['s3_tags']` anymore.
If your PR merges first I'll adjust, otherwise do you suggest changing anything?

now/data_loading/data_loading.py
jupyterjazz: I removed that part

Issue #294, opened 2022-07-03
installing on windows 10


gschian0:
I run install command ... this loops for a long time 
`  Downloading uvloop-0.8.1.tar.gz (1.7 MB)
     |████████████████████████████████| 1.7 MB ...
    ERROR: Command errored out with exit status 1:
     command: 'c:\python39\python.exe' -c 'import io, os, sys, setuptools, tokenize; sys.argv[0] = '"'"'C:\\Users\\GennaroSchiano\\AppData\\Local\\Temp\\pip-install-oikqiajz\\uvloop_d228fdc188f043f1b7f76f1c55d90f8e\\setup.py'"'"'; __file__='"'"'C:\\Users\\GennaroSchiano\\AppData\\Local\\Temp\\pip-install-oikqiajz\\uvloop_d228fdc188f043f1b7f76f1c55d90f8e\\setup.py'"'"';f = getattr(tokenize, '"'"'open'"'"', open)(__file__) if os.path.exists(__file__) else io.StringIO('"'"'from setuptools import setup; setup()'"'"');code = f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base 'C:\Users\GennaroSchiano\AppData\Local\Temp\pip-pip-egg-info-in0fmwpb'
         cwd: C:\Users\GennaroSchiano\AppData\Local\Temp\pip-install-oikqiajz\uvloop_d228fdc188f043f1b7f76f1c55d90f8e\
    Complete output (5 lines):
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "C:\Users\GennaroSchiano\AppData\Local\Temp\pip-install-oikqiajz\uvloop_d228fdc188f043f1b7f76f1c55d90f8e\setup.py", line 10, in <module>
        raise RuntimeError('uvloop does not support Windows at the moment')
    RuntimeError: uvloop does not support Windows at the moment
    ----------------------------------------
[...]

PR #1009, opened 2023-02-02
performance: create concurrent calls for listing s3 buckets

IyadhKhalfallah:
benchmark:

- Current solution: 80s
- Solution in PR: 38s

---

- [ ] Added a test


now/data_loading/data_loading.py
```diff
@@ -320,10 +321,30 @@ def _list_files_from_s3_bucket(
...
+        return result
+
+    prefixes = [
+        obj['Prefix']
+        for obj in bucket.meta.client.list_objects(Bucket=bucket.name, Delimiter='/')[
+            'CommonPrefixes'
+        ]
+    ]
+    objects = []
+    with ThreadPoolExecutor(max_workers=MAX_WORKERS_S3_THREADPOOL) as executor:
```
joschkabraun, line 336: We can probably use the number of cpu cores as `max_workers`.

joschkabraun:
Great! The solution is faster than previously, but it's still lacking behind the potential improvement to below one second. I think, if we improve the speed, we should go the full way.

IyadhKhalfallah:
> Great! The solution is faster than previously, but it's still lacking behind the potential improvement to below one second. I think, if we improve the speed, we should go the full way.

@joschkabraun True, we still need to solve the problem of the prompt toolkit event loop. With that, we could potentially go lower. But, I don't think we can generalize achieving the goal of going below one second, because we can't generatlize the structure of the firsr-level folders (could be also second-level, ...). In the case studied and discussed before, they could achieve very high speed because they knew the structure of the folders already, and they created concurrent calls for them in that pattern. The solution I implemented uses `bucket.meta.client.list_objects(Bucket=bucket.name, Delimiter='/')` as a general way to get the first-level folders since we have no pattern to follow.
We should also note that regarding the `async` that we might add, I'm not sure that this is a problem we want to solve since we need the data for indexing right after the loading part, so we will wait for it to finish either way. 
To conclude, the only thing that I see can help is the `Threadpool`, but it's all about defining the right prefixes. What do you think?

joschkabraun:
> > Great! The solution is faster than previously, but it's still lacking behind the potential improvement to below one second. I think, if we improve the speed, we should go the full way.
> 
> @joschkabraun True, we still need to solve the problem of the prompt toolkit event loop. With that, we could potentially go lower. But, I don't think we can generalize achieving the goal of going below one second, because we can't generatlize the structure of the firsr-level folders (could be also second-level, ...). In the case studied and discussed before, they could achieve very high speed because they knew the structure of the folders already, and they created concurrent calls for them in that pattern. The solution I implemented uses `bucket.meta.client.list_objects(Bucket=bucket.name, Delimiter='/')` as a general way to get the first-level folders since we have no pattern to follow. We should also note that regarding the `async` that we might add, I'm not sure that this is a problem we want to solve since we need the data for indexing right after the loading part, so we will wait for it to finish either way. To conclude, the only thing that I see can help is the `Threadpool`, but it's all about defining the right prefixes. What do you think?

Okay. Given the uncertainty with `asyncio`. How about we test & measure this option in isolation to see if it's worth it to make it work? Worst case scenario we call it in a separate process and in that process we write it in a DocArray to disk and then read it again later.

now/data_loading/data_loading.py
```diff
@@ -315,7 +316,53 @@ def create_docs_from_files(
 
 
 def _list_s3_file_paths(bucket, folder_prefix):
```
joschkabraun, line 318: Can you add some documentation on:
- the overall idea of this approach (in the docs of this function)
- what the single functions do (in the docs of the respective functions, e.g. `get_level_order_prefixes`)
This will make it easier for others, who haven't heard about how you are solving this issue, to understand your solution.

now/data_loading/data_loading.py
```diff
@@ -315,7 +316,53 @@ def create_docs_from_files(
...
+
+    def get_prefixes(max_levels=len(structure_identifier)):
+        level = 1
+        list_prefixes = get_level_order_prefixes(folder_prefix, level)
+        while level < max_levels - 2 and len(list_prefixes) < NUM_FOLDERS_THRESHOLD:
+            level += 1
+            list_prefixes = get_level_order_prefixes(folder_prefix, level)
+        return list_prefixes
+
+    def list_objects(prefix):
```
joschkabraun, line 348: let's remove this function, it's a one-liner

now/data_loading/data_loading.py
```diff
@@ -335,16 +382,14 @@ def _list_files_from_s3_bucket(
...
+    first_file = get_first_file_in_folder_structure_s3(bucket, folder_prefix)
     structure_identifier = first_file[len(folder_prefix) :].split('/')
     folder_structure = (
         'sub_folders' if len(structure_identifier) > 1 else 'single_folder'
     )
 
+    file_paths = _list_s3_file_paths(bucket, folder_prefix)
+
     with yaspin_extended(
         sigmap=sigmap, text="Listing files from S3 bucket ...", color="green"
```
joschkabraun, line 394: Do you mind also change the text output of this spinner to that we create the documents from S3 or so? And add a spinner above where we actually list the files.

joschkabraun (changes requested):
Really great work & experiments! Left some documentation suggestions.

now/data_loading/data_loading.py
IyadhKhalfallah: @joschkabraun this function is actually used inside the executor submit, that's why it is declared that way. I'll investigate if I can directly inject it

now/data_loading/data_loading.py
```diff
@@ -315,7 +316,79 @@ def create_docs_from_files(
...
+            prefix_list = []
+            for prefix in level_prefixes:
+                prefix_list += get_level_order_prefixes(prefix, level - 1)
+        return prefix_list
+
+    def get_prefixes(max_levels=len(structure_identifier) - 2):
+        """
+        Finds the best level for the prefixes
+
+        :param max_levels: The maximum number of level we can get to
```
joschkabraun, line 362: Do you mind also adding intuition why the default value of `max_levels` was chosen as is?

joschkabraun (approved):
LGTM! 🚀

IyadhKhalfallah:
> few questions to understand the scope/motivation of this PR:
> 
> so you're basically stepping into the next level until the number of folders is more than some predefined constant, right? how did you come up with this number?
> 
> also, why do we need to choose the level? shouldn't we expect something like this? data/ folder1 folder2 ... foldern in which case we could just pass each folder to the runners? (i.e. level1)

@jupyterjazz Actually, that number is a constant that we define based on past experiments for now. We can't tell which is the best number but, for now we made an approximation for 100. Although we plan to study this in a different ticket (#294). For the format of the data, we don't necessarily have that, and that's why we introduced this levels logic (Customer data for reference) which answers this question: Where to stop, in terms of levels in the file structure, when choosing the concurrent threads?

jupyterjazz (approved):
LGTM

PR #1024, opened 2023-02-06
fix: add slash to s3 path

joschkabraun:
---

- [ ] Added a test

### DIFF
--- a/now/common/options.py
+++ b/now/common/options.py
@@ -183,6 +183,12 @@ DATASET_PATH = DialogOptions(
     ),
 )
 
+
+def _fix_s3_uri(user_input: UserInput):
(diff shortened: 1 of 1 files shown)


jupyterjazz (commented):
so it doesn't read anything if there's no slash at the end or what's the behavior?

joschkabraun:
> so it doesn't read anything if there's no slash at the end or what's the behavior?

No, the issue would be that if someone gives `s3://bucket-name/data` it would load data from folder `s3://bucket-name/data` but also any other folder which starts with `data` e.g. `s3://bucket-name/data_private`

jupyterjazz (approved):
Nice catch!

florian-hoenicke (approved):

PR #1026, opened 2023-02-06
ci: workaround

florian-hoenicke:
workaround for this issue
https://github.com/actions/runner-images/issues/7048


aqeelat:
Just commenting to point out that changing the version won't help
https://github.com/actions/runner-images/issues/7048#issuecomment-1418823645

florian-hoenicke:
Thanks @aqeelat

PR #1027, opened 2023-02-06
fix: only install the released version of jina now

makram93:
This PR solves the following:
- Install only the released version of Jina NOW to deploy the example, because the pre-release version can sometimes be unstable.
- Set DNS value as a list
- Check for all index fields, earlier it was just checking the first index fields since we allowed only one index field
- ~~Set autoscaling to 1 based on this [slack discussion]~~(https://jinaai.slack.com/archives/C03434J5BND/p1675851878421409) 

---

- [ ] Added a test


.github/workflows/force-deploy-examples.yml
```diff
@@ -31,7 +33,7 @@ jobs:
           sudo apt update && sudo apt install ffmpeg
           python -m pip install --upgrade pip
           python -m pip install wheel
-          pip install --no-cache-dir ".[full,test]"
+          pip install jina-now -U
```
florian-hoenicke, line 36: In that case you don't need to checkout the code base

florian-hoenicke (changes requested):

now/app/search_app/app.py
```diff
@@ -80,7 +80,7 @@ def preprocessor_stub() -> Dict:
             'needs': 'autocomplete_executor',
             'uses': f'jinahub+docker://{name_to_id_map.get("NOWPreprocessor")}/{NOW_PREPROCESSOR_VERSION}',
             'jcloud': {
-                'autoscale': {'min': 0, 'max': 5, 'metric': 'concurrency', 'target': 1}
+                'autoscale': {'min': 1, 'max': 5, 'metric': 'concurrency', 'target': 1}
```
joschkabraun, line 83: why this change of `min`?

now/app/search_app/app.py
```diff
@@ -110,7 +110,7 @@ def sbert_encoder_stub() -> Tuple[Dict, int]:
                 'model_name': 'msmarco-distilbert-base-v3',
             },
             'jcloud': {
-                'autoscale': {'min': 0, 'max': 5, 'metric': 'concurrency', 'target': 1}
+                'autoscale': {'min': 1, 'max': 5, 'metric': 'concurrency', 'target': 1}
```
joschkabraun, line 113: why this change of `min`?

joschkabraun (changes requested):

now/app/search_app/app.py
makram93: During indexing, pods are terminated and restarted very frequently. There is no grace period supported yet by the wolf team autoscaling executors. As a workaround, Zac suggested using the min value as 1.
The restarting of pods causes indexing issues with large datasets
`
batch 37, try 0 HTTP Balancer service in fail-fast
trailing_metadata=Metadata((('content-length', '0'), ('l5d-proxy-error', 'HTTP Balancer service in fail-fast'), ('l5d-proxy-connection', 'close'), ('date', 'Wed, 08 Feb 2023 10:16:49 GMT')))
trailing_metadata=Metadata((('content-length', '0'), ('l5d-proxy-error', 'HTTP Balancer service in fail-fast'), ('l5d-proxy-connection', 'close'), ('date', 'Wed, 08 Feb 2023 10:16:49 GMT'))) |Gateway: Communication error with deployment preprocessor at address(es) {'preprocessor.jnamespace-e77920d288.svc:80'}. Head or worker(s) may be down.
trailing_metadata=Metadata((('content-length', '0'), ('l5d-proxy-error', 'HTTP Balancer service in fail-fast'), ('l5d-proxy-connection', 'close'), ('date', 'Wed, 08 Feb 2023 10:16:49 GMT')))
`
I am not 100% sure if this fixes it, since  I still see it sometimes.

The other reason I had in my mind, which could be unrelated, is that the pods go down to 0 and when the request comes, it takes time to bring back up the pod and that is why the first call to the flow is always slow. By keeping it to 1, I hope that one pod is always active and therefore the flow would always be ready to serve the request.

Let me know what you think.

joschkabraun: > I am not 100% sure if this fixes it, since I still see it sometimes.

So, instead of increasing `min` shouldn't we then rather remove it entirely or just leave it as is?

> The other reason I had in my mind, which could be unrelated, is that the pods go down to 0 and when the request comes, it takes time to bring back up the pod and that is why the first call to the flow is always slow. By keeping it to 1, I hope that one pod is always active and therefore the flow would always be ready to serve the request.

Starting the pods takes 5-15 seconds, which I think is fine, right?

makram93: Yes, this is fine and also acceptable. The only downside is with the cloud apps. Now we are hosting demo deployments and when any demo is selected, the API first calls `dry_run` on the flow to check if it is active. This takes up some time - 2~30s, which is a bit much for the UI since it feels stuck. But anyways, I will revert the code and change it back to 0

florian-hoenicke: I agree, 0 is best for cost saving. I see that it is tempting to have always 1 replica running but I think it is rather fixing a symptom. As far as I understand, the dry_run() is causing it. Can we also check if the flow is running without doing a dry_run? What about the client.flow_running or something like this?

makram93: `client.is_flow_ready` also calls dry_run afaik. Anyways, I have reverted the min to 0.

.github/workflows/force-deploy-examples.yml
```diff
@@ -34,10 +34,9 @@ jobs:
           python-version: 3.7
       - name: Prepare environment
         run: |
-          sudo apt update && sudo apt install ffmpeg
```
jupyterjazz, line 37: so we don't need ffmpeg anymore?

violenil (approved):
LGTM!

scripts/deploy_examples.py
```diff
@@ -100,7 +103,11 @@ def deploy(demo_ds):
     os.environ['JINA_AUTH_TOKEN'] = os.environ.get('WOLF_TOKEN')
     os.environ['NOW_EXAMPLES'] = 'True'
     os.environ['JCLOUD_LOGLEVEL'] = 'DEBUG'
-    deployment_type = os.environ.get('DEPLOYMENT_TYPE', 'partial').lower()
+    deployment_type = os.environ.get('DEPLOYMENT_TYPE', None)
+    if not deployment_type:
+        deployment_type = 'partial'
+    deployment_type = deployment_type.lower()
```
jupyterjazz, line 109: isn't this the same?

.github/workflows/force-deploy-examples.yml
makram93: We do not have a music dataset so it is not needed

jupyterjazz: then we should remove it from `ci.yml` and `cd.yml` too

jupyterjazz (approved):
LGTM!

we should remove ffmpeg from ci/cd too I think

PR #1029, opened 2023-02-06
ci: workaround mirror

florian-hoenicke:
workaround discussed here https://github.com/actions/runner-images/issues/7048#issuecomment-1419418937


florian-hoenicke:
not needed anymore

PR #1033, opened 2023-02-07
fix: s3 bucket tests

jupyterjazz:
S3 bucket data has been changed. This PR adjusts tests to the changes. 

---

- [ ] Added a test

### DIFF
--- a/tests/unit/data_loading/test_data_loading.py
+++ b/tests/unit/data_loading/test_data_loading.py
@@ -181,7 +181,7 @@ def test_from_subfolders_s3(get_aws_info):
     )
 
     loaded_da = _list_files_from_s3_bucket(user_input, data_class)
-    assert len(loaded_da) == 2
+    assert len(loaded_da) == 10
     for doc in loaded_da:
         assert doc.chunks[0].uri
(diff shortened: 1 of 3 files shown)


florian-hoenicke (approved):

tests/unit/data_loading/test_data_loading.py
```diff
@@ -181,7 +181,7 @@ def test_from_subfolders_s3(get_aws_info):
     )
 
     loaded_da = _list_files_from_s3_bucket(user_input, data_class)
-    assert len(loaded_da) == 2
+    assert len(loaded_da) == 10
     for doc in loaded_da:
         assert doc.chunks[0].uri
         assert doc.chunks[1].uri
```
Tanguyabel, line 187: Don't you want to assert all the ten uris here?

jupyterjazz: We’re asserting all 10 uris in the for loop below.
Basically, we expect two index fields for each doc - the image and the txt file, stored in two chunks.
So in the for loop we assert that both of the uris exist for each doc.

PR #1039, opened 2023-02-07
Feat: cli dynamic options

Tanguyabel:
This PR adds support for selecting dynamic options from the command line arguments in the terminal. For now it supports only the model selection options but this could easily become generic when needed.

For example:
```
 python main.py start --dataset_type path --dataset_path ~/Downloads/gif_err --index_fields .gif --model_selection .gif:Clip
```
will instruct the cli to set the Clip encoder for .gif indexes.

---

- [ ] Added a test


Tanguyabel:
In addition, this PR brings a new debug() useful command that can be use to log with file names and lines when troubleshoot ing runs, especially with remote flows.

now/dialog.py
```diff
@@ -38,6 +38,19 @@ def configure_option(
 ) -> DialogStatus:
     # Check if it is dynamic. If it is then spawn multiple dialogs for each option and return with continue
     if option.dynamic_func:
+        if option.name in kwargs:
+            # Expand dynamic options from parent option, expect a dict (supports only model_selection)
+            for user_selection in kwargs[option.name].split(","):
+                if ":" in user_selection:
+                    option_name, option_value = user_selection.split(":")
```
makram93, line 45: Can we also add some syntax separated `option_value` here because the `text` field can have clip and sbert. So, ideally I should be able to do:
```
python main.py start --dataset_type path --dataset_path ~/resources/gif --index_fields file.txt,file.gif --model_selection file.txt:Clip&Sbert,file.gif:Clip
```
`&` can be replaces with any other identifier

makram93 (commented):
I would also suggest updating the docs with an example of how the CLI can be used.

now/dialog.py
Tanguyabel: good poing, thanks! I will add support for multiple models 👍

Tanguyabel: I extended and now support multiple selection using separator '+', like in the following example:
```
python main.py start --dataset_type path --dataset_path test/multi --index_field desc.txt,img.gif --model_selection desc.txt:Clip+Sbert,img:Clip
```

now/dialog.py
```diff
@@ -38,6 +38,25 @@ def configure_option(
...
     # Check if it is dynamic. If it is then spawn multiple dialogs for each option and return with continue
     if option.dynamic_func:
+        if option.name in kwargs:
+            # Expand dynamic options from parent option, expect a dict (supports only model_selection)
+            for user_selection in kwargs[option.name].split(","):
+                if ":" in user_selection:
+                    option_name, option_values = user_selection.split(":")
+                    kwargs[f"{option_name}_model"] = []
+                    for option_value in option_values.split("+"):
+                        kwargs[f"{option_name}_model"].append(
```
makram93, line 48: We can wrap this up in a try-catch block or perform an `if` check on `options_value` and raise a proper message in case the user, provided some wrong model name.

makram93:
Also, update [this documentation](https://github.com/jina-ai/now/blob/main/docs/user-guides/cli_api.md) file with an example of how to provide model selection with cli and what values can be provided

now/dialog.py
Tanguyabel: RIght so initially I wanted to catch errors and raise something to the user. For this I wanted to leverage `parser` from `argparse`, but since we build it in other places I would have to pass it along all the functions all the way to code where we configure the dynamic options here... It was a lot of overhead. Thus I thought at first, we could silently ignore mistyped options. But I agree this is not ideal. What are your thoughts about this?

makram93: If it is possible, then we could pop the value from kwargs and raise `RetryException` (with a message that the passed fields are not correct, please select) so that this dialog will be asked to the user

Tanguyabel: I won't throw a `RetryException` here as this one is used only in the interactive context, while the parsing I added occurs when passing a selection a a command line argument. However, I will display nicer error in case of an error with the selection.

Tanguyabel: Ok I handle now two type of errors:

```
python main.py start --dataset_type path --dataset_path ~/Downloads/gif_err --index_fields .gif --model_selection foo.gif:Clip
```
```
ValueError: Error with --model_selection: foo.gif is not in an index.
```

and

```
python main.py start --dataset_type path --dataset_path ~/Downloads/gif_err --index_fields .gif --model_selection .gif:Sbert
```
```
ValueError: Error with --model_selection: Sbert is not available. for index .gif.
```

makram93: The error message looks much better now. Can we also print possible options that they could select in the same error message?
`ValueError: Error with --model_selection: Sbert is not available. for index .gif. Choose one/many among {..possible models..}`

makram93 (approved):
LGTM! 🚀

PR #1046, opened 2023-02-08
feat: use aws configure profile

IyadhKhalfallah:
Allowed AWS configure to automatically fill the credentials in both tests and CLI.

---

- [ ] Added a test

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -19,6 +19,12 @@ jobs:
         group: [1, 2, 3, 4, 5, 6, 7, 8, 9]     # set to the number of end-to-end tests
     steps:
       - uses: actions/checkout@v2
+      - name: Configure AWS CLI
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: eu-west-1
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v2
         with:
@@ -44,9 +50,7 @@ jobs:
         timeout-minutes: 35
         env:
           WOLF_TOKEN: ${{ secrets.WOLF_TOKEN }}
-          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           S3_SCHEMA_FOLDER_PATH: ${{ secrets.S3_SCHEMA_FOLDER_PATH }}
-          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           S3_CUSTOM_DATA_PATH: ${{ secrets.S3_CUSTOM_DATA_PATH }}
           S3_CUSTOM_MM_DATA_PATH: ${{ secrets.S3_CUSTOM_MM_DATA_PATH }}
 #      - name: Setup tmate session for debugging
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -19,7 +19,7 @@ jobs:
         group: [1, 2, 3, 4, 5, 6, 7, 8, 9]     # set to the number of end-to-end tests
     steps:
       - uses: actions/checkout@v2
-      - name: Configure AWS CLI
+      - name: Set AWS profile
         uses: aws-actions/configure-aws-credentials@v1
         with:
           aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
--- a/now/common/options.py
+++ b/now/common/options.py
@@ -25,7 +25,7 @@ from now.now_dataclasses import DialogOptions, UserInput
 from now.utils import (
     DemoAvailableException,
     RetryException,
-    get_credentials_from_aws_session,
+    get_aws_profile,
     get_info_hubble,
     hide_string_chars,
     jina_auth_login,
@@ -34,7 +34,7 @@ from now.utils import (
 )
 
 AVAILABLE_SOON = 'will be available in upcoming versions'
-aws_access_key_id, aws_secret_access_key, region = get_credentials_from_aws_session()
+aws_profile = get_aws_profile()
 
 # Make sure you add this dialog option to your app in order of dependency, i.e., if some dialog option depends on other
 # than the parent should be called first before the dependant can called.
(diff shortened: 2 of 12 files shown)


.github/workflows/cd.yml
```diff
@@ -19,6 +19,12 @@ jobs:
         group: [1, 2, 3, 4, 5, 6, 7, 8, 9]     # set to the number of end-to-end tests
     steps:
       - uses: actions/checkout@v2
+      - name: Configure AWS CLI
```
Tanguyabel, line 22: I would suggest here "Set AWS profile", as we do not use the cli really. What do you think?

now/common/options.py
```diff
@@ -210,7 +213,8 @@ def _fix_s3_uri(user_input: UserInput):
 
 AWS_SECRET_ACCESS_KEY = DialogOptions(
     name='aws_secret_access_key',
-    prompt_message='Please enter the AWS secret access key:',
+    prompt_message=f'Please enter the AWS secret access key: [{hide_string_chars(aws_secret_access_key)}]',
```
Tanguyabel, line 216: Let's assume we do not get AWS credentials (see comment below), this prompt would display "[None]". I don't think that is a desired behavior, instead we would like not to print any default value.
This comment is valid for the two other prompts.

now/utils.py
```diff
@@ -209,6 +210,19 @@ def modality_string_to_docarray_typing(s: str) -> TypeVar:
     return getattr(docarray.typing, s.capitalize())
 
 
+def get_credentials_from_aws_session():
+    session = boto3.Session()
+    credentials = session.get_credentials()
```
Tanguyabel, line 215: I think here we want to handle the case where the session is not set, ie `session.get_credentials()` returns None. In this case, the function could return three empty strings, ensuring compatibility in places where it is used.

now/utils.py
```diff
@@ -209,6 +210,19 @@ def modality_string_to_docarray_typing(s: str) -> TypeVar:
     return getattr(docarray.typing, s.capitalize())
 
 
+def get_credentials_from_aws_session():
```
Tanguyabel, line 213: I would maybe rename this `get_aws_profile` as we get the full profile, ie credentials + region. What do you think?

tests/conftest.py
```diff
@@ -169,10 +170,13 @@ def setup_service_running(es_connection_params) -> None:
...
     dataset_path = os.environ.get('S3_SCHEMA_FOLDER_PATH')
-    aws_access_key_id = os.environ.get('AWS_ACCESS_KEY_ID')
-    aws_secret_access_key = os.environ.get('AWS_SECRET_ACCESS_KEY')
-    region = 'eu-west-1'
-
+    (
+        aws_access_key_id,
+        aws_secret_access_key,
+        region,
+    ) = get_credentials_from_aws_session()
```
Tanguyabel, line 177: From this usage, maybe I'd recommend create a dataclass `AWSProfile` which represents the tuple (key, secret, region). It would clarify usage like here below `profile.region or 'eu-west-1'`. Let me know if you think that is relevant.

.github/workflows/cd.yml
IyadhKhalfallah: Sure!

now/common/options.py
IyadhKhalfallah: Actually, I would keep [None] because this is what I see in most of default CLIs (example for `aws configure`, if there is nothing set, it is None). What do you think?

now/utils.py
IyadhKhalfallah: They will actually return None, which we don't mind right?

tests/conftest.py
IyadhKhalfallah: Something like this?
```suggestion
    from dataclasses import dataclass

@dataclass
class AWSProfile:
    aws_access_key_id: str
    aws_secret_key: str
    region: str
    
session = boto3.Session()
credentials = session.get_credentials()
aws_profile = AWSProfile(credentials.access_key, credentials.secret_key, session.region_name)

print(aws_profile.aws_access_key_id)
print(aws_profile.aws_secret_key)
print(aws_profile.region)
```

joschkabraun (approved):
Great work!

now/common/options.py
Tanguyabel: I just checked and you are right, and this makes it clear that no value is set yet. Agree then, let's keep it like this!

tests/conftest.py
Tanguyabel: exactly yes! 🎉

now/utils.py
Tanguyabel: no, the session is None thus you can't access its attributes, that would break.

IyadhKhalfallah: You're right, I'll fix this

now/utils.py
```diff
@@ -209,6 +211,26 @@ def modality_string_to_docarray_typing(s: str) -> TypeVar:
...
+    aws_access_key_id: str
+    aws_secret_access_key: str
+    region: str
+
+
+def get_aws_profile():
+    session = boto3.Session()
+    credentials = session.get_credentials()
+    aws_profile = AWSProfile(
+        credentials.access_key, credentials.secret_key, session.region_name
```
Tanguyabel, line 225: Careful, this line will break in case no session is found, because `credentials` is None. Please handle this case.

now/utils.py
```diff
@@ -209,6 +211,29 @@ def modality_string_to_docarray_typing(s: str) -> TypeVar:
...
+
+def get_aws_profile():
+    session = boto3.Session()
+    credentials = session.get_credentials()
+    if credentials:
+        aws_profile = AWSProfile(
+            credentials.access_key, credentials.secret_key, session.region_name
+        )
+    else:
+        aws_profile = AWSProfile(None, None, None)
```
Tanguyabel, line 229:
```suggestion
    aws_profile = AWSProfile(
            credentials.access_key, credentials.secret_key, session.region_name
    ) if credentials else AWSProfile(None, None, session.region_name)
```

Tanguyabel: I think we can always use the session region, it will also be None if the session is not found (I tested).

now/utils.py
```diff
@@ -209,6 +211,28 @@ def modality_string_to_docarray_typing(s: str) -> TypeVar:
...
+
+
+def get_aws_profile():
+    session = boto3.Session()
+    credentials = session.get_credentials()
+    aws_profile = (
+        AWSProfile(credentials.access_key, credentials.secret_key, session.region_name)
+        if credentials
+        else AWSProfile(None, None, session.region_name)
+    )
```
Tanguyabel, line 228: these outer brackets here are not necessary, right?

IyadhKhalfallah: @Tanguyabel this is actually done by the linter. Do you think I should change that structure?

Tanguyabel (approved):

makram93 (approved):

florian-hoenicke (approved):
LGTM

PR #1048, opened 2023-02-08
fix: bulk index error

violenil:
Indexing documents from docarray (eg. demo datasets) with multiple fields (different modalities) AND multiple encoders is unpredictable, resulting in variable numbers of indexed documents (eg. sometimes 26 out of total 50). This is because during dataloading, documents are assigned metadata incorrectly, such that chunks/fields are 'switched' in order. This results in the indexer not being able to correctly access fields given their position/field name.

---

- [x] Added a test

### DIFF
--- a/now/data_loading/data_loading.py
+++ b/now/data_loading/data_loading.py
@@ -70,10 +70,11 @@ def _add_tags_to_da(da: DocumentArray, user_input: UserInput):
     """Add tags to da, remove non-index chunks, and update multi modal schema."""
     if not da:
         return da
-
+    all_fields = da[0]._metadata['multi_modal_schema'].keys()
     for doc in da:
         filtered_chunks = []
-        for field in doc._metadata['multi_modal_schema'].keys():
+        filtered_chunk_names = []
+        for field in all_fields:
             field_doc = get_chunk_by_field_name(doc, field)
             if field not in user_input.index_fields:
                 if field_doc.blob or field_doc.tensor is not None:
@@ -87,14 +88,15 @@ def _add_tags_to_da(da: DocumentArray, user_input: UserInput):
                 )
             else:
                 filtered_chunks.append(field_doc)
+                filtered_chunk_names.append(field)
         doc.chunks = filtered_chunks
         # keep only the index fields in metadata
         doc._metadata['multi_modal_schema'] = {
             field: doc._metadata['multi_modal_schema'][field]
-            for field in user_input.index_fields
+            for field in filtered_chunk_names
         }
         # Update the positions accordingly to access the chunks
-        for position, field in enumerate(user_input.index_fields):
+        for position, field in enumerate(filtered_chunk_names):
             doc._metadata['multi_modal_schema'][field]['position'] = int(position)
 
     return da
--- a/now/data_loading/data_loading.py
+++ b/now/data_loading/data_loading.py
@@ -32,7 +32,7 @@ def load_data(
     if user_input.dataset_type in [DatasetTypes.DOCARRAY, DatasetTypes.DEMO]:
         print_callback('⬇  Pull DocumentArray dataset')
         da = _pull_docarray(user_input.dataset_name, user_input.admin_name)
-        da = _add_tags_to_da(da, user_input)
+        da = _update_fields_and_metadata(da, user_input)
     elif user_input.dataset_type == DatasetTypes.PATH:
         print_callback('💿  Loading files from disk')
         da = _load_from_disk(user_input=user_input, data_class=data_class)
@@ -43,7 +43,7 @@ def load_data(
         print_callback('🔍  Loading data from Elasticsearch')
         da = _extract_es_data(user_input=user_input, data_class=data_class)
     da = set_modality_da(da)
-    _add_metadata_to_da(da, user_input)
+    _add_metadata_to_chunks(da, user_input)
     if da is None:
         raise ValueError(
             f'Could not load DocumentArray dataset. Please check your configuration: {user_input}.'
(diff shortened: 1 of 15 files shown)


now/data_loading/data_loading.py
```diff
@@ -70,10 +70,11 @@ def _add_tags_to_da(da: DocumentArray, user_input: UserInput):
     """Add tags to da, remove non-index chunks, and update multi modal schema."""
     if not da:
         return da
-
+    all_fields = da[0]._metadata['multi_modal_schema'].keys()
     for doc in da:
         filtered_chunks = []
-        for field in doc._metadata['multi_modal_schema'].keys():
+        filtered_chunk_names = []
```
joschkabraun, line 76: When is `filtered_chunk_names` != `user_input.index_fields`? Or is it rather that the positions don't align anymore?

now/executor/gateway/bff/app/v1/routers/info.py
```diff
@@ -26,6 +28,18 @@ async def get_tags(data: BaseRequestModel) -> TagsResponseModel:
...
+@router.post('/count')
+async def get_count(data: CountRequestModel) -> CountResponseModel:
+    response = await jina_client_post(
+        request_model=data,
+        docs=Document(),
+        endpoint='/count',
+        target_executor=r'\Aindexer\Z',
+        parameters={'limit': data.limit},
+    )
+    return CountResponseModel(number_of_docs=response[0].tags['count'])
```
makram93, line 40: Does this line throw an error if the `/count` endpoint returns an empty DocumentArray? SInce the key `count` wont be set in the empty docarray

now/executor/indexer/elastic/elastic_indexer.py
```diff
@@ -332,6 +332,32 @@ def list(self, parameters: dict = {}, **kwargs):
...
+            result = self.es.search(
+                index=self.index_name, size=limit, from_=offset, query={'match_all': {}}
+            )['hits']['hits']
+        except Exception:
+            result = None
+            self.logger.info(traceback.format_exc())
+        if result:
+            return DocumentArray([Document(text='count', tags={'count': len(result)})])
+        else:
+            return DocumentArray()
```
makram93, line 359:
```suggestion
        except Exception:
            result = []
            self.logger.info(traceback.format_exc())
     return DocumentArray([Document(text='count', tags={'count': len(result)})])

```

azayz (approved):

now/data_loading/data_loading.py
```diff
@@ -70,10 +70,11 @@ def _add_tags_to_da(da: DocumentArray, user_input: UserInput):
     """Add tags to da, remove non-index chunks, and update multi modal schema."""
```
joschkabraun, line 70: Can we adapt the name to reflect the documentation & its functionality?

now/executor/gateway/bff/app/v1/routers/info.py
```diff
@@ -17,31 +19,43 @@
~
~@router.post('/tags')
~async def get_tags(data: BaseRequestModel) -> TagsResponseModel:
~    response = await jina_client_post(
~        request_model=data,
```
joschkabraun: We also have a `/list` index in the indexer which follows the same interface as `/count`. Should we maybe expose this directly?

now/run_all_k8s.py
```diff
@@ -13,11 +13,11 @@
~
~
~def stop_now(**kwargs):
~    choices = []
~    # Add all remote Flows that exists with the namespace `nowapi`
~    alive_flows = list_all_wolf(status=FLOW_STATUS)
~    for flow_details in alive_flows:
```
joschkabraun: Haha nice catch

joschkabraun (commented):
Nice catch!

now/data_loading/data_loading.py
violenil: It may be that the positions don't align. This is a safe way to guarantee that positions align

now/data_loading/data_loading.py
violenil: I think the name stays the same, I have not changed the purpose/task of this function, just it's buggy logic

joschkabraun: The name `_add_tags_to_da` was originally not capturing its functionality but we can fix this now.

tests/integration/remote/assertions.py
```diff
@@ -173,3 +173,14 @@ def assert_search_custom_s3(host, mm_type, dataset_length, create_temp_link=Fals
     if mm_type:
         for doc in response_json:
             assert len(doc['tags']) > 0
+
+
+def assert_indexed_all_docs(host, kwargs):
```
jupyterjazz, line 178: can we also add this for local e2e tests and maybe for other remote tests too?

jupyterjazz (approved):
Great work! 🚀

now/data_loading/data_loading.py
violenil: I've renamed it to `_update_fields_and_metadata`, since we select index fields & remove unselected fields, and add metadata at document level. How does that sound?

now/executor/gateway/bff/app/v1/routers/info.py
violenil: I thought about that too, but the `/list` endpoint returns the actual documents, which is unneccessary and creates overhead. A `/count` endpoint serves to just get the number of documents indexed, also seems more like an 'info' endpoint than '/list', which would likely not be used by FE. What do you think?

now/data_loading/data_loading.py
joschkabraun: Yes that sounds good!

now/executor/gateway/bff/app/v1/routers/info.py
joschkabraun: Yes, that makes sense.

joschkabraun (approved):
Nice work!

PR #1049, opened 2023-02-08
feat: integrate es index manager

violenil:
In this PR, the setup of elasticsearch in the cluster is removed. As such, the indexer no longer directly creates and configures the elasticsearch instance. Instead, this resource is provisioned by the `es-index-manager` (see [this repo](https://github.com/jina-ai/es-index-manager) for more info). The `es-index-manager` operator looks for the label `provision-index: "yes"` on our deployment in order to provision resources for it. This label should be found under the`spec.template.metadata.labels`, as shown below in the screenshot. If this label is not present, the operator will not create indices. When the deployment is cleaned up and removed, the operator will take note and also remove the index.

Screenshot 2023-02-20 at 09 35 45

In the case of local integration tests, as well as offline flow tests, fixtures such as `setup_service_running` will ensure the setup of a local elasticsearch instance.

Screenshot of working customer deployment:
Screenshot 2023-02-20 at 09 27 32

---

- [x] Added a test

### DIFF
--- a/now/app/search_app/app.py
+++ b/now/app/search_app/app.py
@@ -152,6 +152,7 @@ class SearchApp(JinaNOWApp):
                     ],
                 ]
             )
+        provision_index = 'yes' if not testing else 'no'
 
         return {
             'name': 'indexer',
@@ -167,7 +168,7 @@ class SearchApp(JinaNOWApp):
             'jcloud': {
                 'labels': {
                     'app': 'indexer',
-                    'provision-index': 'yes',
+                    'provision-index': provision_index,
                 },
                 'resources': {
                     'memory': '8G',
--- a/now/app/search_app/app.py
+++ b/now/app/search_app/app.py
@@ -154,6 +154,10 @@ class SearchApp(JinaNOWApp):
                     ],
                 ]
             )
+        if any['ES_HOSTS', 'ES_INDEX_NAME', 'API_KEY'] not in os.environ:
+            raise ValueError(
+                'Cannot find ES_HOSTS, ES_INDEX_NAME, API_KEY in environment variables'
+            )
 
         return {
             'name': 'indexer',
@@ -163,11 +167,17 @@ class SearchApp(JinaNOWApp):
             else 'NOWElasticIndexer',
             'env': {'JINA_LOG_LEVEL': 'DEBUG'},
             'uses_with': {
+                'hosts': os.environ['ES_HOSTS'],
+                'index_name': os.environ['ES_INDEX_NAME'],
+                'api_key': os.environ['API_KEY'],
                 'document_mappings': document_mappings_list,
-                'index_name': 'now_index' if not index_name else index_name,
             },
             'no_reduce': True,
             'jcloud': {
+                'labels': {
+                    'app': 'indexer',
+                    'es-index-manager/jinaai.provision-index': True,
+                },
                 'resources': {
                     'memory': '8G',
                     'cpu': 0.5,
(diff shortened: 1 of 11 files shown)


now/app/search_app/app.py
```diff
@@ -149,11 +149,15 @@ def indexer_stub(user_input: UserInput, encoder2dim: Dict[str, int]) -> Dict:
             },
             'no_reduce': True,
             'jcloud': {
+                'labels': {
+                    'app': 'indexer',
+                    'jinaai.provision-index': True,
```
girishc13, line 154: I wrote a small program to test the regex used by jcloud.
```go
package main

import (
    "regexp"
    "fmt"
)

func main() {
    var re = regexp.MustCompile(`^[a-zA-Z0-9]([-a-zA-Z0-9_.]{0,61}[a-zA-Z0-9])?$`)

    var str = "es-index-manager/jinaai.provision-index"
    fmt.Println("matching %s: %s", str, re.MatchString(str))

    str = "jinaai.provision-index"
    fmt.Println("matching %s: %s", str, re.MatchString(str))
}
```
The output is:
```
matching %s: %s es-index-manager/jinaai.provision-index false
matching %s: %s jinaai.provision-index true
```

violenil: So we should be good if we keep this label as `'jinaai.provision-index'`?

girishc13: Yes. Ideally this label should be an annotation but custom annotations are not support by jcloud so I moved it to the labels section.

now/executor/indexer/elastic/elastic_indexer.py
```diff
@@ -88,38 +87,26 @@ def __init__(
...
-        config_path = os.path.join(os.path.dirname(__file__), 'elasticsearch.yml')
-        with open(config_path, 'r') as config_file_handler, open(
-            destination_path, 'w'
-        ) as destination_file_handler:
-            for line in config_file_handler.readlines():
-                line = line.replace("{workspace}", workspace)
-                destination_file_handler.write(line)
+    def _do_health_check(self):
+        """Checks that Elasticsearch is up and running"""
+        while True:
```
joschkabraun, line 104: Should we let it timeout at some point?

joschkabraun (approved):
LGTM! :rocket:

now/executor/indexer/elastic/elastic_indexer.py
violenil: Yes! Actually I will set a timeout when getting the env vars as well as in this health check. Good idea.

violenil: Turns out `cluster.health(wait_for_status='yellow')` has a default timeout of 30s. Yay!

now/executor/indexer/elastic/elastic_indexer.py
```diff
@@ -88,38 +85,42 @@ def __init__(
...
-                    '/usr/share/elasticsearch/config/elasticsearch.yml',
+    def _check_env_vars(self):
+        while not all(
+            var in os.environ for var in ['ES_HOSTS', 'ES_INDEX_NAME', 'ES_API_KEY']
+        ):
+            timeout_counter = 0
+            if timeout_counter < TIMEOUT:
+                timeout_counter += 5
+                self.logger.info('Environment variables not set yet. Waiting...')
+                sleep(5)
```
jupyterjazz, line 108: why do we need to wait for this? in other words, what could change while waiting? maybe we should throw an error immediately and tell the user (us) to set envs

jupyterjazz (commented):
Nice! left one comment/question

now/executor/indexer/elastic/elastic_indexer.py
violenil: The operator needs to provision the resource first, and then set these variables in our container. This may take a little while, so we wait :)

PR #1055, opened 2023-02-09
feat: script for preparing laion100m data

jupyterjazz:
Closes https://github.com/jina-ai/now.private/issues/293

Pull original `laion400m` data uploaded by Jie;
Process each subset to match the multimodal docarray structure required by our project;
Aggregate each subset into one;
Push the aggregated data into the s3 bucket;

Created an app using a small subset

Screenshot 2023-02-09 at 16 50 24

---

- [ ] Added a test

### DIFF
--- a/scripts/prepare_laion_data.py
+++ b/scripts/prepare_laion_data.py
@@ -9,7 +9,6 @@ import os
 from docarray.typing import Image
 import glob
 import boto3
-import shutil
 
 from hubble.client.endpoints import EndpointsV2
 from hubble import Client as HubbleClient
@@ -51,19 +50,32 @@ def download_da(dir_name, part_prefix):
 
     local_filename = dir_name + f'/{part_prefix}.bin'
     # NOTE the stream=True parameter below
-    with py_requests.get(url, stream=True) as r:
-        r.raise_for_status()
-        with open(local_filename, 'wb') as f:
-            for chunk in r.iter_content(chunk_size=8192):
-                # If you have chunk encoded response uncomment if
-                # and set chunk_size parameter to None.
-                # if chunk:
-                f.write(chunk)
-
-    docs = DocumentArray.load_binary(local_filename, protocol='protobuf', compress='gzip')
-    docs = process_docs(docs)
-    os.remove(local_filename)
-    docs.save_binary(local_filename)
+    try:
+        print(local_filename, 'being downloaded')
+        with py_requests.get(url, stream=True) as r:
+            r.raise_for_status()
+            with open(local_filename, 'wb') as f:
+                for chunk in r.iter_content(chunk_size=8192):
+                    # If you have chunk encoded response uncomment if
+                    # and set chunk_size parameter to None.
+                    # if chunk:
+                    f.write(chunk)
+            docs = DocumentArray.load_binary(
+                local_filename, protocol='protobuf', compress='gzip'
+            )
+        print(local_filename, 'being processed')
+        docs = process_docs(docs)
+        os.remove(local_filename)
+        docs.save_binary(local_filename)
+        print(local_filename, 'done')
+    except:
+        print(local_filename, 'downloading failed')
+        # remove the failed file if it exists
+        try:
+            os.remove(local_filename)
+            print(local_filename, 'removed')
+        except OSError:
+            pass
 
 
 def get_metadata(doc):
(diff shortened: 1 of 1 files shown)


Tanguyabel (approved):

scripts/prepare_laion_data.py
```diff
@@ -0,0 +1,121 @@
...
+def process_docs(docs):
+    """
+    Transform regular documents into multimodal format
+    """
+    root_metadata, chunk_metadata = get_metadata(docs[0])
+    processed_docs = DocumentArray()
+    for doc in docs:
+        new_doc = Document(chunks=Document(uri=doc.uri))
+        new_doc._metadata = root_metadata
+        new_doc.chunks[0]._metadata = chunk_metadata
```
joschkabraun, line 32: Out of curiosity because I never saw this before, what is in the `_metadata` of the chunks?

scripts/prepare_laion_data.py
```diff
@@ -0,0 +1,121 @@
...
+    Processing involves transforming a regular document into a multimodal one.
+    We only store uris, otherwise data is too big.
+    Not all subsets are available, that's why we check it first.
+
+    :param dir_name: name of the directory where processed subsets will be stored.
+    :param size: number of subsets (each subset has 1m examples) to be downloaded.
+    """
+    print('starting pulling the data')
+    downloaded = 0
+    for i in range(400):
```
joschkabraun, line 50: What about an async download of `(size of RAM) / (size of single docarrays)` at the same time?

joschkabraun (commented):
Small suggestion & a question. Otherwise, LGTM!

scripts/prepare_laion_data.py
```diff
@@ -0,0 +1,121 @@
...
+        if os.path.exists(os.path.join(dir_name, f'laion_{i}.bin')):
+            downloaded += 1
+            continue
+        if downloaded == size:
+            return
+        try:
+            docs = DocumentArray.pull(f'jem-fu/laion400m_part_{i}', show_progress=True)
```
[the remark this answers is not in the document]
florian-hoenicke: @jupyterjazz what do you think?

girishc13:
How and where are you planning to run this script? This is not a trivial size dataset and there will be intermittent network failures. I'm still struggling to run my own custom script that is similar to this script on jcloud and push the data to Elasticsearch using a custom executor and custom k8s deployment.

This script needs to be much more customizable to allow running a single part or batch sequences so that the parts that failed can be retried as a separate job execution. 

This script should ideally be run as multiple jobs processing a particular sequence of batches.

scripts/prepare_laion_data.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+def process_docs(docs):
+    """
+    Transform regular documents into multimodal format
+    """
+    root_metadata, chunk_metadata = get_metadata(docs[0])
+    processed_docs = DocumentArray()
+    for doc in docs:
+        new_doc = Document(chunks=Document(uri=doc.uri))
```
joschkabraun, line 49: Let's create the docarray without hacks.

- https://docs.docarray.org/fundamentals/dataclass/new-type/#create-a-new-type
- https://github.com/jina-ai/now/blob/main/now/data_loading/create_dataclass.py#L112
- https://github.com/jina-ai/now/blob/main/now/data_loading/create_dataclass.py#L146

joschkabraun (changes requested):

scripts/prepare_laion_data.py
jupyterjazz: It had a dictionary with `<DocumentMetadata.IMAGE_TYPE: 'image_type'>`
don't think it's relevant but still kept it
anyways, we're not using this approach anymore

scripts/prepare_laion_data.py
```diff
@@ -0,0 +1,209 @@
...
+def process_docs(docs):
+    """
+    Transform regular documents into multimodal format
+    """
+    from typing import TypeVar
+
+    LaionImage = TypeVar('LaionImage', bound=str)
+
+    def my_setter(value) -> Document:
+        return Document(uri=value)
```
joschkabraun, line 90: For the returned document, let's also set its modality to `image` to make sure that this is all correctly setup.

joschkabraun (approved):
Small improvement, otherwise LGTM!

PR #1057, opened 2023-02-10
fix: field name conflicting with Document attributes leads to error

IyadhKhalfallah:
---

- [ ] Added a test

### DIFF
--- a/now/utils.py
+++ b/now/utils.py
@@ -244,7 +244,7 @@ def hide_string_chars(s):
 
 
 def get_chunk_by_field_name(doc, field_name):
-    field_position = doc._metadata['multi_modal_schema'][field_name]['position']
+    field_position = int(doc._metadata['multi_modal_schema'][field_name]['position'])
     return doc.chunks[field_position]
 
 
--- a/now/utils.py
+++ b/now/utils.py
@@ -243,6 +243,11 @@ def hide_string_chars(s):
     return ''.join(['*' for _ in range(len(s) - 5)]) + s[len(s) - 4 :] if s else None
 
 
+def get_chunk_by_field_name(doc, field_name):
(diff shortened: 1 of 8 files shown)


joschkabraun (approved):
LGTM

now/utils.py
```diff
@@ -243,6 +243,20 @@ def hide_string_chars(s):
...
+    """
+    Gets a specific chunk by field name, using its position instead of getting the attribute directly.
+    This solves the getattr problem when there are conflicting attributes with Document.
+
+    :param doc: Document to get the chunk from.
+    :param field_name: Field needed to extract the position.
+
+    :return: Specific chunk by field.
+    """
+    field_position = int(doc._metadata['multi_modal_schema'][field_name]['position'])
```
makram93, line 256: Can you wrap this around try-catch and raise an error? The possible error could be because `field_name` is not in the metadata and this scenario would never happen since we are the ones calling it after schema selection is done. Still having a check would be nice

IyadhKhalfallah: Sure!

.pre-commit-config.yaml
```diff
@@ -21,10 +21,10 @@ repos:
...
-        entry: scripts/aws-check-credentials.sh
-        language: script
-        pass_filenames: false
+#  - repo: local
+#    hooks:
+#      - id: aws-crendentials-check
+#        name: aws-crendentials-check
+#        entry: scripts/aws-check-credentials.sh
+#        language: script
+#        pass_filenames: false
```
IyadhKhalfallah, line 30: I will uncomment this when fixing the bug in #1058

makram93 (approved):
LGTM!
Nice work 🚀

PR #1063, opened 2023-02-11
fix: revert s3 optimization

florian-hoenicke:
since boto3 bucket is not thread save, we get unwanted behaviour. When trying with customer data, i got only 10% of the files

### DIFF
--- a/now/data_loading/data_loading.py
+++ b/now/data_loading/data_loading.py
@@ -1,7 +1,6 @@
 import json
 import os
 from collections import defaultdict
-from concurrent.futures import ThreadPoolExecutor
 from typing import Dict, List, Type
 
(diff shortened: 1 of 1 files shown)


IyadhKhalfallah (approved):

IyadhKhalfallah:
> since boto3 bucket is not thread save, we get unwanted behaviour. When trying with customer data, i got only 10% of the files. with this pr, it is fixed. @IyadhKhalfallah, can you check how to reactivate it? I guess it would be creating a new bucket object for each thread similar to the preprocessor

I will look into this @florian-hoenicke, even though I tried to prevent the issue of this not being thread safe by making threads independent (the metadata doesn't matter). I'll try to do as you described. Thanks!

PR #1079, opened 2023-02-15
Feat compare flows

### LINKED ISSUE #1054: feat: export search results for queries and flows
UAC
- [x] cli interface for execution
- [x] comparison table (csv/json/pandas..)
  - each row is one multimodal query (image, text, video)
  - each column is one different flow and its header contains flow id, semantic scores
- [x] sharable export like pdf
- [x] a result can consist of multiple modalities (image, text, video)
- [x] ask for a config file to specify the semantic scores for flows

joschkabraun:
Closes #1054. Also enables AWS credentials passing via CLI and increases timeout of presigned URLs from S3 to 1 hour instead of 5 minutes (needed for the `disable_to_datauri` feature).

The HTML file can be quite large (e.g. with GIFs) and I wasn't able to quickly find a way to save it as PDF. That's why the tables only contain comparison of 20 queries. When one opens in a browser, one can export it as PDF which takes a snapshot. This is the follow-up ticket: https://github.com/jina-ai/now/issues/1095.

Due to this [bug](https://github.com/jina-ai/now/issues/1091), note:
- the docarray containing the names of the fields of queries needs to follow the dataclass field names in the package (so, `text_0`, `text_1`, `image_0`, ...)
- the index field in the semantic scores needs to be the name of the dataclass (so e.g. `video_0` instead of `file.gif`)

---

- [ ] Added a test


now/run_all_k8s.py
```diff
@@ -148,3 +157,130 @@ def fetch_logs_now(**kwargs):
 
     container = "gateway" if "gateway" in pod else "executor"
     cmd(f"kubectl logs {pod} -n {namespace} -c {container}", std_output=True)
+
+
+def compare_flows(**kwargs):
```
violenil, line 162: Does it make sense to break up this function `compare_flows` into more functions? There's a lot happening here.

now/run_all_k8s.py
```diff
@@ -148,3 +157,130 @@ def fetch_logs_now(**kwargs):
...
+            )
+            row[
+                f'{flow_name} - {json.dumps(semantic_scores)}'
+            ] = SearchResponseModel.responses_to_html(
+                [SearchResponseModel(**r) for r in response.json()]
+            )
+
+        rows.append(row)
+
+    # get current data and time
```
violenil, line 280:
```suggestion
    # get current date and time
```

now/run_all_k8s.py
violenil: I also understand that it doesn't really make sense to have this `compare_flows` function tested as a whole (because then we need to deploy multiple flows etc, quite time-consuming) but if we have smaller functions here we can more easily test them separately, like the loading of queries, writing of html. Just a thought!

violenil: Maybe it would also be helpful to have some example of what the semantic score json file looks like, and a query da.

joschkabraun: I shared them in the slack channel. I am not sure if GitHub is actually a place to store such files. Wdyt?

violenil: True. Maybe we can push a query docarray for testing this ourselves? Just an idea, maybe its not useful.

now/cli/parser.py
```diff
@@ -138,6 +138,39 @@ def set_logs_parser(sp):
...
+    )
+
+    parser.add_argument(
+        '--path_semantic_scores',
+        help='Path to json file mapping flow IDs to a list of semantic scores configurations',
+        type=str,
+    )
+
+    parser.add_argument(
+        '--cluster',
```
violenil, line 156:
```suggestion
        '--clusters',
```
Intuitively has to be a list of clusters always, since we are comparing? Or could we also just use one cluster + different sem scores?

now/cli/parser.py
```diff
@@ -138,6 +138,39 @@ def set_logs_parser(sp):
...
+        help='Flow names to compare',
+        type=str,
+    )
+
+    parser.add_argument(
+        '--dataset_path',
+        help='Path to the DocArray with the queries in multi-modal format',
```
violenil, line 163: Should we also allow the use of a pushed DA?

now/executor/gateway/bff/app/v1/models/shared.py
```diff
@@ -38,14 +41,57 @@ class ModalityModel(BaseModel):
...
+        set_fields = [name for name, value in values.items() if value]
         if len(set_fields) != 1:
             raise ValueError(f"Only one of {set_fields} can be set.")
         return values
 
     @property
     def content(self):
         for field_name in self.__fields_set__:
-            field_value = get_chunk_by_field_name(self, field_name)
+            field_value = getattr(self, field_name)
```
violenil, line 52: What's the reason for changing this back? I thought it is better to retrieve the subdocument/chunk by its position, because `getattr` may conflict if `field_name` is the same as an attribute of the docarray Document eg. `text`

violenil (commented):
Nice work! Great that you are able to parallelize the comparison. Just left some more questions, but LGTM!

now/run_all_k8s.py
joschkabraun: Good idea, I will do that!

now/executor/gateway/bff/app/v1/models/shared.py
joschkabraun: Because this isn't a `Document` but a `pydantic.BaseModel`, so `get_chunk_by_field_name` doesn't make sense to apply here.

now/cli/parser.py
joschkabraun: Clusters is the right formulation. To specify semantic scores one needs to use the json file.

now/cli/parser.py
joschkabraun: Yes, good idea!

violenil (approved):
This is awesome! I can't wait to try this out myself on a few flows :) Good work!

now/run_all_k8s.py
```diff
@@ -148,3 +153,115 @@ def fetch_logs_now(**kwargs):
...
+        with open(path_semantic_scores) as fp:
+            cluster_ids_2_semantic_scores = json.load(fp)
+        flow_ids = list(cluster_ids_2_semantic_scores.keys())
+        flow_ids_http_semantic_scores = [
+            (flow_id, f'https://{flow_id}-http.wolf.jina.ai', semantic_scores)
+            for flow_id in flow_ids
+            for semantic_scores in cluster_ids_2_semantic_scores[flow_id]
+        ]
+    else:
+        flow_ids = maybe_prompt_user(
```
makram93, line 180: I am not sure if I understood this correctly but we ask for flow ids again even if it is in the kwargs?

joschkabraun: Ah good catch, thank you! Let me fix that

makram93 (approved):
LGTM! 🚀

PR #1080, opened 2023-02-15
feat: increase replicas

IyadhKhalfallah:
---

- [ ] Added a test


IyadhKhalfallah:
In a discussion with @florian-hoenicke, we agreed to directly increase the max instead of having a user-based logic. That is mainly related to the fact that having 5 or 100 as a max can have the same effect on the resources, since one takes more time (costs more) and the other more resources.

azayz (approved):

Tanguyabel (approved):
lgtm

PR #1082, opened 2023-02-15
Feat alternative hubble report

florian-hoenicke:
What do you think about this way of reporting to hubble. From my understanding, it is easier to understand and to maintain.


florian-hoenicke:
image
Client deployment is still working

IyadhKhalfallah (approved):
LGTM 🚀

now/app/base/app.py
```diff
@@ -118,7 +118,11 @@ def get_gateway_stub(self, user_input, testing=False) -> Dict:
             'uses_with': {'user_input_dict': user_input.to_safe_dict()},
             'env': {'JINA_LOG_LEVEL': 'DEBUG'},
             'jcloud': {
-                'resources': {'instance': 'C5', 'capacity': 'spot'},
+                'resources': {
+                    'instance': 'C5',
+                    'capacity': 'spot',
+                    'storage': {'kind': 'ebs', 'size': '10G'},
```
joschkabraun, line 124: does the gateway need `10G`?

azayz (approved):

now/app/base/app.py
florian-hoenicke: no, thanks - I will change

PR #1083, opened 2023-02-16
refactor: remove duplicates and add /info tests

IyadhKhalfallah:
---

- [ ] Added a test


now/common/detect_schema.py
```diff
@@ -242,7 +242,6 @@ def set_field_names_from_s3_bucket(user_input: UserInput, **kwargs):
     if yes then downloads the first folder and sets its content as fields_dict in user_input
     """
     bucket, folder_prefix = get_s3_bucket_and_folder_prefix(user_input)
-    # user has to provide the folder where folder structure begins
```
Tanguyabel, line 245: what is the reason you removed this comment?

Tanguyabel (approved):
nice refactoring and new tests added, looks good to me!

tests/integration/remote/assertions.py
```diff
@@ -139,6 +139,22 @@ def assert_suggest(suggest_url, request_body):
     )
 
 
+def assert_info_endpoints(info_url, request_body):
+    info_uris = [
+        'tags',
+        'field_names_to_dataclass_fields',
+        'encoder_to_dataclass_fields_mods',
```
violenil, line 146: There is also a `count` endpoint I believe

tests/integration/remote/assertions.py
```diff
@@ -139,6 +139,22 @@ def assert_suggest(suggest_url, request_body):
...
+    ]
+    for uri in info_uris:
+        response = requests.post(
+            info_url + uri,
+            json=request_body,
+        )
+        assert (
+            response.status_code == 200
```
violenil, line 154: Good that we are adding this to the integration tests! I wonder if we can also test these endpoints more rigorously as unit tests, similar to the unit tests for the `search-app` in the bff (see `tests/unit/bff/v1/routers/test_search.py`). Could you add some unit tests for each of these endpoints/routers? This will ensure that not only the request is successful, but that the response is also coherent and as expected.

violenil (commented):
Good refactoring! I added some suggestions.

violenil (approved):
LGTM!

PR #1097, opened 2023-02-22
fix: curation

florian-hoenicke:
fixed issues:
- persistency
- expected query field name did not match bff query field name
- update did not overwrite previous curations


now/executor/indexer/elastic/elastic_indexer.py
```diff
@@ -426,25 +441,40 @@ def curate(self, parameters: dict = {}, **kwargs):
...
-                self.es.indices.refresh(index=self.index_name)
                 ids = [r['_id'] for r in resp['hits']['hits']]
-                self.query_to_curated_ids[query] += [
-                    id for id in ids if id not in self.query_to_curated_ids[query]
-                ]
+                self.query_to_curated_ids[query] = remove_duplicates(
+                    self.query_to_curated_ids.get(query, []) + ids
+                )
+
+        self.save_curated(self.query_to_curated_ids)
```
joschkabraun, line 464: I think we can replace this function call with the code itself as it isn't reused anywhere and not doing anything original + 2 lines

now/executor/indexer/elastic/es_query_building.py
```diff
@@ -94,6 +94,8 @@ def build_es_queries(
                 docs[doc.id] = doc
                 docs[doc.id].tags['embeddings'] = {}
 
+            print('# doc.id', doc.id)
+            print('# queries', queries)
```
joschkabraun, line 98: remove print

now/executor/indexer/elastic/es_query_building.py
```diff
@@ -102,10 +104,14 @@ def build_es_queries(
                     custom_bm25_query,
                     filter,
                 )
+                print(
+                    f'# create pinned queries for {doc.id}, {doc}, {query_to_curated_ids}'
+                )
```
joschkabraun, line 109: remove print

now/executor/indexer/elastic/es_query_building.py
```diff
@@ -102,10 +104,14 @@ def build_es_queries(
...
                 pinned_queries[doc.id] = get_pinned_query(
                     doc,
                     query_to_curated_ids,
                 )
+                print(f'# pinned_queries[{doc.id}]', pinned_queries[doc.id])
```
joschkabraun, line 114: remove print

now/executor/indexer/elastic/es_query_building.py
```diff
@@ -195,12 +201,18 @@ def get_default_query(
...
+    for c in doc.chunks:
+        curated_ids += query_to_curated_ids.get(c.text, [])
+    curated_ids = remove_duplicates(curated_ids)
+    if curated_ids:
+        return {'pinned': {'ids': curated_ids}}
+    else:
+        return {}
+
+
+def remove_duplicates(l):
```
joschkabraun, line 214: let's remove this function

joschkabraun: also isn't something like `list(set(l))` more conventional?

joschkabraun (changes requested):

now/executor/indexer/elastic/es_query_building.py
florian-hoenicke: The order has to be maintained. So list(set(l)) wont work.
The function is called twice and therefore, this code would be duplicated

PR #1112, opened 2023-02-24
fix: update semantic scores based on dataclass fields of query

violenil:
Previously, when defining semantic scores, one could not define query fields with arbitrary names. The API only accepted semantic scores defined with query fields `'text_0', 'text_1' ...` and `'image_0', 'image_1' ...`. However, the query field in the semantic scores should be allowed to be any string name, as long as it is **consistent** with the name defined in the `'name'` of the query. For example, one would expect that the following snippet works with query field name `'my_query_text'`:

```
import requests

HOST = "<my_host_url>"

request_body = {
    'query': [{'name':'my_query_text', 'modality': 'text', 'value': 'test'}],
    'semantic_scores': [('my_query_text', 'label', 'encoderclip', 1.0)]
 }

response = requests.post('{HOST}/api/v1/search-app/search', json=request_body)
print(response.status_code)
print(response.content)
```

This PR fixes this issue and allows arbitrary query field naming, by mapping this query field name back to the dataclass field that is created for the query document.

Successful customer deployment:
Screenshot 2023-02-24 at 15 09 49

---

- [x] Added a test

### DIFF
--- a/tests/integration/offline_flow/test_offline_flow.py
+++ b/tests/integration/offline_flow/test_offline_flow.py
@@ -6,19 +6,10 @@ from docarray.typing import Image, Text
 from tests.integration.offline_flow.flow import OfflineFlow
 
 from now.executor.gateway.bff.app.v1.models.search import SearchRequestModel
-from now.executor.gateway.bff.app.v1.routers.search import search
 from now.now_dataclasses import UserInput
 
 
-@pytest.mark.asyncio
-async def test_docarray(
-    monkeypatch, setup_service_running, random_index_name, multi_modal_data
-):
-    """
-    Test all executors and bff together without creating a flow.
-    The Clip Encoder is mocked because it is an external executor.
-    Also, the network call for the bff is monkey patched.
-    """
+def get_user_input():
     user_input = UserInput()
     user_input.index_fields = ['product_title', 'product_description', 'product_image']
     user_input.field_names_to_dataclass_fields = {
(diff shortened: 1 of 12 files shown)


joschkabraun (approved):
LGTM

tests/integration/offline_flow/test_offline_flow.py
```diff
@@ -34,6 +34,7 @@ async def test_docarray(
     search_result = await search(
         SearchRequestModel(
             query=[{'name': 'text', 'value': 'girl on motorbike', 'modality': 'text'}],
+            semantic_scores=[('text', 'product_title', 'encoderclip', 1.0)],
```
Tanguyabel, line 37: Wouldn't it make sense to create another test with semantic scores, rather than updating this one?

violenil: There is a unit test for the bff for testing semantic scores [here](https://github.com/jina-ai/now/blob/main/tests/unit/bff/v1/routers/test_search.py#L110) but it did not fail on this edge case, probably because the problem is related to the indexer not being able to handle the semantic scores.

What might help is to add the semantic scores to other integration tests that run full deployments. I can make this change.

Tanguyabel: Sounds good to me. In addition, we should ensure that `client_with_mocked_jina_client` in `bff/conftest.py` can support semantic scores as well.

Tanguyabel (approved):
looks good to me!

violenil:
After changes in bff and playground, customer deployment still working:
Screenshot 2023-03-01 at 23 59 14

makram93 (approved):

PR #1120, opened 2023-02-27
cost: autoscaling for suggestion executor and gateway

IyadhKhalfallah:
---

- [ ] Added a test


azayz (approved):

now/constants.py
```diff
@@ -5,7 +5,7 @@
 from now.utils import BetterEnum
 
 NOW_GATEWAY_VERSION = '0.0.4-fix-custom-gateway-remote-4'
-NOW_PREPROCESSOR_VERSION = '0.0.124-refactor-spot-instances-1'
+NOW_PREPROCESSOR_VERSION = '0.0.124-feat-autoscale-suggestion-0'
```
Tanguyabel, line 8: I don't think you need to rebuild the image to apply the cloud settings?

IyadhKhalfallah: @Tanguyabel Actually it failed without this (You can check the previous commit). I think I also need to update the other versions as well

IyadhKhalfallah:
> Can you check that the state of the autocomplete executor is persisted?

@joschkabraun we do now, nice catch 🚀  still some tests are failing for some reason.

joschkabraun (approved):
Nice, LGTM!

Issue #1109, opened 2023-02-24
🚨 ci: investigate idle CI runs


joschkabraun:
Find out why there are often idle CI flows and fix their removal.

Hint:
- maybe we can name the flows by the current branch & first few letters of SHA, to link flows with PR
- for most cases the runner times out before the test. Therefore the cleanup fixture can not be called. Solution: timeout the test before the runner times out

florian-hoenicke:
Example

![Image](https://user-images.githubusercontent.com/11627845/221232267-41bcb718-3564-4849-9ed6-caae7d2e05f6.png)

florian-hoenicke:
We need to let the test time out before the runner times out

joschkabraun:
Other potential issue: if wolf deployment reaches phases `Updating` instead of `Started`, then this won't be caught by our cleanup function and the flow still exists on the cluster.

Looking at output of `pytest` to extract the flow ID, can help removing the flow.

PR #1124, opened 2023-02-27
fix: remove flows using their names

jupyterjazz:
Closes https://github.com/jina-ai/now/issues/1109

**Problem**:
Sometimes flow deployment times out at this point
https://github.com/jina-ai/now/blob/3a95729a0cbf417a2df23ac36b02bca145e4fb98/tests/integration/remote/test_e2e_elastic.py#L43
resulting in no flow details stored, thus unable to properly clean it up.

**Solution**:
Generate a random name for each flow, and instead of relying on stored flow ID, get it by iterating over all flows and checking if the flow ID contains the name (because flow ID is generated by name + some code).

Random flow name is generated by branch name + random number.
Knowing branch name can be helpful because we can link each failed and not-deleted flow to specific PR.
 
Tested manually (local runs and CI) and flows are successfully deleted

---

- [x] Added a test


florian-hoenicke:
two of the flows were not cleaned up 
image

florian-hoenicke:
I created many different runs and all of the flows got cleaned up.

florian-hoenicke (approved):
Great Work!

PR #1126, opened 2023-02-27
refactor: external sbert

florian-hoenicke:
- host sbert externally
- less tests on wolf

### DIFF
--- a/now/common/flow/flow-external-sbert.yml
+++ b/now/common/flow/flow-external-sbert.yml
@@ -21,11 +21,12 @@ executors:
       JINA_LOG_LEVEL: DEBUG
     jcloud:
       expose: true
-      autoscale:
-        min: 1
-        max: 15
-        metric: concurrency
-        target: 1
+#      autoscale:
+#        min: 1
+#        max: 15
+#        metric: concurrency
+#        target: 1
+      replicas: 2
       resources:
         instance: C6
       capacity: spot
--- a/now/common/flow/flow-external-sbert.yml
+++ b/now/common/flow/flow-external-sbert.yml
@@ -1,0 +1,30 @@
+jtype: Flow
+with:
+  monitoring: true
+  env:
+    JINA_LOG_LEVEL: DEBUG
+jcloud:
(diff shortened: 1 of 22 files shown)


now/common/flow/flow-external-sbert.yml
```diff
@@ -0,0 +1,32 @@
...
+  - name: encodersbert
+    uses: jinahub+docker://TransformerSentenceEncoder/
+    uses_with:
+      access_paths: '@cc'
+      model_name: msmarco-distilbert-base-v3
+    env:
+      JINA_LOG_LEVEL: DEBUG
+    jcloud:
+      expose: true
+#      autoscale:
```
florian-hoenicke, line 24: seems like autoscaling does not work for the external executor for now

IyadhKhalfallah: @florian-hoenicke Is there a specific reason for autoscaling not working for this? I'm interested because [this bug](https://github.com/jina-ai/now.private/issues/317) is also related

florian-hoenicke: it is an external executor and wolf currently can not autoscale it

IyadhKhalfallah: Thanks 🙏 , I'll move the related task I mentioned to **Blocked**

tests/integration/remote/test_external_executors.py
```diff
@@ -0,0 +1,18 @@
+import pytest
+from docarray import Document
+from jina import Flow
+
+from now.constants import EXTERNAL_CLIP_HOST, EXTERNAL_SBERT_HOST
```
joschkabraun, line 5: let's add `EXTERNAL_SBERT_HOST` to https://jina.ai/now-status/

joschkabraun (approved):
LGTM

tests/integration/remote/test_external_executors.py
florian-hoenicke: thanks - will do

Issue #1130, opened 2023-03-01
ci: cancel previous workflow if new commit is pushed


jupyterjazz:
cancel previous workflows so that we don't waste time verifying old commits and only test the last version of the code

PR #1131, opened 2023-03-01
chore: cancel old workflows on ci

jupyterjazz:
closes https://github.com/jina-ai/now/issues/1130

Canceling previous workflows in order to avoid wasting time testing old commits and creating extra flows. 

Added an extra job which looks at the flows created by the last workflow and deletes them if there are any.
This is needed because if the workflow gets canceled, the flows won't be automatically terminated. 

---

- [ ] Added a test

### DIFF
--- a/scripts/delete_flows.py
+++ b/scripts/delete_flows.py
@@ -19,13 +19,15 @@ def delete_ci_flows():
     and look for those who have the current branch name inside their name, and delete only those.
     """
     print('Deleting the flows that were created during this CI run')
-    branch_name = os.environ.get('GITHUB_HEAD_REF', 'local_setup').lower()[:15]
-    loop = create_eventloop()
-    jflows = loop.run_until_complete(CloudFlow().list_all())['flows']
-    for flow in jflows:
-        if flow['status']['phase'] != 'Deleted' and branch_name in flow['id']:
-            CloudFlow(flow_id=flow['id']).__exit__()
-            print(f"flow {flow['id']} is successfully terminated")
+    branch_name = os.environ.get('GITHUB_HEAD_REF')
+    if branch_name:
+        branch_name = branch_name.lower()[:15]
+        loop = create_eventloop()
+        jflows = loop.run_until_complete(CloudFlow().list_all())['flows']
+        for flow in jflows:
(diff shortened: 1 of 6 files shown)


scripts/delete_flows.py
```diff
@@ -0,0 +1,37 @@
...
+    """
+    Removes all flows that were created during a CI run and were not deleted
+    for some reason (e.g. because the workflow was canceled).
+
+    Since flow names include the branch name, we just iterate over all flows
+    and look for those who have the current branch name inside their name, and delete only those.
+    """
+    print('Deleting the flows that were created during this CI run')
+    branch_name = os.environ.get('GITHUB_HEAD_REF')
+    if branch_name:
```
Tanguyabel, line 23: can you call `get_branch_name_for_flows` here instead, to ensure consistency?

Tanguyabel (commented):
I was making some research in order to understand if a job could be gracefully canceled, ie while being canceled it could still have a chance to clean up its own resources. After reading this thread https://github.com/orgs/community/discussions/26311 it seems to me that, no, the only way is to have a separate clean-up job that is set to run "always". This is exactly the way this is done here and thus it looks good to me!
I have just one small refactoring comment, once this is addressed I will approve.

scripts/delete_flows.py
jupyterjazz: Good point! that's what I was doing in the beginning, however, if I do that, I'll have to add the project as a dependency which will make the job slower (+1.5 minute), so during every new commit you'll have to wait +1.5 minute before the new workflow starts, so I decided not to do it. wdyt?

violenil (approved):
LGTM!

scripts/delete_flows.py
jupyterjazz: That seems like a proper way of handling it, but is it worth doing just for this line?
`return os.environ.get('GITHUB_HEAD_REF').lower()[:15]`

Maybe if we have more of these functions then we can go for it? wdyt?

Tanguyabel: Ok, let's keep it like this for now then.

Tanguyabel (approved):

PR #1136, opened 2023-03-01
feat: define logger using JinaLogger

Tanguyabel:
This PR brings the usage of JinaLogger to log from the app in a consistent way and adds some logs.

---

- [ ] Added a test


now/executor/abstract/auth/auth.py
```diff
@@ -30,7 +31,34 @@ def wrapper(*args, **kwargs):
                 args[0].admin_emails,
                 args[0].api_keys,
             )
-            return func(*args, **kwargs)
+
+            cls_instance = args[0]
+            cls_instance.logger.info(f"responding to endpoint")
```
joschkabraun, line 36: should we mention which endpoint? I also think there is already some logging in jina which says "received some DataRequest", isn't there?

now/executor/abstract/auth/auth.py
```diff
@@ -30,7 +31,34 @@ def wrapper(*args, **kwargs):
...
+                cls_instance.logger.debug(f"args: {kwargs['args']}")
+
+            if 'docs' in kwargs and kwargs['docs']:
+                cls_instance.logger.info(f"processing {len(kwargs['docs'])} docs")
+                if cls_instance.logger.logger.level <= LogVerbosity.DEBUG:
+                    kwargs['docs'][0].summary()
+
+            if 'docs_matrix' in kwargs and kwargs['docs_matrix']:
+                cls_instance.logger.info(
+                    f"processing {len(kwargs['docs_matrix'])} docs_matrix"
```
joschkabraun, line 47: should we also add the length of the single docarrays in the docs_matrix?

now/executor/abstract/auth/auth.py
```diff
@@ -30,7 +31,34 @@ def wrapper(*args, **kwargs):
...
+                )
+
+            if 'docs_map' in kwargs and kwargs['docs_map']:
+                cls_instance.logger.info(
+                    f"processing {len(kwargs['docs_map'])} docs_map"
```
joschkabraun, line 52: Should we additionally say (maybe for DEBUG log level) which keys have which length of docarrays?

now/app/base/app.py
```diff
@@ -11,6 +11,9 @@
 from now.executor.name_to_id_map import name_to_id_map
 from now.now_dataclasses import DialogOptions, UserInput
 
+JINA_LOG_LEVEL = os.environ.get("JINA_LOG_LEVEL", "DEBUG")
+GATEWAY_LOG_LEVEL = os.environ.get("INDEXER_LOG_LEVEL", JINA_LOG_LEVEL)
```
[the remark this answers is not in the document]
Tanguyabel: In practice, you may want to troubleshoot only one or a couple of applications and set their log level to debug, without affecting the other ones that work fine and do not require extensive logging. This is why we now optionally support one log level per application. Here there is a typo indeed, thanks for catching it!

now/executor/abstract/auth/auth.py
Tanguyabel: You are right, there is already some logging in Jina that gives such output:
```
DEBUG  indexer@ 1 got an endpoint discovery request          [03/20/23 19:34:51]
DEBUG  indexer@ 1 recv DataRequest at /tags with id:                            
       636e0418dfd4470faf62e47b53fbe7f1           
```
Therefore I will just remove this line here.

now/executor/abstract/auth/auth.py
Tanguyabel: We could also print a summary of the first documents from docs_matric and docs_map. What do you think?

joschkabraun: Yes, we can print a summary of the actual docarray as well as the first document

now/executor/abstract/auth/auth.py
Tanguyabel: I updated the logging to handle properly docs_map and docs_matric, please have a look again.

florian-hoenicke (approved):
Great work! Debugging will become fun.

joschkabraun (approved):
Nice work!

PR #1146, opened 2023-03-02
fix: naming cd flows

jupyterjazz:
---

- [ ] Added a test

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -86,8 +86,7 @@ def get_branch_name_for_flows():
     # !IMPORTANT! if you modify this function, make sure `delete_flows.py` is adjusted.
     if 'GITHUB_HEAD_REF' in os.environ:
         return os.environ['GITHUB_HEAD_REF'].lower()[:15] or 'cd_flow'
-    else:
-        return 'local_setup'
+    return 'local_setup'
 
 
 def get_flow_id_from_name(flow_name):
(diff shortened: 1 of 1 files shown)


tests/integration/conftest.py
```diff
@@ -84,7 +84,9 @@ def get_branch_name_for_flows():
     In case of a local run, returns 'local_setup'.
     """
     # !IMPORTANT! if you modify this function, make sure `delete_flows.py` is adjusted.
-    return os.environ.get('GITHUB_HEAD_REF', 'local_setup').lower()[:15]
+    if 'GITHUB_HEAD_REF' in os.environ:
+        return os.environ['GITHUB_HEAD_REF'].lower()[:15] or 'cd_flow'
```
violenil, line 88: Are flow names allowed to have underscores?

jupyterjazz: oh, that is a great point
idk but let's go with a safer option

jupyterjazz: They should be though because I ran e2e tests locally with 'local_setup' name
but I still changed

violenil: Ah I think wolf does some preprocessing of the underscores and turns them into hyphens, so it would cause trouble for our script trying to find exactly the flow with name containing an underscore.

violenil: Also, how does `delete_flows.py` need to change to accommodate these changes?

violenil (approved):
LGTM!

PR #1169, opened 2023-03-06
refactor: create utils module

IyadhKhalfallah:
---

- [ ] Added a test


now/utils/implementation/helpers.py
```diff
@@ -0,0 +1,77 @@
+import signal
```
joschkabraun, line 1: what does `implementation` refer to?

now/utils/flow/helpers.py
```diff
@@ -0,0 +1,22 @@
+import yaml
```
joschkabraun, line 1: Let's call it `jcloud` instead of `flow` as these utils are needed for `jcloud`

IyadhKhalfallah: Sure!

now/utils/common/helpers.py
```diff
@@ -0,0 +1,77 @@
+import signal
+import sys
+from collections.abc import MutableMapping
+from inspect import stack
+from typing import Any
+
+from pyfiglet import Figlet
+
+
+def debug(msg: Any):
```
violenil, line 10: Is this function even used anywhere?

now/utils/common/helpers.py
```diff
@@ -0,0 +1,77 @@
...
+
+class BetterEnum:
+    def __iter__(self):
+        return [getattr(self, x) for x in dir(self) if ('__' not in x)].__iter__()
+
+
+sigmap = {signal.SIGINT: my_handler, signal.SIGTERM: my_handler}
+
+
+def print_headline():
```
violenil, line 62: Can this function not be moved to the dialog.py? It is not used anywhere else in the codebase.

violenil (changes requested):
Looks good! Nice refactoring.

joschkabraun (approved):
LGTM, nice refactoring!

violenil (approved):
🚀

PR #1174, opened 2023-03-07
fix: client retry, grpc server options, autoscale args

makram93:
### Client side retry

With different serverless Executors used across the Flow for cost control, network errors are bound to happen, and the best workaround, for now, is `retry` during failure. This PR introduces retrying of requests during `client.post(...)`. 

- Currently, we use `continue_on_error=True` in `client.post()` which doesn't work during network errors as the documentation mentions. 
  image
- `client.post()` is vulnerable to different kinds of  Exceptions raised, they need to be caught and managed.

Retry logic implemented in this PR depends on `doc.id` remaining the same from the Client to the last Executor (indexer) as jina doesn't provide access to `request.id` on the Client side. _This logic is tested locally with a dummy Flow that doesn't change the doc.id, but that doesn't seem to be the case in `indexer`_ (where we return an empty `DocumentArray`). 

**To make the retry work, I've changed from returning an empty DA to a DA only with the ids of the original `docs`.**

https://github.com/jina-ai/now/blob/fcf36c23a7d2cdcc0f990bc106681be4ce3e5908/now/executor/indexer/elastic/elastic_indexer.py#L195

```python
return DocumentArray([Document(id=i) for i in docs[:, 'id']]
```

image

> We should adapt to jina's native retry mechanism, once [this](https://github.com/jina-ai/jina/issues/5730) is fixed.

---

### `timeout_ready` 

Add a higher (`180s`) for gateway

---

### gRPC server options

While debugging the request failures, I noticed intermittent `HTTP Logical service in fail-fast` errors in linkerd. One of the [comments](https://github.com/linkerd/linkerd2/issues/5599#issuecomment-772426026) suggested setting a higher value of `max_concurrent_streams`. jina allows passing `grpc_server_options` to all Executors & Gateway. This PR adds that option.

---

### `autoscale.target` arg

Right now, we set `autoscale.target: 1` for all autoscaled Executors along with `autoscale.metric: concurrency`. This value actually denotes the [number of simultaneous requests that can be processed by each replica](https://knative.dev/docs/serving/autoscaling/concurrency/). Setting this to 1 would be a bottleneck to concurrent requests. This PR sets `autoscale.target: 100` for all Executors in the Flow.

---

- [ ] Added a test


now/app/search_app/app.py
```diff
@@ -77,16 +78,17 @@ def autocomplete_stub(testing=False) -> Dict:
             'jcloud': {
                 'autoscale': {
                     'min': 0,
-                    'max': 1,
+                    'max': 100,
                     'metric': 'concurrency',
-                    'target': 1,
+                    'target': 100,
```
joschkabraun, line 83: Do you observe a scaling up with the updated `target=100`?

now/deployment/deployment.py
```diff
@@ -36,10 +36,15 @@ def list_all_wolf(status='Serving', namespace='nowapi'):
...
-        flows.append(
-            {'id': flow['id'], 'name': flow['status']['endpoints'][executor_name]}
-        )
+        try:
+            executor_name = list(flow['status']['endpoints'].keys())[0]
+            flows.append(
+                {'id': flow['id'], 'name': flow['status']['endpoints'][executor_name]}
+            )
+        except Exception as ex:
+            print(f'Failed to parse flow {flow["id"]}: {ex}')
```
joschkabraun, line 45: I believe `traceback.print_exec` is a better way to print exceptions

now/run_backend.py
```diff
@@ -108,52 +109,133 @@ def index_docs(user_input, dataset, client, print_callback, **kwargs):
...
+        ) % 5 == 0:  # print after every 5 inserts
+            print_if_example(f'Indexed {len(completed_req_ids)} docs.')
+
+    def _on_error(r: Request):
+        print_if_example(
+            'Got an executor error while indexing. '
+            'Should trigger a retry using Jina\'s inbuilt mechanism.'
+        )
+
+    def stream_requests_until_done(docs: DocumentArray):
```
joschkabraun, line 202: let's remove this function as it's only called once + shallow

joschkabraun (commented):
Small suggestions only

now/app/search_app/app.py
makram93: still investigating this

PR #1175, opened 2023-03-08
feat: convert multiple html to pdf

azayz:
---

- [ ] Added a test

### DIFF
--- a/now/compare/compare_flows.py
+++ b/now/compare/compare_flows.py
@@ -1,3 +1,4 @@
+import glob
 import json
 import os
 from concurrent.futures import ProcessPoolExecutor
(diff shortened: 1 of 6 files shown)


now/compare/compare_flows.py
```diff
@@ -150,3 +153,10 @@ def _evaluate_query(
         )
 
     return row
+
+
+def _convert_multiple_html_to_pdf(path: str):
+    """Combines multiple html files into one pdf file."""
+    html_files = glob.glob(f'{path}/*.html')
+    html_files.sort()
```
IyadhKhalfallah, line 161: Could you describe more the pattern of the html file names? Is the sorting always accurate in this case?

PR #1179, opened 2023-03-09
test: add preprocessor tests

azayz:
---

- [ ] Added a test

### DIFF
--- a/now/executor/preprocessor/executor.py
+++ b/now/executor/preprocessor/executor.py
@@ -14,6 +14,14 @@ from now.executor.preprocessor.s3_download import maybe_download_from_s3
 Executor = get_auth_executor_class()
 
 
+def move_uri(d: Document) -> Document:
+    cloud_uri = d.tags.get('uri')
+    if isinstance(cloud_uri, str) and cloud_uri.startswith('s3://'):
+        d.uri = cloud_uri
+        d.chunks[:, 'uri'] = cloud_uri
+    return d
+
+
 class NOWPreprocessor(Executor):
     """Applies preprocessing to documents for encoding, indexing and searching as defined by app.
     If necessary, downloads files for that from cloud bucket.
@@ -67,13 +75,6 @@ class NOWPreprocessor(Executor):
                 self.user_input
                 and self.user_input.dataset_type == DatasetTypes.S3_BUCKET
             ):
-
-                def move_uri(d: Document) -> Document:
-                    cloud_uri = d.tags.get('uri')
-                    if isinstance(cloud_uri, str) and cloud_uri.startswith('s3://'):
-                        d.uri = cloud_uri
-                        d.chunks[:, 'uri'] = cloud_uri
-                    return d
 
                 for d in docs:
                     for c in d.chunks:
--- a/now/executor/preprocessor/executor.py
+++ b/now/executor/preprocessor/executor.py
@@ -1,9 +1,4 @@
-import io
-import os
-import random
-import string
 import tempfile
-import urllib
 
 from jina import Document, DocumentArray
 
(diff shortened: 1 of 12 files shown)


tests/executor/preprocessor/test_unit.py
```diff
@@ -30,3 +41,101 @@ def test_text(mm_dataclass):
...
+    da = load_data(user_input)
+    maybe_download_from_s3(da, tmpdir, user_input, 2)
+
+    assert len(da) == 10
+    for doc in da:
+        assert doc.chunks[0].uri.startswith(str(tmpdir))
+        assert doc.chunks[0].uri.endswith('.png')
+        assert doc.chunks[1].uri.startswith(str(tmpdir))
+        assert doc.chunks[1].uri.endswith('.txt')
+        assert doc.tags
```
joschkabraun, line 140: let's assert the `tags` being not empty

tests/executor/preprocessor/test_unit.py
```diff
@@ -30,3 +41,101 @@ def test_text(mm_dataclass):
...
+        isinstance(path, str)
+        and path.startswith(str(tmpdir))
+        and path.endswith('.json')
+    )
+
+    path = get_local_path(tmpdir, 'test')
+    assert isinstance(path, str) and path.startswith(str(tmpdir)) and path.endswith('.')
+
+
+def test_move_uri():
```
joschkabraun, line 95: let's separate the testing for the "positive" and the "negative" in two functions

tests/executor/preprocessor/test_unit.py
```diff
@@ -30,3 +41,101 @@ def test_text(mm_dataclass):
...
+        aws_secret_access_key=os.environ['AWS_SECRET_ACCESS_KEY'],
+        region_name='eu-west-1',
+    )
+
+    assert len(d.tags) > 0
+    for value in d.tags.values():
+        assert not isinstance(value, dict)
+
+
+def test_convert_fn(tmpdir):
```
joschkabraun, line 61: I think that's a point to finally change the nme of this function

joschkabraun (changes requested):
Nice tests you added!
Can you also add one test in which we instaniate a preprocessor and test the `preprocess` function across data sources (docarray, s3, maybe elastic)? Then we cover the main functionality of the preprocesor and we have 100% test coverage

tests/executor/preprocessor/test_unit.py
```diff
@@ -30,3 +41,123 @@ def test_text(mm_dataclass):
...
+        isinstance(path, str)
+        and path.startswith(str(tmpdir))
+        and path.endswith('.json')
+    )
+
+    path = get_local_path(tmpdir, 'test')
+    assert isinstance(path, str) and path.startswith(str(tmpdir)) and path.endswith('.')
+
+
+def test_move_uri():
```
joschkabraun, line 95: maybe call it `test_move_uri_not_s3`

tests/executor/preprocessor/test_unit.py
```diff
@@ -30,3 +41,123 @@ def test_text(mm_dataclass):
...
+)
+def test_all_cases(data, request):
+
+    docs, user_input = request.getfixturevalue(data)
+
+    preprocessor = NOWPreprocessor()
+    result = preprocessor.preprocess(docs)
+
+    assert result
+    assert len(result[0].chunks) > 0 and len(result[0].chunks[0].chunks) >= 0
```
joschkabraun, line 163: let's check the entire dataset + the last check should be strict, shouldn't it be (otherwise something went wrong in the preprocessing)?

joschkabraun (changes requested):

joschkabraun:
Looking at test coverage, do know why `maybe_download_from_s3` isn't hit in the preprocessor?

now/executor/preprocessor/executor.py
```diff
@@ -19,6 +14,14 @@
 Executor = get_auth_executor_class()
 
 
+def move_uri(d: Document) -> Document:
```
violenil, line 17: Why does this function need to be moved outside of the executor?

now/now_dataclasses.py
```diff
@@ -22,7 +22,6 @@ class UserInput(BaseModel):
     flow_name: Optional[str] = None
     dataset_type: Optional[DatasetTypes] = None
     dataset_name: Optional[str] = None
-    dataset_url: Optional[str] = None
```
violenil, line 25: Do we not need this for S3?

violenil (approved):
Looks good! I just have some questions for my own understanding.

now/now_dataclasses.py
azayz: we don't need it at all that's why I deleted it

now/executor/preprocessor/executor.py
azayz: to be able to make the import previously it was inside the method

jupyterjazz (commented):
Great PR!

can you also modify `test_text` in the unit tests to work for several index fields (including image fields)? currently it only considers one text field

jupyterjazz:
Also, `download_mock` isn't used anymore so we can remove it

jupyterjazz:
It is but not the case when docs have more than one index field. 
Also, I think we should move the file/tests to the preprocessor directory in the unit tests

jupyterjazz (approved):
LGTM!

joschkabraun (approved):
LGTM

PR #1180, opened 2023-03-09
feat: allow multiple filters

violenil:
Enabling multiple filters sent in the search request. Slight change to the filter parameter, also updated the documentation for NOW and API documentation to reflect new format. 

Note that numerical and categorical filters now have different input fields in the playground: if a filter has numerical information (eg. price, size -> `float` or `int`), then we use a range slider with min + max values. If the filter has categorical information (eg. color, category -> `str`), then we use a multiselect field where multiple values can be selected.

Screenshot 2023-03-10 at 15 18 42

As for the API request body with a filter, this is what it should look like:

```
{
  "filters": {
    "color": ["blue", "red"],
    "price": {"lte": 100, "gte": 50}
  },
  "query": [
    ...
  ],
  "create_temp_link": false,
  "score_calculation": [
   ...
  ]
}
```

For more info on which operators are supported by Elasticsearch in range queries (ie. for numerical filters), see [this documentation](https://www.elastic.co/guide/en/elasticsearch/reference/master/query-dsl-range-query.html).

Successful customer deployment using filters for query:

Screenshot 2023-03-10 at 15 15 13

---

- [x] Added a test
- [x] Updated the documentation

### DIFF
--- a/now/executor/gateway/bff/app/v1/models/search.py
+++ b/now/executor/gateway/bff/app/v1/models/search.py
@@ -29,7 +29,7 @@ class SearchRequestModel(BaseRequestModel):
     filters: Optional[Dict[str, Union[List, Dict[str, int]]]] = Field(
         default={},
         description='dictionary with filters for search results',
-        example={'tags__color': ['blue'], 'tags__price': {'lt': 50.0}},
+        example={'color': ['blue'], 'price': {'lt': 50.0}},
     )
     query: List[Dict] = Field(
         default={},
@@ -85,7 +85,7 @@ class SearchResponseModel(BaseModel):
         ]
     ] = Field(
         description='Additional tags associated with the file.',
-        example={'tags__price': {'$lt': 50.0}},
+        example={'price': {'$lt': 50.0}},
     )
     fields: Dict[str, ModalityModel] = Field(
         default={},
--- a/now/executor/gateway/bff/app/v1/models/search.py
+++ b/now/executor/gateway/bff/app/v1/models/search.py
@@ -26,10 +26,10 @@ class SearchRequestModel(BaseRequestModel):
     limit: int = Field(
         default=10, description='Number of matching results to return', example=10
     )
-    filters: Optional[Dict[str, str]] = Field(
+    filters: Optional[Dict[str, Union[List, Dict[str, int]]]] = Field(
         default={},
         description='dictionary with filters for search results',
-        example={'tags__color': {'$eq': 'blue'}},
+        example={'tags__color': ['blue'], 'tags__price': {'lt': 50.0}},
     )
     query: List[Dict] = Field(
         default={},
--- a/now/executor/gateway/bff/app/v1/models/search.py
+++ b/now/executor/gateway/bff/app/v1/models/search.py
@@ -26,7 +26,7 @@ class SearchRequestModel(BaseRequestModel):
     limit: int = Field(
         default=10, description='Number of matching results to return', example=10
     )
-    filters: Optional[Dict[str, Union[List, Dict[str, int]]]] = Field(
+    filters: Optional[Dict[str, Union[List, Dict[str, Union[int, float]]]]] = Field(
         default={},
         description='dictionary with filters for search results',
         example={'color': ['blue'], 'price': {'lt': 50.0}},
@@ -85,7 +85,7 @@ class SearchResponseModel(BaseModel):
         ]
     ] = Field(
         description='Additional tags associated with the file.',
-        example={'price': {'$lt': 50.0}},
+        example={'price': {'lt': 50.0}},
     )
     fields: Dict[str, ModalityModel] = Field(
         default={},
--- a/now/executor/gateway/bff/app/v1/routers/search.py
+++ b/now/executor/gateway/bff/app/v1/routers/search.py
@@ -62,8 +62,8 @@ search_examples = {
         'value': {
             'limit': 10,
             'filters': {
-                'tags__color': ['blue', 'red'],
-                'tags__price': {'lte': 100, 'gte': 50},
+                'color': ['blue', 'red'],
+                'price': {'lte': 100, 'gte': 50},
             },
             'query': [
                 {
--- a/now/executor/gateway/bff/app/v1/routers/search.py
+++ b/now/executor/gateway/bff/app/v1/routers/search.py
@@ -132,7 +132,7 @@ async def search(
 
     query_filter = {}
     for key, value in data.filters.items():
-        key = 'tags__' + key if not key.startswith('tags__') else key
+        key = 'tags__' + key
         query_filter[key] = value
 
     docs = await jina_client_post(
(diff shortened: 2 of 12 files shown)


now/executor/gateway/bff/app/v1/models/search.py
```diff
@@ -26,10 +26,10 @@ class SearchRequestModel(BaseRequestModel):
     limit: int = Field(
         default=10, description='Number of matching results to return', example=10
     )
-    filters: Optional[Dict[str, str]] = Field(
+    filters: Optional[Dict[str, Union[List, Dict[str, int]]]] = Field(
         default={},
         description='dictionary with filters for search results',
-        example={'tags__color': {'$eq': 'blue'}},
+        example={'tags__color': ['blue'], 'tags__price': {'lt': 50.0}},
```
joschkabraun, line 32: maybe we should change the examples to not include leading `tags__` due to the change in this PR

now/executor/gateway/bff/app/v1/routers/search.py
```diff
@@ -62,8 +62,8 @@
         'value': {
             'limit': 10,
             'filters': {
-                'tags__color': {'$eq': 'blue'},
-                'tags__price': {'$lte': 100, '$gte': 50},
+                'tags__color': ['blue', 'red'],
+                'tags__price': {'lte': 100, 'gte': 50},
```
joschkabraun, line 66: maybe we should change the examples to not include leading `tags__` due to the change in this PR

now/executor/gateway/playground/playground.py
```diff
@@ -136,47 +121,79 @@ def deploy_streamlit(user_input: UserInput):
...
+            elif isinstance(values[0], int) or isinstance(values[0], float):
+                processed_filters[field] = {'gte': values[0], 'lte': values[1]}
+        elif isinstance(values, int) or isinstance(values, float):
+            processed_filters[field] = {'gte': values}
+        else:
+            raise ValueError(f'Filter values {values} are not supported.')
+    st.session_state.filter_selection = processed_filters
+
+
+def apply_filters(params):
```
joschkabraun, line 170: except for `process_filters` this function is more something aligned with creating filter interface or do I misunderstand? For context: `apply_filters` evoked in me that filters will be applied for e.g. a search request

tests/executor/indexer/test_es_query_building.py
```diff
@@ -69,3 +72,29 @@ def test_build_es_queries(es_inputs):
...
+def test_filter_processing(filters, error):
+    """
+    This test tests the process_filter function from es_query_building, including an errorful filter case.
+    """
+    if error:
+        with pytest.raises(error):
+            process_filter(filters)
+    else:
+        processed_filters = process_filter(filters)
+        print(processed_filters)
```
joschkabraun, line 99: let's remove print

joschkabraun (changes requested):
Nice feature! Some small comments

now/executor/gateway/playground/playground.py
violenil: So true. I was struggling to think of a correct function name. Perhaps `create_filter_interface()`?

violenil: Or `render_filters`?

joschkabraun: Yes, I guess `render_filters` is the most intuitive

now/executor/gateway/bff/app/v1/models/search.py
```diff
@@ -26,10 +26,10 @@ class SearchRequestModel(BaseRequestModel):
     limit: int = Field(
         default=10, description='Number of matching results to return', example=10
     )
-    filters: Optional[Dict[str, str]] = Field(
+    filters: Optional[Dict[str, Union[List, Dict[str, int]]]] = Field(
```
makram93, line 29:
```suggestion
    filters: Optional[Dict[str, Union[List, Dict[str, float]]]] = Field(
```
since in the example provided value is float type?

now/executor/gateway/bff/app/v1/models/search.py
```diff
@@ -85,7 +85,7 @@ class SearchResponseModel(BaseModel):
         ]
     ] = Field(
         description='Additional tags associated with the file.',
-        example={'tags__price': {'$lt': 50.0}},
+        example={'price': {'$lt': 50.0}},
```
makram93, line 88: Is `$` needed here? Can we have the same format for `tags` and `filters` so that FE does not have confusion while integrating?

tests/executor/indexer/test_elastic_in_flow.py
```diff
@@ -163,7 +163,7 @@ def test_search_with_filtering(self, metas, setup_service_running, flow):
             on='/search',
             inputs=docs_query,
             return_results=True,
-            parameters={'filter': {'tags__price': {'$lt': 50.0}}},
+            parameters={'filter': {'tags__price': {'lt': 50.0}}},
```
makram93, line 166:
```suggestion
            parameters={'filter': {'price': {'lt': 50.0}}},
```
test cases should replicate the values we expect

makram93 (commented):
Nice work! Already in a good shape! 🚀 
Just some changes suggested to maintain uniformity

now/executor/gateway/bff/app/v1/models/search.py
violenil: It could be float or int actually! I can fix that.

now/executor/gateway/bff/app/v1/models/search.py
violenil: It is actually not allowed, the $ will cause an error. Will fix.

tests/executor/indexer/test_elastic_in_flow.py
violenil: Actually, we do expect this format for the filter parameters arriving at the indexer. The BFF will modify an incoming filter that looks like this: `'filter': {'price': {'lt': 50.0}}` to -> `'filter': {'tags__price': {'lt': 50.0}}`

violenil: The `tags__` prefix helps the es query building to access the correct filter fields found in the tags.

azayz (commented):
Nice changes, I would investigate why patch test coverage is dropping and add tests to that, I would also add some docstring to functions to make sure they are easy to understand

violenil:
> Nice changes, I would investigate why patch test coverage is dropping and add tests to that, I would also add some docstring to functions to make sure they are easy to understand

@azayz the codecov patch is failing because I am making a few changes to the playground, but we dont test the playground at all (with streamlit it is difficult to test). I added tests for the changes I made, somehow the codecov calculation is off: if you look at the test coverage for query building and for the search model in bff, these are both increased by `2.94%` and `0.04%` respectively

joschkabraun:
Just so we don't forget about it:

I believe that any key-value pairs in not DocArray data sources will be treated as strings such that the range filter won't be applicable.

makram93 (approved):
LGTM!

PR #1192, opened 2023-03-14
fix: merge two level sub-app

makram93:
Merge the two sub level app into one as both of these are related and are needed

---

- [ ] Added a test


IyadhKhalfallah (approved):
LGTM after small comments 🚀

now/executor/gateway/bff/app/v1/routers/search.py
```diff
@@ -169,7 +149,8 @@ async def search(
                 for field_name in doc._metadata['multi_modal_schema'].keys()
             ]
         else:
-            # TODO remove else path. It is only used to support the inmemory indexer since that one is operating on chunks while elastic responds with root documents
+            # TODO remove else path. It is only used to support the inmemory
+            #  indexer since that one is operating on chunks while elastic responds with root documents
```
joschkabraun, line 153: we can delete this `else` path as we don't have hat anymore right?

now/executor/indexer/elastic/elastic_indexer.py
```diff
@@ -489,8 +489,8 @@
             updated_tags = {}
             for tag, agg in aggregations.items():
                 updated_tags[tag] = [bucket['key'] for bucket in agg['buckets']]
-            self.doc_id_tags = updated_tags
-        except Exception:
+            self.doc_filter_tags = updated_tags
```
joschkabraun, line 492: if we rename this field, I believe something which reflects the nature that the keys are the filter field names and that the values are the values is best

joschkabraun (commented):
Nice refactoring & cleanup!
Some additional small suggestions.

PR #1198, opened 2023-03-15
fix: add m2m token in env of gateway exec

makram93:
m2m token should be added in the env of the gateway exec when creating flow.yml

---

- [ ] Added a test

### DIFF
--- a/now/executor/gateway/hubble_report.py
+++ b/now/executor/gateway/hubble_report.py
@@ -49,6 +49,7 @@ def report_search_usage(user_token):
 def report(quantity_basic, quantity_pro, authorized_jwt=None, user_token=None):
     if not authorized_jwt and not user_token:
         raise Exception('Either authorized_jwt or user_token must be provided')
+    logger.info(f'Charging {authorized_jwt or user_token} at {current_time()}')
     app_id = 'search'
     product_id = 'mm_query'
     try:
@@ -57,7 +58,9 @@ def report(quantity_basic, quantity_pro, authorized_jwt=None, user_token=None):
             logger.info('M2M_TOKEN not set in the environment')
         payment_client = PaymentClient(m2m_token=m2m_token)
         if not authorized_jwt:
-            authorized_jwt = payment_client.get_authorized_jwt(user_token)['data']
+            authorized_jwt = payment_client.get_authorized_jwt(user_token=user_token)[
+                'data'
+            ]
         summary = get_summary(authorized_jwt, payment_client)
         logger.info(f'Payment summary: \n{summary}')
         if summary['internal_product_id'] == 'free-plan':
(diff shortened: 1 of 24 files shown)


now/now_dataclasses.py
```diff
@@ -46,6 +46,7 @@ class UserInput(BaseModel):
     cluster: Optional[str] = None
     secured: Optional[StrictBool] = False
     jwt: Optional[Dict[str, str]] = None
+    authorized_jwt: Optional[str] = None
```
joschkabraun, line 49: If I correctly understand we fetch `authorized_jwt` in the gateway from Hubble and only really use it there. Should we either fetch it when creating the flow or should we maybe not write it in the flow into `UserInput` and leave it as a global variable in the gateway?

makram93: Ideally, authorized_jwt should be used for `base_fee_thread` func to avoid the usage of user_token since it may expire. However, it is not needed in the user_input.

now/executor/gateway/hubble_report.py
```diff
@@ -51,64 +49,67 @@ def report_search_usage(user_token):
 
 
 def report(user_token, quantity_basic, quantity_pro):
+    logger.info('*******************************')
```
Tanguyabel, line 52: is this log line really necessary?

now/executor/gateway/hubble_report.py
```diff
@@ -51,64 +49,67 @@ def report_search_usage(user_token):
...
-                },
+            logger.info(
+                f'**** `{round(quantity, 3)}` credits charged at time: {current_time()} ****'
             )
+            summary = get_summary()
+            logger.info(f'Credits after: {summary["credits"]}')
+        else:
+            logger.info(f'**** Could not charge. Check payment summary ****')
+        logger.info(f'Payment summary: {summary}')
+        logger.info('*******************************')
```
Tanguyabel, line 88: same comment as above

now/executor/gateway/now_gateway.py
```diff
@@ -127,11 +131,25 @@ def __init__(self, user_input_dict: Dict = {}, **kwargs):
...
+    def _check_env_vars(self):
+        while 'M2M' not in os.environ:
+            timeout_counter = 0
+            if timeout_counter < TIMEOUT:
+                timeout_counter += 5
+                self.logger.info('Environment variables not set yet. Waiting...')
+                sleep(5)
+            else:
+                self.logger.error(
+                    'Elasticsearch environment variables not set after 60 seconds. Exiting...'
```
Tanguyabel, line 149: I think "Elasticsearch" is a typo here?

makram93: ah yes 👍

joschkabraun (approved):
LGTM

PR #1200, opened 2023-03-16
refactor: improve test coverage for CLI

IyadhKhalfallah:
---

- [ ] Added a test


joschkabraun (changes requested):
In `cli/__init__` test coverage is only ~50%. We can improve this by:
- adding more tests
- removing `logs` from parser (as it's removed either way
- removing `survey` from parser (as this is handled by JAC)

now/cli/parser.py
```diff
@@ -37,25 +37,6 @@ def set_base_parser():
     return parser
 
 
-def set_help_parser(parser=None):
```
makram93, line 40: Why are we removing this?

now/cli/__init__.py
```diff
@@ -18,13 +17,15 @@
...
+def get_run_args(parser=None):
     from now.cli.parser import get_main_parser
 
-    parser = get_main_parser()
+    if not parser:
+        parser = get_main_parser()
     if len(sys.argv) == 1:
         parser.print_help()
         exit()
+    print(parser.parse_known_args())
```
makram93, line 28: remove this

now/cli/parser.py
IyadhKhalfallah: @makram93 it is not used anywhere

tests/unit/test_cli.py
```diff
@@ -0,0 +1,32 @@
...
+
+from now.cli import cli, get_run_args
+
+
+class TestClass:
+    def __init__(self):
+        self.example = 'test'
+
+
+class Args:
```
joschkabraun, line 13: let's use `Namespace`

joschkabraun (commented):
As discussed in the pairing, nice increase in code cov!
Let's make sure that the other two tasks are also tested

now/cli/__init__.py
```diff
@@ -18,10 +17,11 @@
 # fix
 
 
-def _get_run_args():
+def get_run_args(parser=None):
```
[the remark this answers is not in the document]
IyadhKhalfallah: You're right, I was passing the argument in a previous test then I removed it. Thanks for the catch

tests/unit/test_cli.py
```diff
@@ -0,0 +1,32 @@
+from argparse import ArgumentParser, Namespace
+
+from pytest_mock import MockerFixture
+
+from now.cli import cli, get_run_args
+
+
+class TestClass:
```
joschkabraun, line 8: let's use `Namespace`

joschkabraun (changes requested):
Small suggestion, otherwise LGTM

Tanguyabel (approved):
LGTM

PR #1205, opened 2023-03-20
fix: use session token

makram93:
The current implementation of m2m token relies on storage being attached to the gateway, however this is not supported yet from JCloud. So, in case of pod restart, the `authorized_jwt` would be lost and if the user_token passed in `uses_with` expires then the billing won't work. Hence, we use an impersonation token, which is a fixed value per user per app and has no expiry date and to obtain this we need only user id and m2m token, hence we give up our dependency of user_token. We can always rely on impersonation token and treat it as a session token

---

- [ ] Added a test

### DIFF
--- a/now/executor/gateway/hubble_report.py
+++ b/now/executor/gateway/hubble_report.py
@@ -1,5 +1,4 @@
 import datetime
-import json
 import logging
 import os
 import threading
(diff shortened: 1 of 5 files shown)


now/executor/gateway/hubble_report.py
```diff
@@ -102,26 +97,20 @@ def report(quantity_basic, quantity_pro, auth_jwt=None):
...
 
-def get_summary(auth_jwt=None):
-    global authorized_jwt
-    global payment_client
-    m2m_token = os.environ.get('M2M')
-    payment_client = payment_client or PaymentClient(m2m_token=m2m_token)
-    authorized_jwt = authorized_jwt or auth_jwt
-    resp = payment_client.get_summary(token=authorized_jwt, app_id='search')  # type: ignore
+def get_summary():
+    resp = payment_client.get_summary(token=session_token, app_id='search')  # type: ignore
```
Tanguyabel, line 113: don't you need to declare `global session_token` here?

now/executor/gateway/now_gateway.py
```diff
@@ -141,21 +122,8 @@ def __init__(self, user_input_dict: Dict = {}, **kwargs):
...
-            else:
-                self.logger.info('No cred.json file found to load from')
-
-        try:
-            init_payment_params(
-                self.user_input.jwt['token'], self.authorized_jwt, self.storage_dir
-            )
-        except Exception as e:
-            self.logger.error(f'Could not init payment params')
+        self._check_env_vars()
```
Tanguyabel, line 125: isn't this redundant and/or conflicting with the check of M2M token line 55?

makram93: It is not conflicting but you could say it is redundant. I added an additional check at line 55 just as an extra precaution(which actually will never be needed). Should we keep it for better readability or do you recommend removing it?

now/executor/gateway/billing/client.py
```diff
@@ -0,0 +1,91 @@
...
+                    product_id,
+                    quantity,
+                    meta={
+                        'appId': 'search-app',
+                        'feeType': 'search-fee',
+                        'source': 'gateway',
+                    },
+                )
+                logger.info(f'**** `{round(quantity, 3)}` credits charged ****')
+                acc_status = self.get_account_status()
```
girishc13, line 60: Is the acc_status being fetched twice? The first time is in line 38.

makram93: Yes, to report credits before and after being charged. But it is unnecessary, as it is just for logs.

girishc13:
@Tanguyabel Please review. Looks ok from my side.

PR #1215, opened 2023-03-22
fix: remove grpc

joschkabraun:
---

- [ ] Added a test


now/app/base/app.py
```diff
@@ -111,13 +111,19 @@ def get_gateway_stub(self, user_input, testing=False) -> Dict:
...
             else 'NOWGateway',
-            'protocol': ['http', 'grpc'],
-            'port': [8081, 8085],
+            'protocol': ['http'],
+            'port': [8081],
             'monitoring': True,
             'cors': True,
             'uses_with': {'user_input_dict': user_input.to_safe_dict()},
-            'env': {'JINA_LOG_LEVEL': 'DEBUG'},
+            'env': {'JINA_LOG_LEVEL': 'DEBUG', 'M2M_TOKEN': os.getenv('M2M_TOKEN')},
```
IyadhKhalfallah, line 119: How do we get the M2M like this? I think this is not feasible because it needs to be injected in every flow, especially in JAC. What do you think?

IyadhKhalfallah (changes requested):

PR #1218, opened 2023-03-23
refactor: update jina version

florian-hoenicke:
At the moment, we don't have retries. But the new jina version supports them.


makram93:
Jina supports retries and we can see it in action [here](https://github.com/jina-ai/now/actions/runs/4512113138/jobs/7945173727#step:7:89). However, the retries also didn't help and this could be the corner case and should be investigated separately. Below is the failed message
```
ConnectionError: gRPC error: StatusCode.UNAVAILABLE DNS resolution failed for autocomplete-executor.jnamespace-1f297e756d.svc:80: C-ares status is not ARES_SUCCESS qtype=A name=autocomplete-executor.jnamespace-1f297e756d.svc is_balancer=0: Domain name not found |Gateway: Communication error with deployment autocomplete_executor at address(es) {'autocomplete-executor.jnamespace-1f297e756d.svc:80'}. Head or worker(s) may be down.
```

Even if we add another try-catch block outside `client.post` as done in this PR #1174, to catch the above exception and retry indexing with the remaining docs, the exec still does not come back online leading to failure at the indexing step as it can be seen in this [actions](https://github.com/jina-ai/now/actions/runs/4513708182/jobs/7958873723#step:7:158)

This could possibly be because of some problems in the deployment, i.e., some pods are not configured/set up properly as I have observed issues with pods in rare cases recently

now/deployment/deployment.py
```diff
@@ -38,10 +39,16 @@ def list_all_wolf(status='Serving', namespace='nowapi'):
...
-        flows.append(
-            {'id': flow['id'], 'name': flow['status']['endpoints'][executor_name]}
-        )
+        try:
+            executor_name = list(flow['status']['endpoints'].keys())[0]
+            flows.append(
+                {'id': flow['id'], 'name': flow['status']['endpoints'][executor_name]}
+            )
+        except Exception:  # noqa
+            print(f'Failed to parse flow {flow["id"]}')
```
joschkabraun, line 48: Is this still necessary? If so, let's print also `flow` variable

joschkabraun (approved):
LGTM!
Only small suggestion.

IyadhKhalfallah (approved):

PR #1223, opened 2023-03-24
test: improve gateway coverage

IyadhKhalfallah:
---

- [ ] Added a test


now/executor/gateway/now_gateway.py
```diff
@@ -14,7 +14,7 @@
 from streamlit.web.server import Server as StreamlitServer
 
 from now.constants import NOWGATEWAY_BFF_PORT
-from now.deployment.deployment import cmd
+from now.deployment import deployment
```
Tanguyabel, line 17: what is the rationale for this change?

tests/unit/test_gateway.py
```diff
@@ -0,0 +1,29 @@
...
+    save_cred(None)
+
+
+def test_gateway(mocker: MockerFixture):
+    runtime_args = {
+        'protocol': [0, 1],
+        'port': [8085, 8081],
+    }
+    mocker.patch('now.deployment.deployment.cmd', return_value=['PASSED', ''])
+    NOWGateway({}, runtime_args=runtime_args)
```
Tanguyabel, line 29: So you think there is no need for an assert, you only test that the Gateway built without breaking?

IyadhKhalfallah: Exactly, there should be nothing to expect as a result. What would you suggest?

now/executor/gateway/now_gateway.py
IyadhKhalfallah: This change is made in order to mock the cmd function. In the other case, it will be referenced from the packages and it becomes impossible to test it.

Tanguyabel: got it, makes sense thanks.

Tanguyabel (approved):

.github/codecov.yml
```diff
@@ -14,4 +14,4 @@ coverage:
     branches:               # branch names that can post comment
       - "main"
 ignore:
-  - "now/executor/gateway/playground/playground.py"
+  - "now/executor/gateway/playground/**/*"
```
joschkabraun, line 17: let's make sure that this is actually actually ignored

joschkabraun (commented):
Overall nice changes! Let's wait for codecov to actually ignore it

PR #1242, opened 2023-03-31
fix: remove autocomplete executor

joschkabraun:
Removes autocomplete executor as:
- it isn't needed
- the need to persist data on disk makes autoscale take a long time or fail

---

- [x] Updated the documentation
Screenshot 2023-03-31 at 11 35 32


IyadhKhalfallah (approved):
LGTM, we also need to cherry-pick this in jac_release branch.
Farewell autocomplete 🥺

PR #1243, opened 2023-04-02
refactor: update jina version and adapt

florian-hoenicke:
Update the jina version to latest and also adapt our codebase, if needed


tests/unit/test_gateway.py
```diff
@@ -16,12 +15,3 @@ def test_init_payment_params(mocker: MockerFixture):
 def test_save_cred():
     save_cred('./')
     save_cred(None)
-
-
-def test_gateway(mocker: MockerFixture):
```
makram93, line 21: Removed this test since it is not possible to instantiate `NOWGateway` standalone (due to it inheriting `CompositeGateway`)

Tanguyabel (approved):

PR #1244, opened 2023-04-04
fix: startup event in Starlette app BFF

Tanguyabel:


now/executor/gateway/bff/app/app.py
```diff
@@ -94,13 +87,20 @@ def build_app():
...
+    for route, sub_app in sub_apps.items():
+        app.mount(route, sub_app)
+
+    @app.on_event('startup')
+    def startup():
+        for sub_app in sub_apps.values():
+            logger.info(
+                f'Jina NOW started! ' f'Listening to [::]:{api_settings.DEFAULT_PORT}'
+            )
+
```
makram93, line 103:
```suggestion
    # startup event can only be fired for the parent app due to the lifecycle of the apps. 
    # In order to add some code for mounted (or sub) apps to be executed on the startup 
    # event, below is the hack.
    @app.on_event('startup')   # `app` here is the parent app
    def startup():
            logger.info(
                f'Jina NOW started! ' f'Listening to [::]:{api_settings.DEFAULT_PORT}'
            )
            # for sub_app in sub_apps.values():
                     # add your code specific to each sub-app which should run 
                     # when the startup event is fired

```

makram93 (changes requested):
Suggested some comment line to make it more verbose

now/executor/gateway/bff/app/app.py
Tanguyabel: Good idea!

makram93 (approved):
LGTM! 🚀

IyadhKhalfallah (approved):

PR #1251, opened 2023-04-05
feat: add s3 param for demos

IyadhKhalfallah:
---

- [ ] Added a test


now/data_loading/data_loading.py
```diff
@@ -407,22 +407,26 @@ def _list_s3_file_paths(bucket, folder_prefix):
...
     return [
         obj.key
         for obj in objects
         if not obj.key.endswith('/') and not obj.key.split('/')[-1].startswith('.')
     ]
 
 
-def _list_files_from_s3_bucket(
-    user_input: UserInput, data_class: Type
+def list_files_from_s3_bucket(
```
makram93, line 421: Do you intend to call this function on the API side for the S3 data preview?

now/data_loading/data_loading.py
```diff
@@ -407,22 +407,26 @@ def _list_s3_file_paths(bucket, folder_prefix):
     #             objects += f.result()
     # else:
     #     objects = list(bucket.objects.filter(Prefix=folder_prefix))
-    objects = list(bucket.objects.filter(Prefix=folder_prefix))
+    if top_n:
+        objects = list(bucket.objects.filter(Prefix=folder_prefix).limit(top_n))
+    else:
+        objects = list(bucket.objects.filter(Prefix=folder_prefix))
```
Tanguyabel, line 413:
```suggestion
    objects = bucket.objects.filter(Prefix=folder_prefix)
    if top_n:
        objects = objects.limit(top_n))
    objects = list(objects)
```

now/data_loading/data_loading.py
IyadhKhalfallah: Yes

now/data_loading/data_loading.py
IyadhKhalfallah: @Tanguyabel The suggested solution actually filters everything, and then it limits. Thus, the performance will drop when I want to take only the first 10 elements. Adding `.limit(top_n)` however will limit before getting the actual data, so it is very efficient that way!

now/data_loading/data_loading.py
```diff
@@ -407,22 +407,54 @@ def _list_s3_file_paths(bucket, folder_prefix):
...
     return [
         obj.key
         for obj in objects
         if not obj.key.endswith('/') and not obj.key.split('/')[-1].startswith('.')
     ]
 
 
-def _list_files_from_s3_bucket(
-    user_input: UserInput, data_class: Type
+def validate_shrink_data(file_paths):
```
joschkabraun, line 421: Can you add documentation on the functionality this function executes?

joschkabraun (changes requested):
After adding documentation + test this LGTM

now/data_loading/data_loading.py
IyadhKhalfallah: Sure, will do!

now/data_loading/data_loading.py
Tanguyabel: Sure, I thought of this, and thought the query is not evaluated until you generate a list from it. But if you are sure performances are impacted, then please leave it as it.

Tanguyabel (approved):

joschkabraun (approved):

PR #1253, opened 2023-04-06
feat: add report usage meta

IyadhKhalfallah:
Add some additional meta when reporting credit usage to Hubble (This meta can be used to filter on JAC side or for statistics)
Note: meta only includes `search` fee-related credits param. `base` fee-related param will be added in the base fee script in now.private

---

- [ ] Added a test


now/executor/gateway/hubble_report.py
```diff
@@ -90,7 +90,17 @@ def report(quantity_basic, quantity_pro, auth_jwt=None):
...
             quantity = quantity_pro
         if can_charge(summary):
-            payment_client.report_usage(authorized_jwt, app_id, product_id, quantity)  # type: ignore
+            payment_client.report_usage(
+                authorized_jwt,
+                app_id,
+                product_id,
+                quantity,
+                meta={
+                    'appId': 'search-app',
```
Tanguyabel, line 99: Isn't it a duplicate of the parameter `app_id`?

makram93: yes, but the meta is like additional filtering fields. We can add any key-value pair here. This is to help filter the charges based on the meta

Tanguyabel (approved):

PR #1268, opened 2023-04-20
fix: delete failed flows

makram93:
---

- [ ] Added a test


now/deployment/flow.py
```diff
@@ -1,28 +1,56 @@
...
+        try:
+            flow = deployment.deploy_wolf(path=flow_yaml)
+        except BaseException as e:  # noqa
+            # Fetch flow for this user and delete the recent failed app
+            deleted_flow_id = None
+            all_flows = deployment.list_all_wolf(status='Failed')
+            all_flows.sort(key=lambda x: x['created_at'], reverse=True)
+            for f in all_flows:
+                # find the first failed flow containing the name and delete it
+                if flow_name in f['id']:
```
Tanguyabel, line 46: This is very dangerous. We cannot guarantee that this is the flow that was just deployed and failed. I think we cannot rely on this. Instead, let's see if we can get more info returned by the deployment call, in case of failure. If not we need to carry some tags or find a way to identify the right flow.

now/deployment/flow.py
```diff
@@ -1,28 +1,56 @@
...
+                    flow = deployment.terminate_wolf(flow_id=f['id'])
+                    deleted_flow_id = f['id']
+                    break
+            print(f'Failed to deploy flow. Deleting flow id `{deleted_flow_id}`')
+            raise e
 
+    time.sleep(10)  # sleep for some time before starting with indexing
```
Tanguyabel, line 53: This should not be needed when using a proper retry strategy, with `max_attempts`.

tests/unit/test_flow.py
```diff
@@ -0,0 +1,42 @@
...
+from now.deployment.flow import deploy_flow
+
+
+@pytest.fixture
+def get_dummy_flow(tmpdir, mocker):
+    with open(tmpdir / 'flow.yml', 'w') as f:
+        f.write('jtype: Flow')
+    flow = CloudFlow(path=f'{tmpdir}/flow.yml', flow_id='nowapi-e3fd500bdf')
+    flow.endpoints = {'gateway (http)': 'http://somehost:1234'}
+    mocker.patch('time.sleep', return_value=None)
```
Tanguyabel, line 14: I am not sure why you need to patch this.

now/deployment/flow.py
```diff
@@ -1,28 +1,56 @@
...
+            write_flow_file(flow_yaml, flow_file)
+            flow_yaml = flow_file
+        else:
+            print(f'Yaml file exists at {flow_yaml}.')
+
+        # Get the flow name from the yaml file
+        try:
+            flow_name = read_flow_file(flow_yaml)['jcloud']['name']
+        except:  # noqa
+            flow_name = 'nowapi'
```
Tanguyabel, line 34: Let's make this name unique with a uuid suffix, so we can use it later to filter out the flows and fetch the one.

now/deployment/flow.py
Tanguyabel: Let's reach out to team wolf to see if they can send back the uuid generated for the flow.

now/deployment/flow.py
makram93: Tao recommended adding this for the first index call.

IyadhKhalfallah (approved):
LGTM

PR #1270, opened 2023-04-20
fix: transform tensor to blob

IyadhKhalfallah:
---

- [ ] Added a test


makram93 (changes requested):

now/data_loading/data_loading.py
```diff
@@ -79,6 +82,36 @@ def _add_metadata_to_chunks(da, user_input):
...
+        _uri_to_blob(uri)
+        return True
+    except:
+        return False
+
+
+def _convert_document_tensors_to_blob(doc):
+    for chunk in doc.chunks:
+        if chunk.uri and _is_valid_uri(chunk.uri):
+            return
```
makram93, line 96:
```suggestion
            chunk.tensor = None
            return
```

makram93 (changes requested):

now/data_loading/data_loading.py
```diff
@@ -79,6 +82,43 @@ def _add_metadata_to_chunks(da, user_input):
...
+def _convert_da_tensors_to_blob(documents):
+    with ThreadPoolExecutor(max_workers=10) as executor:
+        futures = []
+        for doc in documents:
+            f = executor.submit(
+                _convert_document_tensors_to_blob,
+                doc,
+            )
+            futures.append(f)
+        for f in futures:
```
girishc13, line 118: If using the `with` [statement](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.shutdown) then there is not need to wait for the futures explicitly.
```
You can avoid having to call this method explicitly if you use the [with](https://docs.python.org/3/reference/compound_stmts.html#with) statement, which will shutdown the [Executor](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor) (waiting as if [Executor.shutdown()](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.shutdown) were called with wait set to True):
```

girishc13 (approved):

